PR #98, opened 2016-06-08
SW-HarvestBroker - seeds calls author.approved_sciencewire_ids

darrenleeweber:
Revised the seeds method so the code/specs are consistent and added a spec for the seeds method.

### DIFF
--- a/lib/science_wire/harvest_broker.rb
+++ b/lib/science_wire/harvest_broker.rb
@@ -71,8 +71,7 @@ module ScienceWire
       # Accessors for custom Author information. Eventually could migrate to
       # AuthorAttributes as the client changes underneath.
       def seed_list
-        @seed_list ||= author.publications.approved.with_sciencewire_id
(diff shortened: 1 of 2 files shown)


lib/science_wire/harvest_broker.rb
```diff
@@ -71,8 +71,7 @@ def ids_from_smart_query(author_attributes)
       # Accessors for custom Author information. Eventually could migrate to
       # AuthorAttributes as the client changes underneath.
       def seed_list
-        @seed_list ||= author.publications.approved.with_sciencewire_id
-                             .pluck(:sciencewire_id).uniq
+        @seed_list ||= author.approved_sciencewire_ids
       end
```
darrenleeweber: For reference, the `Author` model contains:
```ruby
  # @return [Array<Integer>] ScienceWireIds for approved publications
  def approved_sciencewire_ids
    publications.approved.with_sciencewire_id.pluck(:sciencewire_id).uniq
  end
```

PR #103, opened 2016-06-09
handle case of missing ValidYN attribute

### LINKED ISSUE #74: PubmedSourceRecord missing ValidYN attribute in CAP-QA
_From @drh-stanford on June 6, 2016 20:57_

Reported by squash in the CAP-QA environment during the nightly `cap:poll`:

https://sul-squash-prod.stanford.edu/projects/cap/environments/cap-qa/bugs/4#history

Here is an example Author on which it failed: 

```
E, [2016-06-06T05:15:10.973745 #21451] ERROR -- : ScienceWireHarvester: Harvest failed for Author.find_by(id: 47525, cap_profile_id:
 93004)
```

_Copied from original issue: sul-dlss/sul-pub#521_

drh-stanford:
This PR fixes #74. I've made the assumption that if `ValidYN` is *missing* then we should ignore it.

### DIFF
--- a/app/models/pubmed_source_record.rb
+++ b/app/models/pubmed_source_record.rb
@@ -184,8 +184,8 @@ class PubmedSourceRecord < ActiveRecord::Base
       ##
       # Ignore an <Author> that contains only <CollectiveName>
       return if author.xpath('CollectiveName').present?
-      # Ignore an <Author ValidYN="N">
-      return if author.attributes['ValidYN'].value == 'N'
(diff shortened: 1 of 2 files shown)


darrenleeweber:
Let's investigate the assumption, because it's very important to parse author name data.  It might be better to try doing it when that field is missing.

darrenleeweber:
Ran into this bug while processing http://www.ncbi.nlm.nih.gov/pubmed/11736333

PR #104, opened 2016-06-09
inject the author identity (cap_profile_id) into the authorship data

### LINKED ISSUE #75: ArgumentError: Invalid authorship in CAP-QA
_From @drh-stanford on June 6, 2016 21:2_

During the `cap:poll` we're seeing these exceptions:

```
E, [2016-06-06T05:06:18.071007 #21451] ERROR -- : CapAuthorsPoller: Invalid authorship: Author.find_by(cap_profile_id: 29639); {"fea
tured"=>false, "status"=>"new", "sulPublicationId"=>219726, "visibility"=>"public"}
E, [2016-06-06T05:06:18.071032 #21451] ERROR -- : #<ArgumentError: Invalid authorship: Author.find_by(cap_profile_id: 29639); {"feat
ured"=>false, "status"=>"new", "sulPublicationId"=>219726, "visibility"=>"public"}>

PRODUCTION [1] pry(main)> Author.find_by(cap_profile_id: 29639)
=> #<Author:0x00000007012690
 id: 13921,
 cap_profile_id: 29639,
 active_in_cap: true,
 sunetid: "REDACTED",
 university_id: REDACTED,
 email: "REDACTED",
 cap_first_name: "Jelena",
 cap_last_name: "Obradovic",
 cap_middle_name: "",
 official_first_name: "Jelena",
 official_last_name: "Obradovic",
 official_middle_name: "",
 preferred_first_name: "Jelena",
 preferred_last_name: "Obradovic",
 preferred_middle_name: "",
 created_at: Mon, 17 Jun 2013 23:20:28 UTC +00:00,
 updated_at: Fri, 24 Jan 2014 09:17:40 UTC +00:00,
 california_physician_license: "",
 cap_import_enabled: true,
 emails_for_harvest: "REDACTED">

PRODUCTION [3] pry(main)> Contribution.where(publication_id: 219726)
=> [#<Contribution:0x00000005063718
  id: 322425,
  author_id: 13921,
  cap_profile_id: 29639,
  publication_id: 219726,
  status: "new",
  featured: false,
  visibility: nil,
  created_at: Tue, 30 Jul 2013 09:36:57 UTC +00:00,
  updated_at: Sun, 27 Mar 2016 07:42:28 UTC +00:00>]

```

_Copied from original issue: sul-dlss/sul-pub#522_

drh-stanford:
This PR fixes #75 by injecting the `cap_profile_id` into the `authorship` data we get from the CAP database before validating it for an existing author.

### DIFF
--- a/lib/cap_authors_poller.rb
+++ b/lib/cap_authors_poller.rb
@@ -126,6 +126,11 @@ class CapAuthorsPoller
   end
 
   def update_existing_contributions(author, incoming_authorships)
+    # inject author identity into the authorship data
+    incoming_authorships = incoming_authorships.dup.map do |contrib|
+      contrib[:cap_profile_id] = author.cap_profile_id
+      contrib
+    end
     incoming_authorships.each do |authorship|
       if !Contribution.authorship_valid? authorship
         msg = "Invalid authorship: Author.find_by(cap_profile_id: #{author.cap_profile_id}); #{authorship.inspect}"


lib/cap_authors_poller.rb
```diff
@@ -126,6 +126,11 @@ def queue_author_for_harvest(author, skip_message)
   end
 
   def update_existing_contributions(author, incoming_authorships)
+    # inject author identity into the authorship data
+    incoming_authorships = incoming_authorships.dup.map do |contrib|
```
peetucket: does this mean we would store the injected cap_profile_id in this data structure, or is this purely to pass validation?

lib/cap_authors_poller.rb
```diff
@@ -126,6 +126,11 @@ def queue_author_for_harvest(author, skip_message)
...
+      contrib[:cap_profile_id] = author.cap_profile_id
+      contrib
+    end
```
darrenleeweber: This seems to be an odd addition to require, because the `incoming_authorships` data is coming from the CAP API and it should already contain the `cap_profile_id` or a SUL Author.id value.

lib/cap_authors_poller.rb
drh-stanford: It does a `dup` so it's just for the validation to pass.

lib/cap_authors_poller.rb
darrenleeweber: For more perspective on this, see also helper methods in
https://github.com/sul-dlss/sul_pub/blob/master/app/api/sul_bib/authorship_api.rb

drh-stanford: See #75 for details on the data coming back from the CAP API -- I'm not seeing the profile id values...

darrenleeweber: OK, I see.  Perhaps try:
```ruby
contrib[:sul_author_id] ||= author.id  # this is reliable
contrib[:cap_profile_id] ||= author.cap_profile_id  # this is not reliable
```

lib/cap_authors_poller.rb
darrenleeweber: /shrug on the `dup` because it gets assigned back to the same variable name within the method, but at least avoids changing the mutable object elsewhere, although the changes are not likely to impact anything else.

lib/cap_authors_poller.rb
darrenleeweber: Note also that
- https://github.com/sul-dlss/sul_pub/blob/master/lib/cap_authors_poller.rb#L130
Could be modified to only validate the fields by instead just calling
- https://github.com/sul-dlss/sul_pub/blob/master/app/models/contribution.rb#L32

drh-stanford: Yeah, thanks. That's probably the better solution.

drh-stanford:
Not going with this solution. See PR #107 instead.

PR #108, opened 2016-06-14
86 missing wos ids

### LINKED ISSUE #86: Some publications from WoS (provenance=sciencewire) are missing WoS IDs?
Pointed out by Tina: Investigate missing WoS ID's for publications that come from WoS (provenance is sciencewire), but for some reason do not have the WoS id in the json.

darrenleeweber:
This script will log some details for all the ScienceWire publications with a missing WoSItemID.  On a prod-db-dump, all of the identified publications did not have a WoSItemID in the ScienceWireSourceRecord.  No attempt is made to re-harvest the publications; if that is a requirement, it can be added, but that will be extra work to change this script from a read-only operation into an update operation.

This should fix #86

### DIFF
--- a/app/models/publication.rb
+++ b/app/models/publication.rb
@@ -232,7 +232,7 @@ class Publication < ActiveRecord::Base
       if i.persisted?
         i.save
       else
-        publication_identifiers << i
+        publication_identifiers << i unless publication_identifiers.include? i
       end
     end
   end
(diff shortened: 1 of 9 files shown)


darrenleeweber:
We've discovered that some SW pubs have been updated with WoSItemID values, so they should be harvested when missing.  This script can be enhanced to do so, marked this PR as [WIP] to evaluate this.

darrenleeweber:
Running the new script on my laptop prod-db-dump to check it out.

drh-stanford:
This PR looks good. Can you add the scriptie description to https://github.com/sul-dlss/sul_pub/wiki/Useful-Rails-runner-scripties?

darrenleeweber:
Updated https://github.com/sul-dlss/sul_pub/wiki/Useful-Rails-runner-scripties

PR #124, opened 2016-10-13
Replace squash with honeybadger and simplify exception reporting

### LINKED ISSUE #121: monitoring framework: Switch from squash to honeybadger
see https://github.com/sul-dlss/sul_pub/issues/119

cbeer:
Fixes #121

### DIFF
--- a/Capfile
+++ b/Capfile
@@ -25,8 +25,9 @@ require 'capistrano/bundler'
 require 'capistrano/rails'
 require 'capistrano/passenger'
 require 'dlss/capistrano'
-require 'squash/rails/capistrano3' # Note that this inclusion requires a branch, not a tag, to be deployed
 require 'whenever/capistrano'
 
 # Loads custom tasks from `lib/capistrano/tasks' if you have any defined.
 Dir.glob('lib/capistrano/tasks/*.rake').each { |r| import r }
+
+require 'capistrano/honeybadger'
--- a/Gemfile
+++ b/Gemfile
@@ -46,9 +46,7 @@ gem 'whenever', require: false
 gem 'yaml_db'
 
 # -------------------
-# Use Squash for exception reporting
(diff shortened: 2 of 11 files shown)


drh-stanford:
The coverage report shouldn't be a problem although it marked the PR as failed. 

But I'm confused as to where the Honeybadger configuration comes from (i.e., the API key)? If it's shared configs, shouldn't the deploy scriptie link to it?

cbeer:
@drh-stanford you're right, thanks for catching it. fixed up now.

drh-stanford:
hmmm. i'm not seeing the `honeybadger.yml` in the `sul-pub-prod` branch in shared configs but it is in `sul-pub-cap-qa`.  also, is there a noop `honeybadger.yml` file that we can put for developer laptops in `sul-pub-dev-solo`?

cbeer:
I put the file on this branch: https://github.com/sul-dlss/shared_configs/blob/sulcap-prod/config/honeybadger.yml. I don't see a `sul-pub-prod` branch?

The absence of the file should be sufficiently a no-op for our purposes, no?

drh-stanford:
@tallenaz what are the correct shared config branches for CAP?

tallenaz:
@drh-stanford puppet shows that the shared_configs branches are named `sul-pub-prod`, `sul-pub-cap-dev`, `sul-pub-cap-qa` and `sul-pub-dev`

drh-stanford:
Ok, so we need some PRs that puts the `honeybadger.yml` into each of the shared config branches described ^^^. Because the CAP-QA and CAP-Dev both run in Rails `production` mode, we'll need to set the `env` parameter in `honeybadger.yml` I believe (see http://docs.honeybadger.io/lib/ruby.html).

Also, there are "CAP" and "cap" projects in Honeybadger. It looks like "CAP" is being used but not "cap" -- can we delete that one?

PR #135, opened 2016-10-21
Use faraday https for pubmed calls

### LINKED ISSUE #126: PubMed search - must use HTTPS before Nov 9, 2016
PubMed is moving towards HTTPS.  Does the `Net::HTTP` internals using HTTPS in this code at
https://github.com/sul-dlss/sul_pub/blob/master/lib/pubmed_client.rb#L13-L40

peetucket:
fixes #126


lib/pubmed_client.rb
```diff
@@ -15,25 +15,25 @@ def fetch_records_for_pmid_list(pmids)
...
+    conn.options.open_timeout = 10
+
+    response = conn.send(:post) do |req|
+      req.url Settings.PUBMED.FETCH_PATH
+      req.headers['Host'] = Settings.PUBMED.HOST
+      req.body = pmidValuesForPost
+    end
+
+    response.body
+
```
darrenleeweber: Looks good.  Seems a little odd that the host must be set in the header as well.

The only thing missing now is the exception handling and sending `NotificationManager#error`.  There should be a unit test for this somewhere too.

peetucket: yes, removed that extra header setting

what do you mean by the exceptional handling with NotificationManager?  I think that is still handled as before with the rescue block still intact

peetucket: There are tests here that use cassettes to verify the connection: https://github.com/sul-dlss/sul_pub/blob/master/spec/lib/pubmed_client_spec.rb

Do you think we need something else?  And stubbing the call out without actually making it?

darrenleeweber: Let's pair to review and discuss details.  As I read the changes here on github, the rescue block has been removed.  Are we looking at the same thing?

darrenleeweber: I've changed my github view to the 'split' mode to take another look at this.  Now I see that the exception block has been changed but not removed entirely.  For some reason, github was hiding the block of code at the bottom.

VCR should replace those cassettes when this code changes.  To be sure, it can help to `rm` those files and run the specs again and then check the `git diff`

darrenleeweber: Usually the `rescue` indentation matches the method `def` when it is a global method rescue (rather than a rescue on a block within the method).

darrenleeweber:
The `Settings.PUBMED.HOST` is used only in this PubMed client, nowhere else in this project.  It's not likely to be a setting that needs to be changed.  As it is, it does not include the `https://` protocol prefix in the setting (so we had to make a code change to fix this).  It makes sense to either:
- a.  include 'https://' protocol prefix in the setting, or
- b.  remove the setting and put the entire URL in the client

peetucket:
Updated branch with new indentation and moving protocol to settings

config/settings.yml
```diff
@@ -29,7 +29,7 @@ CAP:
 PUBMED:
   LOG: log/pubmed.log
   FETCH_PATH: /entrez/eutils/efetch.fcgi?db=pubmed&retmode=xml
-  HOST: eutils.ncbi.nlm.nih.gov
+  HOST: https://eutils.ncbi.nlm.nih.gov
```
darrenleeweber: Sorry to be so pendantic, but now I'm wondering whether this should be `PUBMED.BASE_URL` to distinguish it from the other settings that do not have the protocol in the `.HOST` values?  Another option is to separate out the protocol, as in:
```ruby
"#{PUBMED.PROTOCOL}#{PUBMED.HOST}"
```
If the current solution is sufficient, go with it.  Just offering these thoughts rather than let them drift into the ether.  If you want to keep the current solution, we can merge as is (everything else looks good).

darrenleeweber:
Looks like it needs to resolve conflicts.

peetucket:
rebased

config/settings.yml
```diff
@@ -29,15 +29,14 @@ CAP:
...
   FETCH_PATH: /entrez/eutils/efetch.fcgi?db=pubmed&retmode=xml
-  HOST: https://eutils.ncbi.nlm.nih.gov
-
+  BASE_URI: https://eutils.ncbi.nlm.nih.gov
+  
 ## Sciencewire Auth Config
 SCIENCEWIRE:
   LOG: log/sciencewire.log
-  BASE_URI: sciencewirerest.discoverylogic.com
+  BASE_URI: https://sciencewirerest.discoverylogic.com:443
```
darrenleeweber: The default HTTPS port is 443, so do we need to specify this port in the URL?

peetucket:
yes, i doubt the port is necessary (though it was there before), so I removed it from the sciencewire URL

PR #139, opened 2016-10-25
notify Honeybadger when external service monitoring fails

drh-stanford:
@eefahy notifies Honeybadger when the following monitoring endpoints generate a failure:

- `/status/external-CapHttpClient`
- `/status/external-PubmedClient`
- `/status/external-ScienceWireClient`

### DIFF
--- a/config/initializers/okcomputer.rb
+++ b/config/initializers/okcomputer.rb
@@ -24,10 +24,12 @@ class DelegateCheck < OkComputer::Check
     else
       mark_failure
       mark_message 'not working'
+      Honeybadger.notify(RuntimeError.new("#{delegate.name} is not working"))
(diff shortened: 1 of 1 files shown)


peetucket:
is this a notification when the service is down or when the attempt to check if the service is working fails

drh-stanford:
it's the latter -- it'll only notify Honeybadger when the monitoring endpoint is triggered

eefahy:
so that means that we will get alerts from both nagios (to the listserv) and send an exception to honeybadger? Do we need both?

drh-stanford:
Right. I'll close this PR and leave it so that Nagios will be the only thing that sends alerts for the external services checks.

PR #140, opened 2016-10-25
switch pubmed connections to use Faraday and HTTPS so they can contin…

peetucket:
…ue working after Nov 9 2016

addresses #135 in prod


drh-stanford:
This is a straight port of PR https://github.com/sul-dlss/sul_pub/pull/135, right?

peetucket:
Yes, basically a straight port (not done automatically though due to a bit of pain with cherry-picking).  VCR cassettes were updated to match the new HTTPS calls to PubMed.

PR #152, opened 2016-12-09
set up a scheduled rake task for performing external monitor checks

### LINKED ISSUE #119: monitoring framework: Switch from is_it_working to okcomputer and split checks
We need to monitor two things:
1. Are the parts of our application that are under our immediate control operational (e.g. the app is running, the database is up, solr is up).  
2. Are all of the external dependencies working (e.g. the CAP API, ScienceWire API, PubMed).

Currently is_it_working monitors both.  What we want to do is have Nagios/OpsGenie alert only one 1 above (since this is what we can actually act on), while having a second monitoring system (HoneyBadger?) send email alerts to the team on 2 above (since it is still useful to know and may trigger an email or investigation on why the API call is not working).

peetucket:
Added a new rake task that calls the external okcomputer checks for a given URL.  Added a new role for the whenever gem that schedules this task every 30 minutes in cap-qa and production.

If an exception occurs during the checks, a honeybader notification will be sent out (this is already working in the code). 

With this, we can remove the nagios alerts for these three external checks.

Note I shelled out to 'hostname' to get the URL to check.  Since cap-qa and prod both run under Rails.env == production, you can't use a hostname that is configured in the Rails app.  I also couldn't find a way to pass a value from the cap script to the schedule.rb file.

fixes #119

### DIFF
--- a/config/schedule.rb
+++ b/config/schedule.rb
@@ -15,3 +15,8 @@ end
 every 1.day, at: stagger(4), roles: [:harvester] do
   rake 'cap:poll[1]'
 end
+
+# call a rake task that hits the OK computer external checks, triggering honeybadger notifications if there is an error
+every 30.minutes, roles: [:external_monitor] do
+  rake "sul:check_external_services[\"https://#{`hostname`.chomp!}\"]"
+end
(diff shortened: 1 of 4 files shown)


config/schedule.rb
```diff
@@ -15,3 +15,8 @@ def stagger(hour)
 every 1.day, at: stagger(4), roles: [:harvester] do
   rake 'cap:poll[1]'
 end
+
+# call a rake task that hits the OK computer external checks, triggering honeybadger notifications if there is an error
+every 30.minutes, roles: [:external_monitor] do
+  rake "sul:check_external_services[\"https://#{`hostname`.chomp!}\"]"
```
drh-stanford: i think you can just use `localhost` here, but this is fine.

PR #156, opened 2017-01-19
add new rake task to bulk import a list of publications

peetucket:
given a list of PMIDs and cap profile id

Useful when I need to batch import pubs for a new faculty member.  There is an existing similar rake task for doing this by WoSID, but sometimes there are pubs that only exist in PubMed and not yet in Web of Science

### DIFF
--- a/lib/tasks/pubmed.rake
+++ b/lib/tasks/pubmed.rake
@@ -10,4 +10,36 @@ namespace :pubmed do
     doc = client.fetch_records_for_pmid_list(pmids)
     puts doc # XML document
   end
+
+  desc 'Harvest using a text file with a list of publications by PubmedID for the supplied cap_profile_id'
+  task :pmid_profile_id_import, [:path_to_report,:cap_profile_id] => :environment do |_t, args|
+    lines = IO.readlines args[:path_to_report]
+    author = Author.find_by_cap_profile_id(args[:cap_profile_id])
+
+    puts "Cap_profile_id #{args[:cap_profile_id]} is #{author.first_name} #{author.last_name}"
+    total_pub_count=author.contributions.size
+    new_pub_count=author.contributions.where(:status=>'new').size
+    approved_pub_count=author.contributions.where(:status=>'approved').size
+    puts "total publications: #{total_pub_count}"
+    puts "total new publications: #{new_pub_count}"
+    puts "total accepted publications: #{approved_pub_count}"
+    puts "attempting to harvest #{lines.count} new publications by pubmedID"
+    puts
+
+    harvester=ScienceWireHarvester.new
+    lines.each do |line|
+      pmid=line.chomp
+      puts "working on #{pmid}..."
+      pub=Publication.find_or_create_by_pmid(pmid)
+      harvester.add_contribution_for_harvest_suggestion(author,pub)
+      pub.add_all_db_contributions_to_my_pub_hash
+      pub.save
+    end
+
+    author.reload
+    new_pub_count=author.contributions.where(:status=>'new').size - new_pub_count
+    puts ""
+    puts "added publications: #{new_pub_count}"
+
+  end
 end
--- a/lib/tasks/pubmed.rake
+++ b/lib/tasks/pubmed.rake
@@ -11,10 +11,14 @@ namespace :pubmed do
     puts doc # XML document
   end
 
-  desc 'Harvest using a text file with a list of publications by PubmedID for the supplied cap_profile_id'
+  desc 'Harvest using a plain text file with a list of publications by PubmedID (no header row) for the supplied cap_profile_id'
+  # file format is a plain text file, no header row, one line per PubmedID
   task :pmid_profile_id_import, [:path_to_report,:cap_profile_id] => :environment do |_t, args|
-    lines = IO.readlines args[:path_to_report]
     author = Author.find_by_cap_profile_id(args[:cap_profile_id])
+    abort "cap_profile_id #{args[:cap_profile_id]} not found" if author.nil?
+
+    abort "#{args[:path_to_report]} not found" unless File.file?(args[:path_to_report])
+    lines = IO.readlines args[:path_to_report]
 
     puts "Cap_profile_id #{args[:cap_profile_id]} is #{author.first_name} #{author.last_name}"
     total_pub_count=author.contributions.size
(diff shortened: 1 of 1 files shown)


lib/tasks/pubmed.rake
```diff
@@ -10,4 +10,36 @@ namespace :pubmed do
     doc = client.fetch_records_for_pmid_list(pmids)
     puts doc # XML document
   end
+
```
darrenleeweber: Describe input file format expectations - CSV?  One PMID per line?

lib/tasks/pubmed.rake
```diff
@@ -10,4 +10,36 @@ namespace :pubmed do
...
+  desc 'Harvest using a text file with a list of publications by PubmedID for the supplied cap_profile_id'
+  task :pmid_profile_id_import, [:path_to_report,:cap_profile_id] => :environment do |_t, args|
```
darrenleeweber: Suggesting an alt-name for the task:  `profile_import_pmid_file` (no change required if the current name is good enough or consistent with some other task names).

lib/tasks/pubmed.rake
```diff
@@ -10,4 +10,36 @@ namespace :pubmed do
...
+    lines = IO.readlines args[:path_to_report]
+    author = Author.find_by_cap_profile_id(args[:cap_profile_id])
```
darrenleeweber: Just curious about whether it requires `args[:cap_profile_id].to_i` (maybe A.R. takes care of any String/Integer conversions).

lib/tasks/pubmed.rake
```diff
@@ -10,4 +10,36 @@ namespace :pubmed do
...
+
+    puts "Cap_profile_id #{args[:cap_profile_id]} is #{author.first_name} #{author.last_name}"
+    total_pub_count=author.contributions.size
+    new_pub_count=author.contributions.where(:status=>'new').size
+    approved_pub_count=author.contributions.where(:status=>'approved').size
+    puts "total publications: #{total_pub_count}"
+    puts "total new publications: #{new_pub_count}"
+    puts "total accepted publications: #{approved_pub_count}"
+    puts "attempting to harvest #{lines.count} new publications by pubmedID"
+    puts
```
darrenleeweber: It's not important here, but consider the performance tradeoffs of doing multiple SQL queries on contributions vs. getting them all back and then doing the filtered counts using a ruby `select` or `count` (in-memory), e.g.
```ruby
x = [*1..10]
# => [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
> x.count {|i| i % 2 == 0 }
#=> 5
```

lib/tasks/pubmed.rake
```diff
@@ -10,4 +10,36 @@ namespace :pubmed do
     doc = client.fetch_records_for_pmid_list(pmids)
     puts doc # XML document
   end
+
+  desc 'Harvest using a text file with a list of publications by PubmedID for the supplied cap_profile_id'
+  task :pmid_profile_id_import, [:path_to_report,:cap_profile_id] => :environment do |_t, args|
+    lines = IO.readlines args[:path_to_report]
```
darrenleeweber: Not really important - maybe read the file after first confirming that the CAP profile id can be found?  If the wrong profile ID is input, no point in reading the file.

lib/tasks/pubmed.rake
```diff
@@ -10,4 +10,36 @@ namespace :pubmed do
...
+
+    harvester=ScienceWireHarvester.new
+    lines.each do |line|
+      pmid=line.chomp
+      puts "working on #{pmid}..."
+      pub=Publication.find_or_create_by_pmid(pmid)
+      harvester.add_contribution_for_harvest_suggestion(author,pub)
+      pub.add_all_db_contributions_to_my_pub_hash
+      pub.save
+    end
```
darrenleeweber: :+1: - this is the solution that was recently tested, so we know that it works out OK right?

lib/tasks/pubmed.rake
```diff
@@ -10,4 +10,36 @@ namespace :pubmed do
...
+
+    author.reload
+    new_pub_count=author.contributions.where(:status=>'new').size - new_pub_count
+    puts ""
+    puts "added publications: #{new_pub_count}"
+
```
darrenleeweber: :+1: - looks like good feedback to the console on how things are going.  But, wondering what happens when there is an exception raised for any one of the `line` values?  It might be worth using a `begin - rescue` block within the `line` iteration to catch occasional exceptions and continue on with the next line.

lib/tasks/pubmed.rake
peetucket: will describe in a comment -- it is just a plain text file with no header, one row per PMID (no other columns)

lib/tasks/pubmed.rake
peetucket: i agree the current name is a bit clunky, but I named it this way on purpose for consistency with the corresponding and similar WOS ID import here https://github.com/sul-dlss/sul_pub/blob/master/lib/tasks/sw.rake#L58

lib/tasks/pubmed.rake
peetucket: yup, will check to see if author exists before reading the file

lib/tasks/pubmed.rake
peetucket: it will work either way.  AR is smart

lib/tasks/pubmed.rake
peetucket: personally i would leave the load on SQL to do the joins and counts and count on its own internal intelligence/caching instead of pulling everything into a ruby array object and iterating there...pragmatically its not likely to matter much either way since any one given author is unlikely to have a large enough number of contributions to cause a problem either way

lib/tasks/pubmed.rake
peetucket: yup, this was the final solution that worked for me

lib/tasks/pubmed.rake
peetucket: good idea, will add a begin rescue block within the loop to continue even if there is an error (but display the exception that occurred)

peetucket:
Updated with new commit based on comments

lib/tasks/pubmed.rake
darrenleeweber: OK, no problem.  (Not sure how to mark this bit resolved, but no worries.)

PR #169, opened 2017-03-24
add shared config rake task capability

peetucket:
To use, run "bundle exec cap [ENVIRONMENT] shared_configs:update", e.g.:

```
└─▪ bundle exec cap staging shared_configs:update
Pseudo-terminal will not be allocated because stdin is not a terminal.
[Deprecation Notice] Future versions of Capistrano will not load the Git SCM
plugin by default. To silence this deprecation warning, add the following to
your Capfile:

    require "capistrano/scm/git"
    install_plugin Capistrano::SCM::Git

00:00 shared_configs:pull
      01 cd /opt/app/pub/sul-pub/shared/repo_configs; git pull
      01 Already up-to-date.
    ✔ 01 pub@sul-pub-stage.stanford.edu 1.705s
00:02 shared_configs:symlink
      01 cd /opt/app/pub/sul-pub/shared/repo_configs; cp -rlf * /opt/app/pub/sul-pub/shared
    ✔ 01 pub@sul-pub-stage.stanford.edu 0.283s
Pseudo-terminal will not be allocated because stdin is not a terminal.
```


darrenleeweber:
How do we know what branch is checked out in the shared_configs?  How is that configured for this gem?

peetucket:
Based on the output of the rake task, it simply goes onto the server for the environment you specify (already configured in one of the config/deploy/ENV.rb files, goes into the correct shared_config repo location and does a git pull followed by the symlink.  Since it is a git pull and not a git clone, the knowledge of which branch is checked out is already on the server and does not need to be configured here.  This means this is only useful for updating shared configs on a server where it was already setup, not setting it up in the very first instance (which is done by devops per the README for shared_configs).

PR #173, opened 2017-05-02
Add pmc ID to pub hash

peetucket:
Add PMC identifiers to pub_hash from pubmed source record when available.

Need to also rebuild all pub hashes

### DIFF
--- a/app/models/publication.rb
+++ b/app/models/publication.rb
@@ -312,6 +312,8 @@ class Publication < ActiveRecord::Base
 
     pub_hash[:mesh_headings] = pubmed_hash[:mesh_headings] unless pubmed_hash[:mesh_headings].blank?
     pub_hash[:abstract] = pubmed_hash[:abstract] unless pubmed_hash[:abstract].blank?
+    pmc_id = pubmed_hash[:identifier].detect {|id| id[:type] == 'pmc'}
+    pub_hash[:identifier] << pmc_id if pmc_id
   end
 
   def set_last_updated_value_in_hash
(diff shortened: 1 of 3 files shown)


app/models/publication.rb
```diff
@@ -312,6 +312,8 @@ def add_any_pubmed_data_to_hash
 
     pub_hash[:mesh_headings] = pubmed_hash[:mesh_headings] unless pubmed_hash[:mesh_headings].blank?
     pub_hash[:abstract] = pubmed_hash[:abstract] unless pubmed_hash[:abstract].blank?
+    pmc_id = pubmed_hash[:identifier].detect {|id| id[:type] == 'pmc'} if pubmed_hash[:identifier]
+    pub_hash[:identifier] << pmc_id if pmc_id
```
darrenleeweber: Looks like `pubmed_hash[:identifier]` always exists (app/models/pubmed_source_record.rb:147)

peetucket: ahh, yes, it was blowing up a test when i didn't check for its existence but that is probably because the test data setup was not complete ... let me check again

app/models/pubmed_source_record.rb
```diff
@@ -149,7 +149,8 @@ def convert_pubmed_publication_doc_to_hash(publication)
     doi = publication.at_xpath('//ArticleId[@IdType="doi"]')
     doi = publication.at_xpath('//ELocationID[@EIdType="doi"]') unless doi.present? && doi.text.present?
     record_as_hash[:identifier] << { type: 'doi', id: doi.text, url: "#{Settings.DOI.BASE_URI}#{doi.text}" } if doi.present? && doi.text.present?
-
+    pmc = publication.at_xpath('//ArticleId[@IdType="pmc"]')
+    record_as_hash[:identifier] << { type: 'pmc', id: pmc.text } if pmc.present? && pmc.text.present?
```
darrenleeweber: `pmc` or `PMCID`, does this thing have a formal or conventional name that is widely used?  If not, no problem.  If it does, maybe we should use it here?  (Suggestion, not requirement.)

darrenleeweber: Looks like the PUBMED data uses 'pmc', so that's good enough for me.  Forget it.

peetucket: removed check for existence above and updated test that failed to ensure proper data input

PR #174, opened 2017-05-02
[WIP] add nlm citation format

peetucket:
Add the NLM as a citation format


peetucket:
No longer needed by SoM

PR #175, opened 2017-05-03
rename deployment names to dlss standards

peetucket:


peetucket:
In my experience, the rails_env needs to be set if the name of the deploy stage does not match the rails environment you are running it.  The cap deploy script uses this to call rake commands with the proper environment.  I am not sure if there is a tie-in to puppet, but confirmed it doesn't work unless it's set now (the stages used to be "development" and "production" before, which matched the rails environments).

darrenleeweber:
What is the Rails.env value in the config/deploy/stage.rb file?  I _guess_ it should be `production`.

peetucket:
yes, it is production.  it was set that way before (didn't change in this PR)

PR #176, opened 2017-05-04
remove broken rake tasks for updating hash and authorship, create a n…

peetucket:
…ew one which includes feedback on how its working and actually saves the pubs each time

actual code to do the updates is minimal (basically a find_each with a block that calls the rebuild method and a save) ... everything else is friendly output feedback, since the task is going to take a couple days to run in nohup mode, and having output to see where it is will be very helpful

### DIFF
--- a/lib/tasks/sul.rake
+++ b/lib/tasks/sul.rake
@@ -1,21 +1,50 @@
 
 namespace :sul do
-  desc 'rebuild authorship for selected pubs'
-  task update_contribs_in_hash: :environment do
-    #  publications = Publication.arel_table
-    #	Publication.where(publications[:pub_hash].
-    #    	matches("%okogiri%")).
-    #    	each {|publication| publication.rebuild_pub_hash }
-    Publication.find_each(&:rebuild_authorship)
-  end
-
-  desc 'rebuild authorship for selected pubs'
-  task rebuild_bad_hashes: :environment do
-    publications = Publication.arel_table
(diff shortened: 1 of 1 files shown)


lib/tasks/sul.rake
```diff
@@ -1,21 +1,29 @@
...
+      elapsed_time = current_time - start_time
+      avg_time_per_pub = elapsed_time/(index+1)
+      total_time_remaining = (avg_time_per_pub * (total_pubs-index)).floor
+      if index % output_each == 0 # provide some feedback every X pubs
+        puts "...#{current_time}: on publication #{number_with_delimiter(index+1)} of #{number_with_delimiter(total_pubs)} : ~ #{distance_of_time_in_words(start_time,start_time + total_time_remaining.seconds)} left"
+      end
+      pub.send(method)
+      pub.save
+    end
+    puts "Called #{method} for #{number_with_delimiter(total_pubs)} publications.  Ended at #{Time.now}"
```
darrenleeweber: You might need exception handling on each pub.send(method) and pub.save, so it can continue to process the next pub?  Occasional failures don't have to crash the entire batch.  But count the failures and stop if they exceed a reasonable threshold (some way to crash if everything is failing every time).

peetucket:
Added logging output to a file as well as to the screen.  Catch exceptions on the method call and the save, log ID of errored out pub and continue.  If more than 500 errors, halt execution completely.

PR #178, opened 2017-05-23
add a new method to repull pubmed record so we can update our source

peetucket:
also add a rake task to repull all pubmed records for given cap_profile_ids and then update the pub_hash

This is useful to refresh PMID based publications (for example, when we need to be sure we add PMCIDs to our source records where they have been added to PubMed after we first pulled)

### DIFF
--- a/app/models/pubmed_source_record.rb
+++ b/app/models/pubmed_source_record.rb
@@ -68,6 +68,20 @@ class PubmedSourceRecord < ActiveRecord::Base
     PubmedSourceRecord.import source_records
   end
 
+  # Retrieve this pubmed record from PubMed and update
+  # is_active, source_data and the source_fingerprint fields.
+  # Used to update the pubmed source record on our end (similar to .sciencewire_update)
+  # @return [Boolean] the return value from update_attributes!
+  def pubmed_update
+    pubmed_source_xml = PubmedClient.new.fetch_records_for_pmid_list pmid
+    pub_doc = Nokogiri::XML(pubmed_source_xml).xpath('//PubmedArticle')[0]
+    return false unless pub_doc
+    attrs = {}
+    attrs[:source_data] = pub_doc.to_xml
+    attrs[:source_fingerprint] = Digest::SHA2.hexdigest(pub_doc)
+    update_attributes! attrs
+  end
+
   def extract_abstract_from_pubmed_record(pubmed_record)
     pubmed_record.xpath('MedlineCitation/Article/Abstract/AbstractText').text
   end
--- a/lib/tasks/pubmed.rake
+++ b/lib/tasks/pubmed.rake
@@ -3,6 +3,71 @@ namespace :pubmed do
     @client ||= PubmedClient.new
   end
 
+  desc 'update pubmed source records and rebuild pub_hashes for publications for a given list of cap_profile_ids passed in a text file'
+  # filename should be a list of cap_profile_id, one per line with no header
+  # call with RAILS_ENV=production bundle exec rake pubmed:update_pubmed_source_records_for_cap_profile_ids['filename.txt']
+  task :update_pubmed_source_records_for_cap_profile_ids, [:filename] => :environment do |_t, args|
+    filename = args[:filename]
+    fail "filename is required." unless filename.present?
+    cap_profile_ids=File.readlines(filename)
+    logger = Logger.new(Rails.root.join('log', 'update_pubmed_source_records_for_cap_profile_ids.log'))
+    include ActionView::Helpers::DateHelper
+    $stdout.sync = true # flush output immediately
+    total_authors = cap_profile_ids.size
+    error_count = 0
+    success_count = 0
+    authors_not_found = 0
+    pubs_updated_count = 0
+    start_time = Time.now
+    max_errors = 10
+    message = "Started at #{start_time}.  Working on #{total_authors} authors."
+    puts message
+    logger.info message
+    cap_profile_ids.each_with_index do |cap_profile_id, index|
+      id = cap_profile_id.chomp
+      current_time = Time.now
+      elapsed_time = current_time - start_time
+      avg_time_per_author = elapsed_time/(index+1)
+      total_time_remaining = (avg_time_per_author * (total_authors-index)).floor
+      message = "...#{current_time}: on cap_profile_id #{id} : #{index+1} of #{total_authors} : ~ #{distance_of_time_in_words(start_time,start_time + total_time_remaining.seconds)} left"
+      puts message
+      logger.info message
+      begin
+        author = Author.find_by_cap_profile_id(id)
+        unless author.blank?
+          pubs = author.publications.where("pmid is not null OR pmid !=''")
+          message = ".....#{author.first_name} #{author.last_name} has #{pubs.size} publications with a PMID"
+          puts message
+          logger.info message
+          pubs.each do |pub|
+            result = pub.update_from_pubmed
+            pubs_updated_count += 1 if result
+          end
+          success_count += 1
+        else
+          authors_not_found += 1
+          message = "*****cap_profile_id #{id} not found"
+          puts message
+          logger.error message
+        end
+      rescue  => e
+        message = "*****ERROR on cap_profile_id #{id}: #{e.message}"
+        puts message
+        logger.error message
+        error_count += 1
+      end
+      if error_count > max_errors
+        message = "Halting: Maximum number of errors #{max_errors} reached"
+        logger.error message
+        raise message
+      end
+    end
+    end_time = Time.now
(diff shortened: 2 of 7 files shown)


lib/tasks/pubmed.rake
```diff
@@ -3,6 +3,76 @@ namespace :pubmed do
     @client ||= PubmedClient.new
   end
 
+  desc 'update pubmed source records and rebuild pub_hashes for publications for a given list of cap_profile_ids passed in a text file'
+  # filename should be a list of cap_profile_id, one per line with no header
+  # call with RAILS_ENV=production bundle exec rake pubmed:refresh_source_records_for_cap_profile_ids['filename.txt']
+  task :refresh_source_records_for_cap_profile_ids, [:filename] => :environment do |_t, args|
+    filename = args[:filename]
+    fail "filename is required." unless filename.present?
+    cap_profile_ids=File.readlines(filename)
```
darrenleeweber: Validate they exist in our authors.

spec/models/pubmed_source_record_spec.rb
```diff
@@ -145,6 +145,18 @@ def check_author_hash(author)
...
 
+  describe '.pubmed_update' do
+    it 'updates the :source_data field' do
+      record = described_class.get_pubmed_record_from_pubmed(pmid_created_1999)
+      original_source_data = record.source_data
+      expect(record.source_data).not_to be_empty
+      record.source_data = ''
+      expect(record.source_data).to be_empty
+      expect(record.pubmed_update).to be true
+      expect(record.source_data).to eq original_source_data
```
darrenleeweber: This seems to be an artifact of the spec fixtures or something, because an update should change something.  Can the fixture data be adjusted so that something changes and the spec can test for a specific change, everything else remains the same?

app/models/pubmed_source_record.rb
```diff
@@ -68,6 +68,19 @@ def self.get_and_store_records_from_pubmed(pmids)
...
+  # Used to update the pubmed source record on our end (similar to .sciencewire_update)
+  # @return [Boolean] the return value from update_attributes!
+  def pubmed_update
+    pubmed_source_xml = PubmedClient.new.fetch_records_for_pmid_list pmid
+    pub_doc = Nokogiri::XML(pubmed_source_xml).xpath('//PubmedArticle')[0]
+    attrs = {}
+    attrs[:source_data] = pub_doc.to_xml
+    attrs[:source_fingerprint] = Digest::SHA2.hexdigest(pub_doc)
+    update_attributes! attrs
+  end
```
darrenleeweber: This looks OK.  Just a few thoughts or questions arise from this:
- this seems to use a high-level command that accepts a list of PMID values, or a single value is OK too?
- is there any Pubmed HEAD request that contains cache invalidation data that we can track to detect updates?  If there is, do we already have the data in our PubmedSourceRecord data (if not, then it's not going to work, unless we start to add it).  If there is, we could run a rake task over all the Pubmed data, say monthly or something (no less than weekly because their update schedule is weekly anyway).

lib/tasks/pubmed.rake
peetucket: Do you mean validate that the cap_author_ids are valid?  That is checked later (invalid cap author ids are just skipped in the loop rather than validating all upfront)

spec/models/pubmed_source_record_spec.rb
peetucket: the update just reverts back to the original data (after i set it to blank)...not sure i understand what the suggestion is

app/models/pubmed_source_record.rb
peetucket: * a single value should work, i tested it
* that is a good idea, I'll make a ticket for it, but outside the scope of this rather limited change for which I just want to run a specific set of author IDs

spec/models/pubmed_source_record_spec.rb
darrenleeweber: Can we mock the update so it gets back a new PMC, for example?

app/models/pubmed_source_record.rb
darrenleeweber: OK

lib/tasks/pubmed.rake
darrenleeweber: I'd modify the call stack on this rake task.  First create a rake task that accepts just one PMID.  Then create a rake task that accepts one CAP profile id (it can then call the PMID task).  Then create rake task that takes a file of profile-ids.  The break-down of the rake tasks will help to reduce the complexity of this one and provide more flexibility at the rake command line.

darrenleeweber: - :refresh_source_records_for_cap_profile_id
  - get all the PMIDs for this author and then call:
- :refresh_source_records_for_pmid

darrenleeweber: Also the name of this rake task might be too generic, as in
- `refresh_source_records_for_cap_profile_ids` suggests all the source records, not just Pubmed

Rename it?  Maybe `update_pubmed_source_records_for_cap_profile_ids`?

peetucket: I'd rather implement the incremental stuff as methods than have a rake task calling other rake tasks (which complicates logging into a single file).  And I'm not sure of the utility of a single author (which could still be run as this same rake task with just a single ID as input).  But I will implement the pubmed update into a single method.

spec/models/pubmed_source_record_spec.rb
peetucket: I can't get mocking to work (getting an ```ActiveRecord::StaleObjectError``` error from the database on the "update_attributes!" call in the model).  I am reluctant to sink too much time into testing this new method.

peetucket: Figured out a way to stub the whole thing .

spec/models/publication_spec.rb
```diff
@@ -252,6 +252,46 @@
...
+      publication.pmid = pmid
+      source_data = '<PubmedArticle><MedlineCitation Status="Publisher" Owner="NLM"><PMID Version="1">1</PMID><OriginalData/></PubmedArticle>'
+      new_source_data = '<PubmedArticle><MedlineCitation Status="Publisher" Owner="NLM"><PMID Version="1">1</PMID><SomeNewData/></PubmedArticle>'
+      pubmed_record = PubmedSourceRecord.create(pmid: pmid, source_data: source_data)
+      allow(PubmedSourceRecord).to receive(:find_by_pmid).with(pmid).and_return(pubmed_record)
+      expect(pubmed_record.source_data).to be_equivalent_to source_data
+      allow_any_instance_of(PubmedClient).to receive(:fetch_records_for_pmid_list).with(pmid).and_return(new_source_data)
+      expect(publication.pmid).to_not be_nil
+      expect(publication.update_from_pubmed).to be true
+      expect(PubmedSourceRecord.find_by_pmid(pmid).source_data).to be_equivalent_to new_source_data
```
darrenleeweber: Also check that some specific values in the PubMed update make their way into the PubHash.

spec/models/publication_spec.rb
```diff
@@ -252,6 +252,46 @@
...
+      publication.sciencewire_id = sciencewire_id
+      source_data = '<?xml version="1.0"?><PublicationItem><PublicationItemID>1</PublicationItemID><Title>Some Title</Title><Abstract/></PublicationItem>'
+      new_source_data = '<?xml version="1.0"?><PublicationItem><PublicationItemID>1</PublicationItemID><Title>Some New Title</Title><Abstract/></PublicationItem>'
+      sw_record = SciencewireSourceRecord.create(sciencewire_id: sciencewire_id, source_data: source_data)
+      allow(SciencewireSourceRecord).to receive(:find_by_sciencewire_id).with(sciencewire_id).and_return(sw_record)
+      expect(sw_record.source_data).to be_equivalent_to source_data
+      allow_any_instance_of(ScienceWireClient).to receive(:get_sw_xml_source_for_sw_id).with(sciencewire_id).and_return(Nokogiri::XML(new_source_data).xpath('//PublicationItem').first)
+      expect(publication.sciencewire_id).to_not be_nil
+      expect(publication.update_from_sciencewire).to be true
+      expect(SciencewireSourceRecord.find_by_sciencewire_id(sciencewire_id).source_data).to be_equivalent_to new_source_data
```
darrenleeweber: Also check that some specific values in the SW update make their way into the PubHash.

peetucket: but that wouldn't happen until the rebuild_pub_hash step, which i would argue is not what this tests is looking for

I checked and sadly "rebuild_pub_hash" is one of the pending tests... and while I think it would be great to add tests for that, I don't think I have the time to do that before needing to use this code.  I did a ticket for it though : https://github.com/sul-dlss/sul_pub/issues/180

ditto for pubmed source record test

peetucket: oops, never mind...the publication level method is calling rebuild_pub_hash...hmmm

PR #181, opened 2017-05-31
updates to pubmed update rake task

peetucket:
* move exception handling check to the publication update so that a single publication update failure does not block all other pubs from that author


lib/tasks/pubmed.rake
```diff
@@ -18,6 +18,7 @@ namespace :pubmed do
     success_count = 0
```
darrenleeweber: rename generic `success_count` to `authors_found`?

peetucket:
good suggestion, thanks -- fixed and rebased ... will merge after tests complete

PR #182, opened 2017-06-19
standardize cap deployment names

peetucket:


peetucket:
Sorry, missed all that stuff.  Fixed and made a change so the default branch deployed is master and not the currently checked out branch (which is more dangerous to default to).

PR #219, opened 2017-09-26
[WIP] basic version of WebOfScience::Client class

darrenleeweber:
This pulls from Peter's script and basically pulls up the various
components that went into the Sciencewire client and request classes.

Once fully conformant to the new API, including limits and such, I
believe this class signature would be sufficient to stand in for the existing
ScienceWire client (and some related classes).

### Related code
- https://github.com/sul-dlss/sul_pub/blob/rialto-search/bin/faculty_wos_search.rb#L28-L110

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -30,6 +30,7 @@ gem 'delayed_job_active_record'
 gem 'dotiw'
 gem 'faraday'
 gem 'httpclient', '~> 2.7'
+gem 'net-http-persistent'
 gem 'high_voltage'
 # To use Jbuilder templates for JSON
 gem 'jbuilder'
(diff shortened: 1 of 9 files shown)


darrenleeweber:
We've had some discussion in slack around placing this in the `lib` path, where the existing client code exists.  My preference is to put it there to be consistent (use a consistent pattern for the location of the API client code), but I'm not wanting to block this work based on this preference alone.

darrenleeweber:
I've also noted that the existing app models use the `SourceRecord` suffix and I would like to see any new models do so as well.  IMO, this is a blocker to accept this, but I'll defer that to a later review once the work is ready.

atz:
This initial work does not engage the question of persistence, only scaffolding a client around the new API.  So there is no `SourceRecord` yet, which is where the other APIs have their ActiveRecord models.

darrenleeweber:
While adapting to the SOAP response, this approach is proving to be more difficult that using the savon SOAP client.  We may come back to this approach later when WoS moves to a REST/JSON API service.

Issue #171, opened 2017-04-17
Analysis of work required to use new ScienceWire API


peetucket:
Clarivate (the new company that provides the Web of Science/ScienceWire API) has updated their API and is requesing that we use it.  The new API documentation is attached, and it appears that 

1. There is a new authentication mechanism - you post a username/password to get an authentication key you use for a session.

2. The calls have a different data structure both for the request and the response.

3. The response no longer includes a PMID for publications.  There is a separate call to determine if a publication has a PMID (though you check multiple publications at once).  This implies we would have an extra call beyond what we do now.

We should do a bit more analysis to estimate the work needed to update sul_pub to work with the new API and to see if this would cause any problems with current functionality.

peetucket:
updated API documentation: 

[WoSWS20150707.pdf](https://github.com/sul-dlss/sul_pub/files/926876/WoSWS20150707.pdf)

peetucket:
I have a username/password combo we can use for the new API (will not post in ticket).  I have tested with a single query and it works.

darrenleeweber:
The WoS temporary auth credential is available in shared_configs, see
- https://github.com/sul-dlss/shared_configs/tree/sul-pub-laptop

darrenleeweber:
Peter's initial code to explore the new WoS API is in the `rialto-search` branch, i.e.
- https://github.com/sul-dlss/sul_pub/blob/rialto-search/bin/faculty_wos_search.rb

PR #222, opened 2017-09-27
Savon WSDL client for WOS SOAP API

darrenleeweber:
This explores using the ruby savon gem for SOAP.

- it can authenticate to the new SOAP API for Web of Science
- it initializes a search SOAP client

Use in `rails c`
```ruby
[2] pry(main)> wos = WosClient.new(Settings.WOS.AUTH_CODE);
I, [2017-09-27T08:58:32.938193 #12354]  INFO -- : SOAP request: http://search.webofknowledge.com/esti/wokmws/ws/WOKMWSAuthenticate
I, [2017-09-27T08:58:32.938383 #12354]  INFO -- : Authorization: Basic {{SECRET_SNIPPED}}, SOAPAction: , Content-Type: text/xml;charset=UTF-8, Content-Length: 397
I, [2017-09-27T08:58:33.238701 #12354]  INFO -- : SOAP response (status 200)
[3] pry(main)> wos.session_id
=> "1F28d8jTIRdrY9Oi4aF"
[4] pry(main)> wos.auth.class
=> Savon::Client
[5] pry(main)> wos.search.class
=> Savon::Client
[6] pry(main)> wos.search.operations
=> [:cited_references_retrieve, :related_records, :cited_references, :retrieve, :search, :citing_articles, :retrieve_by_id]
[7] pry(main)> wos.auth.operations
=> [:authenticate, :close_session]
```

I use this with a `config/settings.local.yml` file that contains the actual auth-token.

TODO:
- [x] specs on the `WosClient.search` functionality (see #223)

Related issues:
- https://github.com/sul-dlss/sul_pub/issues/171

### DIFF
--- a/spec/lib/wos_client_spec.rb
+++ b/spec/lib/wos_client_spec.rb
@@ -1,0 +1,65 @@
+# http://savonrb.com/version2/testing.html
+# require the helper module
+require 'savon/mock/spec_helper'
+
+describe WosClient do
+  include Savon::SpecHelper
+
+  # set Savon in and out of mock mode
+  before(:all) { savon.mock!   }
+  after(:all)  { savon.unmock! }
+
+  let(:wos_auth) { 'secret' }
+  let(:wos_client) { described_class.new(wos_auth) }
+  let(:auth_xml) { File.read('spec/fixtures/wos_client/authenticate.xml') }
+
+  describe '#new' do
+    it 'works' do
+      expect(wos_client).to be_an described_class
+    end
+  end
+
+  describe '#auth' do
+    it 'works' do
+      result = wos_client.auth
+      expect(result).to be_an Savon::Client
+    end
+  end
+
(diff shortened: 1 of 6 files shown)


darrenleeweber:
My only thoughts on that client are around how to close the session automatically when the object is detroyed (so it doesn’t leave open sessions dangling).  The problem can be pushed up to consumers to explicitly call the close_session.  There are some patterns for auto-closing stuff that generally work with class methods, like `File.open` can take a block and then do stuff after the yield is done to clean up file handlers.  I’ll put this commentary in a tech-debt issue, it could be addressed later.

spec/lib/wos_client_spec.rb
```diff
@@ -0,0 +1,67 @@
+require 'spec_helper'
```
atz: Don't need this line because it is done in the `.rspec` file.

PR #223, opened 2017-09-27
Web of Science queries

### LINKED ISSUE #197: Write new science_wire/api classes for making calls

darrenleeweber:
Fix #197 

Provides query functionality for all the new WoS SOAP API endpoints

TODO:
- [x] rebase on master after #222 merged
- [x] parse the publication results
- [x] add specs
- [x] squash commits

Related issues:
- https://github.com/sul-dlss/sul_pub/issues/171

## Integration test
Use `bundle exec rails c` and, when you have the WoS credentials in place, run
```
wos_queries = WosQueries.new(WosClient.new(Settings.WOS.AUTH_CODE, Settings.WOS.LOG_LEVEL.to_sym))
records = wos_queries.search_by_name('Altman, Russ')
records.count #=> 464 (matched the WoS total records available)
records.uids # an array of the UID values
records.print # long listing of all the records XML
```


lib/wos_queries.rb
```diff
@@ -0,0 +1,184 @@
+
+# Queries on the Web of Science (or Web of Knowledge)
+class WosQueries
+
+  # Default database 'WOK' is an umbrella for everything
+  DATABASE = 'WOK'.freeze
```
atz: Don't need a constant here, just put the string as the default in `initialize`.  That makes it part of the YARD.

lib/wos_queries.rb
```diff
@@ -0,0 +1,184 @@
...
+
+  QUERY_LANGUAGE = 'en'.freeze
+
+  # limit the start date when searching for publications, format: YYYY-MM-DD
+  START_DATE = '1970-01-01'.freeze
+
+  attr_reader :wos_client
+
+  # @param wos_client [WosClient] a Web Of Science client
+  # @param database [String] a WOS database identifier (default 'WOS')
```
atz: default now `WOK`

darrenleeweber:
I'm stuck trying to work around what seems to be a bug in Nokogiri, reported at
- https://github.com/sparklemotion/nokogiri/issues/1677

darrenleeweber:
I've worked around the Nokogiri issues related to immutability, by using the `Document.dup` deep copy method.

lib/wos_queries.rb
```diff
@@ -0,0 +1,279 @@
...
+        queryLanguage: QUERY_LANGUAGE,
+        retrieveParameters: retrieve_parameters(options: retrieve_options)
+      }
+    end
+
+    # @param uid [String] a WOS UID
+    # @return [Hash] citingArticles parameters
+    def citing_articles_params(uid)
+      {
+        databaseId: @database,
```
atz: Probably should have an `attr_reader :database` and use it in these methods instead of referencing the instance variable.

lib/wos_queries.rb
```diff
@@ -0,0 +1,279 @@
...
+        uid: uid,
+        timeSpan: time_span,
+        queryLanguage: QUERY_LANGUAGE,
+        retrieveParameters: retrieve_parameters
+      }
+    end
+
+    # @param uids [Array<String>] a list of WOS UIDs
+    # @return [Hash] retrieveById parameters
+    def retrieve_by_id_params(uids)
```
atz: A ton of repetition here in the `params` methods.  Would it make sense to just have a `base_params` method (that looks exactly like the body of this one) and then invoke: 
```ruby
base_params.merge(different_key: different_val)
```
in the places where individual key/val pairs need to be added or overridden?  This would put the focus on what is unique to each set of params rather than repeating all the things that are in common.

spec/lib/wos_records_spec.rb
```diff
@@ -0,0 +1,199 @@
...
+
+    it 'works' do
+      expect(rec_nodes).not_to be_nil
+    end
+  end
+
+  describe '#to_xml' do
+    it 'works with encoded records' do
+      result = wos_records_encoded.to_xml
+      expect(result).to be_an String
```
atz: Can we get a stronger expectation here than just "gimme a String"?  Since we are dealing with encoding, it is sorta important that not all strings are equivalent.

spec/lib/wos_records_spec.rb
```diff
@@ -0,0 +1,199 @@
...
+    end
+  end
+
+  # PRIVATE
+
+  describe '#decode_records' do
+    it 'works' do
+      result = wos_records_encoded.send(:decode_records)
+      expect(result).to be_an String
```
atz: Same.

lib/wos_records.rb
```diff
@@ -0,0 +1,128 @@
...
+  # Find duplicate WoS records
+  # @param record_setB [WosRecords]
+  # @return [Nokogiri::XML::NodeSet] duplicate records
+  def duplicate_records(record_setB)
+    duplicates = []
+    uid_intersection = duplicate_uids(record_setB)
+    unless uid_intersection.empty?
+      # create a new set of records, use a philosophy of immutability
+      # Nokogiri::XML::NodeSet enumerable methods do not return new objects
+      # Nokogiri::XML::Document.dup is a deep copy
```
atz: Note: the hangup with Nokogiri is unsurprising and doesn't require a bunch of commentary.  If you want to build a nodeset from existing nodes without retaining any connection back to the parent document (which all nodes have), then you can round-trip the node through serialization `.to_xml` and back into a new Nokogiri XML node/fragment.  This would reduce the memory utilization by not copying the whole document.

lib/wos_queries.rb
```diff
@@ -0,0 +1,279 @@
...
+
+  # @param uids [Array<String>] a list of WOS UIDs
+  # @return [WosRecords]
+  def retrieve_by_id(uids)
+    retrieve_by_id_collator(uids)
+  end
+
+  # @param name [String] a CSV name pattern: {last name}, {first_name} [{middle_name} | {middle initial}]
+  # @return [WosRecords]
+  def search_by_name(name)
```
atz: Why do you have these 5 public methods that are effectively aliases for private ones?  Put the body of the private methods here and remove the "collator" methods.

lib/wos_records.rb
atz: Also, it is not a bug.

spec/lib/wos_records_spec.rb
darrenleeweber: Maybe an XSD validation for the record data?

lib/wos_records.rb
```diff
@@ -0,0 +1,128 @@
...
+  # @return uid_nodes [Nokogiri::XML::NodeSet] the rec_nodes UID nodes
+  def uid_nodes
+    rec_nodes.search('UID')
+  end
+
+  # Find duplicate WoS UID values
+  # @param record_setB [WosRecords]
+  # @return [Set] duplicate WoS UID values
+  def duplicate_uids(record_setB)
+    uids.to_set.intersection(record_setB.uids.to_set)
```
atz: `Array` already has an [intersection operator](https://ruby-doc.org/core-2.2.0/Array.html#method-i-26) in Ruby, the `&`.  So without the need to convert `to_set`, this should be `uids & record_setB.uids`.  Intersection also deduplicates multiples and preserves order!

lib/wos_queries.rb
darrenleeweber: I'll try to move some stuff where this is done in #242 already, into this PR.  This pattern of using a collator emerged from doing one request first and then another and another and then noticing that the collators themselves could be refactored, until they became so small that they, well, don't need to exist any more.

spec/lib/wos_records_spec.rb
atz: Or at least expect it not to contain `&lt;` or something.

lib/wos_records.rb
darrenleeweber: OK, I'll try that, good tip.

spec/lib/wos_records_spec.rb
atz: I don't think full XSD is required.  Just look for something that you expect the encoding to adjust (or not adjust) and target it.

lib/wos_queries.rb
darrenleeweber: Done

spec/lib/wos_records_spec.rb
darrenleeweber: Done

lib/wos_records.rb
darrenleeweber: We can discuss this in person, if necessary, as I like it as it is.  I hope it's not a blocker.  The round trip via `.to_xml` will not gain anything, as the doc only contains records anyway.  The explanation fills a gap in the Nokogiri documentation and I believe will serve as a note to prevent later refactoring that could easily mistakenly introduce violations of the immutability desired (which should be caught by specs, but might not be understood without this commentary).

lib/wos_queries.rb
darrenleeweber: Done

lib/wos_records.rb
darrenleeweber: Yep, this works for me:
```ruby
> x = [*1..10]
=> [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
> y = [*1..10].reject {|i| i % 2 == 0 }.reverse
=> [9, 7, 5, 3, 1]
> x & y
=> [1, 3, 5, 7, 9]
> x.reverse & y
=> [9, 7, 5, 3, 1]
```

darrenleeweber: Done

darrenleeweber:
Review changes pushed in separate commits to clearly identify how each commit addressed the review concerns.

lib/wos_queries.rb
darrenleeweber: If you look closely at it, there are slight differences in each set of params.  Given that we know the request XML is order-sensitive (right?  why? weird, if so, but yea ...), doesn't this make it difficult to use a Hash.merge strategy with any confidence?  I think we should be glad that savon allows us to use Hash params.  These methods have already been refactored a bit to allow for some existing use-case flexibility; we might discover more as we go.  I'm open to exploring whether or not a Hash.merge strategy might work and will try it out soon.  If it's not a blocker right now, maybe we can move on and adapt as we go?

atz: Yes, I acknowledged those slight differences in my comment. My intent is to have a common starting point, then merge/reject from there.  This would make it obvious what each of those differences is, I.E., what is special about *this method*.  

The ordered XML thing does not apply since we are still giving the Savon gem a Hash.  It figures out the order now, and it would figure out the order the same way, since we would be passing it exactly the same param.  This is just about how we build the hash and how much we repeat ourselves.  In general, this codebase has suffered from lack of facility manipulating ruby Hash, so I would like to start repairing that.

lib/wos_records.rb
atz: The round trip is what divorces it from the parent document (since it becomes just a String value, with no ability to point to location of the originating document in Ruby memory).  The separation is the basis of your motivation to `.dup`.  I agree this may benefit from better docs, but it is not unexpected behavior.

spec/lib/wos_records_spec.rb
```diff
@@ -0,0 +1,216 @@
...
+  describe '#each' do
+    it 'yields WoS record elements' do
+      wos_records_encoded.all? { |rec| expect(rec).to be_an Nokogiri::XML::Element }
+    end
+  end
+
+  describe 'Enumerable mixin' do
+    it 'responds to Enumerable.instance_methods' do
+      methods = Enumerable.instance_methods
+      methods.each { |method| expect(wos_records_encoded).to respond_to(method) }
```
atz: There is no need to enumerate methods individually, because that is how Ruby inheritance works.  Please replace this spec with `described_class.is_a?(Enumerable)` or consider removing it.

atz:
Thanks for the responses.  

The `.dup` and Hash composition comments are not blockers, as long as you are open to their revision in the future.  In that regard, I would as: are you comfortable enough with the tests coverage to permit that?  

Would appreciate attention to one outstanding item re: `Enumerable` testing.

spec/lib/wos_records_spec.rb
darrenleeweber: Doesn't work, i.e.
```
  1) WosRecords Enumerable mixin responds to Enumerable.instance_methods
     Failure/Error: expect(described_class).to be_an Enumerable
       expected WosRecords to be a kind of Enumerable
     # ./spec/lib/wos_records_spec.rb:74:in `block (3 levels) in <top (required)>'
```
```
  1) WosRecords Enumerable mixin responds to Enumerable.instance_methods
     Failure/Error: expect(described_class).to be_kind_of Enumerable
       expected WosRecords to be a kind of Enumerable
     # ./spec/lib/wos_records_spec.rb:74:in `block (3 levels) in <top (required)>'
```
Leave it as is?

lib/wos_records.rb
darrenleeweber: I get how the `.to_xml` works.  What I'm saying is that the doc just contains REC elements, so there's nothing else duped by the `doc.dup` anyway, so there's not much gained from all the extra serialization transforms involved in the `.to_xml`; it will just slow things down and that might be worse that a faster `doc.dup` when collecting 100's of records.

lib/wos_queries.rb
darrenleeweber: I'll take another look at this.

spec/lib/wos_records_spec.rb
darrenleeweber: Never mind, got it working on an instance object.

darrenleeweber: Done.

lib/wos_records.rb
atz: I agree there may be a computational tradeoff for the benefit of smaller memory footprint.  We can defer on this one for later.

PR #226, opened 2017-09-29
CI was running twice because of system fork out of Rakefile

atz:
Reduce cruft, improve accuracy.

The purpose was to avoid running CI in production, but that can be accomplished without a fork.

Do `db:setup` in Travis' `before_script`.  Now the `Rails.env.production?` check is unnecessary, because `RSpec` won't even be part of the bundle on a production system, so all it would do is run rubocop then fail harmlessly on a NameError.

This incorporates the changes from @mjgiarlo's gist:
https://gist.github.com/mjgiarlo/ee079cd476b94066886f6f541e624a5b

And resolves the double-running of spec and rubocop.

NOTE: Explicitly (re)defining the `:spec` task was part of the mechanism of duplication.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -9,22 +9,13 @@ Sulbib::Application.load_tasks
 task default: [:ci]
 
 desc 'Continuous integration task run on travis'
-task ci: [:environment, :rubocop] do
-  if Rails.env.test?
-    Rake::Task['db:setup'].invoke # not db:migrate, use the dang schema!
-    Rake::Task['spec'].invoke
-  else
-    system 'rake ci RAILS_ENV=test'
-  end
-end
+task ci: [:rubocop, :spec]
 
 begin
   require 'rspec/core/rake_task'
-  RSpec::Core::RakeTask.new
-  namespace :spec do
-    desc 'run only data-integration tests against live ScienceWire (excluded by default)'
-    RSpec::Core::RakeTask.new('data-integration') { |t| t.rspec_opts = '--tag data-integration' }
-  end
+
+  desc 'Run only data-integration tests against live ScienceWire (excluded by default)'
+  RSpec::Core::RakeTask.new('spec_with_data_integration') { |t| t.rspec_opts = '--tag data-integration' }
 rescue LoadError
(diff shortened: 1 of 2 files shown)


darrenleeweber:
This change also means that dev-laptops must be explict about using `RAILS_ENV=test`, right?  That's a bit of overhead we don't want, so we need some way of setting that for the spec (or rspec) task (if not auto-set already).

atz:
No, if the dev doesn't care, CI doesn't need to care either.  I.E., all the tests should still pass running against the `development` DB, if that's what you want.

Rakefile
```diff
@@ -10,12 +10,9 @@ task default: [:ci]
...
 desc 'Continuous integration task run on travis'
 task ci: [:environment, :rubocop] do
-  if Rails.env.test?
-    Rake::Task['db:setup'].invoke # not db:migrate, use the dang schema!
-    Rake::Task['spec'].invoke
-  else
-    system 'rake ci RAILS_ENV=test'
-  end
+  raise 'Refusing to run CI with RAILS_ENV=production' if Rails.env.production?
+  Rake::Task['db:setup'].invoke # not db:migrate, use the dang schema!
```
darrenleeweber: Could we explicitly set the `ENV['RAILS_ENV'] = test` here or does it have to be set in the shell and that's why the existing code calls back to `system`?

darrenleeweber:
I don't want the development database to get cleaned, so, yea, I do care about the RAILS_ENV

Rakefile
atz: Yeah, it doesn't work like that, because of timing issues.  Since the rails environment is already initialized by this point, changing the ENV does **not** reinitialize rails.

atz:
Then don't run `rake ci` in the wrong environment, which is literally the task for CI.  I don't think there is anything difficult here.  We have tests:
- `bundle exec rake spec`
- `bundle exec rspec`

These are not protected by anything that was attempted by the fork, and still hit the DB, regardless of environment.  

A developer also has rubocop tasks and local executable, so literally all functionality required is available.

PR #235, opened 2017-10-03
Less callback hell

atz:
Use active record for good, not evil.

### DIFF
--- a/app/models/author.rb
+++ b/app/models/author.rb
@@ -45,7 +45,10 @@ class Author < ActiveRecord::Base
 
   # @return [Array<Integer>] ScienceWireIds for approved publications
   def approved_sciencewire_ids
-    publications.approved.with_sciencewire_id.pluck(:sciencewire_id).uniq
+    publications.where("contributions.status = 'approved'")
+                .where.not(sciencewire_id: nil)
+                .pluck(:sciencewire_id)
+                .uniq
   end
 
   has_many :contributions, dependent: :destroy, after_add: :contributions_changed_callback, after_remove: :contributions_changed_callback do
@@ -68,16 +71,7 @@ class Author < ActiveRecord::Base
   def contributions_changed_callback(*_args)
   end
 
-  has_many :publications, through: :contributions do
-    def approved
-      where("contributions.status='approved'")
-    end
-
-    def with_sciencewire_id
-      where(Publication.arel_table[:sciencewire_id].not_eq(nil))
-    end
-  end
-
+  has_many :publications, through: :contributions
   has_many :approved_sw_ids, -> { where("contributions.status = 'approved'") }, through: :contributions,
                                                                                 class_name: 'PublicationIdentifier',
                                                                                 source: :publication_identifier,
--- a/app/models/author.rb
+++ b/app/models/author.rb
@@ -68,7 +68,7 @@ class Author < ActiveRecord::Base
       c.assign_attributes contribution_hash.merge(publication_id: publication.id)
       if c.persisted?
         c.save
-        publication.contributions_changed_callback
+        publication.pubhash_needs_update!
       else
         self << c
       end
--- a/app/models/author.rb
+++ b/app/models/author.rb
@@ -17,22 +17,9 @@ class Author < ActiveRecord::Base
   end
 
   # Provide consistent API for Author and AuthorIdentity
-  # @return [String] preferred_first_name
-  def first_name
-    preferred_first_name
-  end
-
-  # Provide consistent API for Author and AuthorIdentity
-  # @return [String] preferred_middle_name
-  def middle_name
-    preferred_middle_name
-  end
-
-  # Provide consistent API for Author and AuthorIdentity
-  # @return [String] preferred_last_name
-  def last_name
-    preferred_last_name
-  end
+  alias_attribute :first_name, :preferred_first_name
+  alias_attribute :middle_name, :preferred_middle_name
+  alias_attribute :last_name, :preferred_last_name
 
   # Provide consistent API for Author and AuthorIdentity
   # The default institution is set in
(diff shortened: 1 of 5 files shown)


app/models/author.rb
```diff
@@ -17,22 +17,9 @@ def alternative_identities
...
-  end
-
-  # Provide consistent API for Author and AuthorIdentity
-  # @return [String] preferred_last_name
-  def last_name
-    preferred_last_name
-  end
+  alias_attribute :first_name, :preferred_first_name
+  alias_attribute :middle_name, :preferred_middle_name
+  alias_attribute :last_name, :preferred_last_name
```
darrenleeweber: This seems to lose the yard docs, can they be retained?

app/models/author.rb
```diff
@@ -58,7 +45,10 @@ def end_date
 
   # @return [Array<Integer>] ScienceWireIds for approved publications
   def approved_sciencewire_ids
-    publications.approved.with_sciencewire_id.pluck(:sciencewire_id).uniq
+    publications.where("contributions.status = 'approved'")
+                .where.not(sciencewire_id: nil)
+                .pluck(:sciencewire_id)
+                .uniq
   end
```
darrenleeweber: Why is this an improvement?

atz: Because it doesn't rely on the method defined in callback (using `arel_table`) and is just basic ActiveRecord.  The formatting is not the important part, if that is what you are responding to.

app/models/author.rb
```diff
@@ -81,16 +71,7 @@ def build_or_update(publication, contribution_hash = {})
...
-  has_many :publications, through: :contributions do
-    def approved
-      where("contributions.status='approved'")
-    end
-
-    def with_sciencewire_id
-      where(Publication.arel_table[:sciencewire_id].not_eq(nil))
-    end
-  end
-
```
darrenleeweber: I can't approve this change without doing a git grep on all the possible uses of this.  Given that it's not broken and this entails more work for me to review it, why should it be done in this sprint?  Given my general lack of confidence in this app and it's specs at this time, I'm reluctant to change things that are not broken and spend the time on fixing an app that might be entirely replaced by the RIALTO project sprints.

app/models/publication.rb
```diff
@@ -159,13 +140,9 @@ def build_from_sciencewire_hash(new_sw_pub_hash)
     self
   end
 
-  def build_from_pubmed_hash(new_pubmed_pub_hash)
-    self.pub_hash = new_pubmed_pub_hash
-    self
-  end
```
darrenleeweber: Why is this entirely removed?  What replaces this?  How do you know it's not called anywhere?

app/models/author.rb
atz: They are no longer necessary, because the knowledge about these is the same as the knowledge about those ActiveRecord attributes.  You know these are strings because the schema defines them as as strings.  Also, repeating `Provide consistent API for Author and AuthorIdentity` is not informative.  What value do think is lacking?

spec/lib/science_wire_harvester_spec.rb
```diff
@@ -384,15 +384,15 @@ def setup(threshold)
...
+      science_wire_harvester.send(:harvest_sw_pubs_by_wos_id_for_author, 'gorin', ['000224492700003'])
       auth.reload
       expect(auth.publications.count).to eq(1)
       expect(SciencewireSourceRecord.count).to eq(1)
       expect(PubmedSourceRecord.count).to eq(0)
       # Harvest the same publication and it should not duplicate.
       expect(science_wire_harvester).not_to receive(:create_contrib_for_pub_if_exists_by_author_ids)
       expect(science_wire_harvester).not_to receive(:create_or_update_pub_and_contribution_with_harvested_sw_doc)
-      science_wire_harvester.harvest_sw_pubs_by_wos_id_for_author('gorin', ['000224492700003'])
+      science_wire_harvester.send(:harvest_sw_pubs_by_wos_id_for_author, 'gorin', ['000224492700003'])
```
darrenleeweber: Why is this an improvement?

atz: As the commit message describes, this is effectively an internal method: nothing outside the class calls it except for the test.  Obviously the test should not dictate the design of the class, so this allows the method to be marked `private` and still use the same tests.

app/models/publication.rb
atz: Read the commit message.  I know it's not called because I located the couple places that called it.  Because the pubmed case doesn't require any special logic, it doesn't need an extra method to accomplish what can be done with just basic initialization or assignment, like:
```ruby
obj = Publication.new(pub_hash: new_hash1)
obj.pub_hash = new_hash2
```

app/models/author.rb
atz: I already did that analysis, so I don't understand why it would be prohibitive for you.  There is no way I want to rely on anti-pattern methods defined dynamically per object in AR callbacks if we can avoid it.

atz: The output of `git grep with_sciencewire_id` is empty, if that what you are looking for.

atz:
Split off several of the commits to separate PRs and rebased.  111-line delta now.

darrenleeweber:
Can split off d23095f5ab3286e3151ea803c9655393c1754fef for a focused PR on that alone.

darrenleeweber:
It looks like de7e3c3d65a89aa7425b2ce89dbe81051e1902ab could be a self-contained PR.

darrenleeweber:
It looks like ea31cbb5560441e26502b6d5b74db74dcfb79d9a could be a self-contained PR.

atz:
A separate PR per commit is not productive or necessary.  Related commits _belong on the same PR_.  There are only **35** new or edited lines left here, the rest is deletion.  That is well within what can reasonably be expected to be reviewed on a PR.

@darrenleeweber: Click the commits tab, review each commit contents, and complain when you first get lost.  If you can't review this for reasonableness, I'll get @peetucket to do it.

darrenleeweber:
Given the disagreement and tension around this PR, I'm absolving myself from further work on it.

PR #237, opened 2017-10-04
Require cleanup

atz:
A very simple subset of #235.  Remove `require` statements that are not useful for anything.

### DIFF
--- a/app/models/pubmed_source_record.rb
+++ b/app/models/pubmed_source_record.rb
@@ -1,5 +1,4 @@
 require 'nokogiri'
-require 'nokogiri'
 
 class PubmedSourceRecord < ActiveRecord::Base
   # validates_uniqueness_of :pmid
--- a/app/models/sciencewire_source_record.rb
+++ b/app/models/sciencewire_source_record.rb
@@ -1,6 +1,5 @@
 require 'nokogiri'
 require 'activerecord-import'
-require 'dotiw'
 
 class SciencewireSourceRecord < ActiveRecord::Base
   # validates_uniqueness_of :sciencewire_id
--- a/lib/bibtex_ingester.rb
+++ b/lib/bibtex_ingester.rb
@@ -1,6 +1,5 @@
 require 'bibtex'
 require 'citeproc'
-require 'dotiw'
 
 class BibtexIngester
   @@book_type_mapping = %w(book booklet inbook incollection manual techreport)
--- a/lib/science_wire_harvester.rb
+++ b/lib/science_wire_harvester.rb
@@ -1,4 +1,3 @@
-require 'dotiw'
 require 'parallel'
 
 class ScienceWireHarvester


app/models/pubmed_source_record.rb
```diff
@@ -1,5 +1,4 @@
 require 'nokogiri'
-require 'nokogiri'
```
darrenleeweber: weird, it was required twice, good catch

atz:
Weird, getting that intermittent error in 1 part of the build:
```
ActiveRecord::RecordNotUnique:
       Mysql2::Error: Duplicate entry '33' for key 'PRIMARY': INSERT INTO `authors` (`sunetid`, `cap_profile_id`, `active_in_cap`, `email`, `official_first_name`, `official_last_name`, `official_middle_name`, `preferred_first_name`, `preferred_last_name`, `preferred_middle_name`, `emails_for_harvest`, `id`, `created_at`, `updated_at`) VALUES ('183071', 395704, 1, 'alice.edler@stanford.edu', 'Alice', 'Edler', 'Jim', 'Alice', 'Edler', 'Jim', 'alice.edler@stanford.edu', 33, '2017-10-04 00:35:30', '2017-10-04 00:35:30')
```

Restarting that one (not that it helps our CI logic).

darrenleeweber:
Thanks for this focused PR.  What is the `dotiw` gem and why can we remove it in this PR?

https://github.com/radar/dotiw seems to provide `distance_of_time_in_words`.  Is this necessary with the version of rails in use?  In any case, is this PR only removing it from files where there is no obvious use of this method (or anything else it provides).  I don't know this gem at all.

```
$ git grep -l 'distance_of_time_in_words'
lib/cap_authors_poller.rb
lib/cap_profile_id_rewriter.rb
lib/tasks/cap_cutover.rake
lib/tasks/pubmed.rake
lib/tasks/sul.rake
$ git grep -l 'dotiw'
Gemfile
Gemfile.lock
app/models/sciencewire_source_record.rb
lib/bibtex_ingester.rb
lib/cap_authors_poller.rb
lib/cap_profile_id_rewriter.rb
lib/science_wire_harvester.rb
lib/tasks/cap_cutover.rake
```

darrenleeweber:
(Yea, there is something up with the factory creation of records that are not unique or somehow leaking across specs.  IIRC this app uses a database cleaner, not sure.)

atz:
Yeah, there are some places `dotiw` is actually used, otherwise I would have tried to remove it from Gemfile.  Right now I'm only concerned with removing it from places that **pretend** to use it.

darrenleeweber:
Is there an initializer where we can require the 'dotiw' once and for all, then remove all the other individual requires?

darrenleeweber:
I also did a git grep for just `distance_in_time` and for `time_ago_in_words`, without finding any additional files that call those methods.

atz:
Re: initializer, I dunno and it's not my concern right now.  I saw it is used by rake tasks, and sometimes those might not be inside the fully initialized application environment.  I would be happy to pursue converting the spots where we use dotiw and removing it or consolidating its usage, but to me that's not logically related to removing unnecessary `require` statements.  That's a semantic change and this is housekeeping.

darrenleeweber:
FYI, interesting argument for not using these methods from rails/dotiw at all
- https://rails-bestpractices.com/posts/2012/02/10/not-use-time_ago_in_words/

atz:
The blog post only applies when you have a client-side that includes javascript.  Our ugly rake tasks do not.

darrenleeweber:
Spun off #238 to capture thoughts around the use of this gem.

PR #239, opened 2017-10-04
Move ScienceWireHarvester internal methods to private

atz:
Distinguishing between the public API of a class and the internals is important to logic and refactorability.

All these methods are not invoked anywhere outside the class at runtime.

In one case, a method was invoked only in a test.  Since the test should not dictate the design of the class, I revised the spec to call `send` where the method is now `private`.

This shows us more than half the class was actually internal methods.

Split off from #235.

### DIFF
--- a/lib/science_wire_harvester.rb
+++ b/lib/science_wire_harvester.rb
@@ -19,25 +19,6 @@ class ScienceWireHarvester
     initialize_counts_for_reporting
   end
 
-  def harvest_pubs_for_authors(authors)
-    authors.find_each do |author|
-      begin
-        @harvested_for_author_count = 0
-        @author_count += 1
-        if @author_count % 50 == 0
-          msg = "up to Author.find_by(id: #{author.id}) with #{@total_suggested_count} total suggestions so far for #{@author_count} authors - #{Time.zone.now}"
-          logger.info msg
-        end
-        harvest_for_author(author)
-      rescue => e
-        NotificationManager.error(e, "Harvest failed for Author.find_by(id: #{author.id}, cap_profile_id: #{author.cap_profile_id})", self)
-      end
-    end
-    process_queued_sciencewire_suggestions
-    process_queued_pubmed_records
-    write_counts_to_log
-  end
-
   def harvest_pubs_for_author_ids(author_ids)
     logger.info 'Starting harvest for specific authors'
     harvest_pubs_for_authors Author.where(id: author_ids)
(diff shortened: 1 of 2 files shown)


lib/science_wire_harvester.rb
```diff
@@ -339,74 +149,10 @@ def harvest_from_directory_of_wos_id_files(path_to_directory)
...
-  end
-
   # Used for targeted harvesting for an author by sunetid and a json-file with an array of WOS id
   def harvest_for_sunetid_with_wos_json(sunetid, path_to_json, batch_size = 20)
     logger.info "Started Web Of Science harvest for #{sunetid} with wos-json #{path_to_json}"
-
-    wos_ids = JSON.parse IO.read path_to_json
+    wos_ids = JSON.parse(IO.read(path_to_json))
     harvest_sw_pubs_by_wos_array_and_sunetid sunetid, wos_ids, batch_size
   end
```
darrenleeweber: Let's just remove all the empty lines from this short method?

lib/science_wire_harvester.rb
```diff
@@ -443,24 +189,271 @@ def harvest_from_wos_id_cap_profile_id_report(path_to_report)
     process_wos_ids_for_author(batch_profile_id, wos_ids)
   end
 
-  # @param [String] line A string with the format "WOS:000314860700023	38808"
-  # @return [Array<String>] first element wos_id. second_element cap_profile_id
-  def parse_wos_line(line)
-    line =~ /WOS:([^\s]*)\s+(\d+)/
-    [Regexp.last_match(1), Regexp.last_match(2)]
-  end
+  private
```
darrenleeweber: While we're moving methods around, let's put them in alphanumeric order by method name.

lib/science_wire_harvester.rb
```diff
@@ -443,24 +189,271 @@ def harvest_from_wos_id_cap_profile_id_report(path_to_report)
...
+      if sunetid.nil?
+        logger.error "No sunetid found for Author.find_by(cap_profile_id: #{cap_profile_id}). Skipping"
+      else
+        logger.info "Starting harvest for Author.find_by(sunetid: #{sunetid}, cap_profile_id: #{cap_profile_id}) with #{wos_ids.size} WOS ids"
+        harvest_sw_pubs_by_wos_array_and_sunetid sunetid, wos_ids
+      end
+    end
 
     def logger
       @logger ||= NotificationManager.sciencewire_logger
```
darrenleeweber: Convert to attr_reader with init in the initializer?

atz: This is lazier (therefore better), since logger might not get called for a given set of operations.

atz:
Sorted methods and squashed.

darrenleeweber:
Checking the remaining public methods to confirm they are called elsewhere.

- `harvest_pubs_for_authors` might also be private
```
$ git grep 'harvest_pubs_for_authors' | egrep -v '^spec/lib/science_wire_harvester'
lib/science_wire_harvester.rb:  def harvest_pubs_for_authors(authors)
lib/science_wire_harvester.rb:    harvest_pubs_for_authors Author.where(id: author_ids)
lib/science_wire_harvester.rb:    harvest_pubs_for_authors authors unless authors.empty?
lib/science_wire_harvester.rb:      harvest_pubs_for_authors authors
```

Issue #241, opened 2017-10-04
Normalize DOI, ISBN and PMID in PublicationIdentifiers and PubHash data


peetucket:
Our publication identifiers database has inconsistent use of DOI formatting, e.g. ```["10.1016/j.tecto.2012.04.002", "10.1155/2013/342479", "10.1001/jama.2011.1627", "10.1016/j.jcrs.2011.05.043.", "10.1007/s11426-009-0030-2", "DOI: 10.1002/hbm.21338", "10.1029/2005TC001887", "doi44-22/1034"]```

darrenleeweber:
Collect examples of things to normalize:
- [ ] `10.1016/j.jpsychires.2013.07.004. [Epub ahead of print]`

darrenleeweber:
Maybe we can use this gem - https://github.com/altmetric/identifiers

darrenleeweber:
The log file of running the Altmetrics identifier script in #246 on a prod-db dump is too large to upload into github (> 10Mb)

darrenleeweber:
Some example errors identified by the altmetric identifier script:
```
$ grep 'ERROR' log/altmetric_parser.log | grep -v 'BLANK' | sed -e 's/.*ERROR -- : //g' | head -n50
PublicationIdentifier 889213: SulPubID 242841: '' extracted from 'DOI999999'
PublicationIdentifier 889498: SulPubID 242934: '' extracted from '10.1378'
PublicationIdentifier 889500: SulPubID 242935: '' extracted from '10.1002'
PublicationIdentifier 889585: SulPubID 242989: '' extracted from '10.1523'
PublicationIdentifier 889733: SulPubID 243041: '' extracted from '978-1-137-03424-3'
PublicationIdentifier 889735: SulPubID 243042: '' extracted from '978-1-137-03403-8'
PublicationIdentifier 889737: SulPubID 243043: '' extracted from '978-1-137-03420-5'
PublicationIdentifier 889739: SulPubID 243044: '' extracted from '978-1-137-03428-1'
PublicationIdentifier 890388: SulPubID 243439: '' extracted from 'DOI888-chg'
PublicationIdentifier 890389: SulPubID 243440: '' extracted from 'DOI8888-chg'
PublicationIdentifier 890474: SulPubID 243496: '' extracted from 'DOI222-30/4578'
PublicationIdentifier 890475: SulPubID 243497: '' extracted from 'DOI333-10/4567'
[...]

PR #242, opened 2017-10-04
WoS DOI Search

### LINKED ISSUE #243: Modify the DOI lookup query to account for the fact that a DO search on WoS API does partial matches
DOI search on `10.1378` hits 34,005 records, thus suggesting it is matching partial results, which makes no sense for a DOI, which is a specific record lookup.

Our use case for doing this search is for CAP users looking a single publication, but we may need to deal with the fact that many more results get returned (by perhaps restricting/limiting the results that come from a lookup)

darrenleeweber:
Fix #243 

WoS DOI queries, see
https://github.com/sul-dlss/sul_pub/wiki/Web-Of-Science-API-Queries#find-doi-records

Related work:
- https://github.com/sul-dlss/sul_pub/issues/241


lib/wos_queries.rb
```diff
@@ -48,6 +48,19 @@ def retrieve_by_id(uids)
...
+  # @param doi [String] a digital object identifier (DOI)
+  # @return [WosRecords]
+  def search_by_doi(doi)
+    message = search_by_doi_params(doi)
+    response = wos_client.search.call(:search, message: message)
+    records = records(response, :search_response)
+    # Return a unique DOI match or nothing, because the WoS API does partial string matching
+    # on the `DO` field.  When the result set is only one record, it's likely to be a good match; but
+    # otherwise the results could be nonsense.
+    return records if records.count == 1
```
peetucket: just reminding myself -- this return will prevent the following line from executing when reached, correct?

spec/lib/wos_queries_spec.rb
```diff
@@ -29,6 +31,25 @@
...
+      result = wos_queries.search_by_doi(doi)
+      expect(result).to be_an WosRecords
+      expect(result.count).to eq 1
+    end
+    it 'returns nothing for partial matches' do
+      savon.expects(:authenticate).returns(wos_auth_response)
+      savon.expects(:search).with(message: :any).returns(wos_search_by_doi_mismatch_response)
+      result = wos_queries.search_by_doi(doi)
+      expect(result).to be_an WosRecords
+      expect(result).to be_empty
```
peetucket: took me a second to understand why the same DOI passed in the two tests led to different results ... which of course works since the response is mocked differently as needed (i.e. the actual value of the incoming DOI is irrelevant for the tests) - maybe a small comment next to the let(:doi) statement to point that out?

lib/wos_queries.rb
darrenleeweber: Yes, when `records.count == 1` it returns `records` and short-circuits the rest of the block.

PR #250, opened 2017-10-06
WosRecords#empty?

darrenleeweber:
Simply adds an `WosRecords#empty?` method.


Maatary:
Sounds ok to me, but need to understand the bigger picture of that code. Giving it a pass for now

PR #255, opened 2017-10-06
WosClient#session_close fault handing

darrenleeweber:
While testing something is a stale console session, I ran into a rejected request because the SID had expired, so when I tried to reset it I couldn’t because the WosClient wasn’t handling the SOAP fault.  This PR will fix that.

This handles SOAP faults in the WosClient#session_close to ensure that it resets the session_id and other client objects.  It doesn't matter whether the session_close is successful or not, really, the session will expire after 4 hours of inactivity anyway.

- Adds exception handling for SOAP fault when closing a session that is not found
- Adds `NotificationManager.wos_logger`
  - configuration added to `Settings.WOS.LOG`
  - used in `WosClient#logger` private method

### DIFF
--- a/lib/wos_client.rb
+++ b/lib/wos_client.rb
@@ -62,11 +62,22 @@ class WosClient
   # Resets the session_id and the search client
   # @return [nil]
   def session_close
-    auth.globals[:headers]['Cookie'] = "SID=\"#{session_id}\""
-    auth.call(:close_session)
+    begin
+      auth.globals[:headers]['Cookie'] = "SID=\"#{session_id}\""
+      auth.call(:close_session)
+    rescue Savon::SOAPFault => fault
+      # Savon::SOAPFault: (soap:Server) No matches returned for SessionID
+      logger.warn(fault.inspect)
+    end
     @auth = nil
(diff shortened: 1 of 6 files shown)


lib/wos_client.rb
```diff
@@ -62,11 +62,22 @@ def session_id
   # Resets the session_id and the search client
   # @return [nil]
   def session_close
-    auth.globals[:headers]['Cookie'] = "SID=\"#{session_id}\""
-    auth.call(:close_session)
+    begin
+      auth.globals[:headers]['Cookie'] = "SID=\"#{session_id}\""
+      auth.call(:close_session)
+    rescue Savon::SOAPFault => fault
```
peetucket: so you just need to be sure an exception is caught and logged so the session_id can be set to nil?  (because if you don't rescue it, you'll still have an expired session_id)?

darrenleeweber: Yes, without catching the exception, the method was not resetting the `session_id==nil` and because it is memoized, it could not be reset and it was not valid any more.  This change allows it to be reset, but only by way of the courtesy call to close the session.

PR #267, opened 2017-10-10
HarvestBroker also has methods that should be private

atz:
Same principles as before.  Move internal methods to private and sort by alphanumeric order.

I also collapsed one test case that tested if the return value was an Array and Empty separately, to just test `.eq []` in one go.  No need to execute twice.

### DIFF
--- a/lib/science_wire/harvest_broker.rb
+++ b/lib/science_wire/harvest_broker.rb
@@ -23,35 +23,6 @@ module ScienceWire
     end
 
     ##
-    # The traditional Author only harvest approach
-    # @return [Array<Integer>]
-    def ids_for_author
-      name = author_name(author)
-      institution = sciencewire_harvester.default_institution
-      if seed_list.size < 50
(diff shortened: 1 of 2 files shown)


darrenleeweber:
There is 100% code coverage for `HarvestBroker` on the `master` branch.  It looks like the only place this class is called is in the `lib/science_wire_harvester.rb`, i.e.
```
$ git grep -l 'HarvestBroker' | grep -v 'spec'
lib/science_wire/harvest_broker.rb
lib/science_wire_harvester.rb
```

darrenleeweber:
I don't think this is a necessary set of changes and I recall cases where these methods have been used in console work (where using `.send` is an inconvenience).  I'm leaving this open, but I do not want to approve it.

atz:
All functionality a dev needs is still available in the console.  Nothing suggests **intention** for these to be public rather than *de facto*.   Not sure why you could accept the same revision in the other class, but not thisdf one. The undocumented, unnecessary and entirely hypothetical convenience of invocation in console is not a legitimate justification.

Issue #264, opened 2017-10-09
Analysis: how are the WoS results and the MEDLINE results different at the field level


peetucket:
This will help define if we need one or two different WoS Source Record tables.  If the fields are different, we will likely new two source records with their own parsing of the different fields.  If they have the same fields, it may be possible to use the same database table.

darrenleeweber:
Looking at some example records for Russ Altman, using the following query:
```ruby
wos_client = WosClient.new(Settings.WOS.AUTH_CODE, :info);
wos_queries = WosQueries.new(wos_client);
records = wos_queries.search_by_name('Altman, Russ');
```

darrenleeweber:
An example `WOS` record, extracted using:
```ruby
puts records.by_database('WOS').first.to_xml
```

```XML
> 
<REC r_id_disclaimer="ResearcherID data provided by Clarivate Analytics">
  <UID>WOS:000172527000005</UID>
  <static_data>
    <summary>
      <EWUID>
        <WUID coll_id="WOS"/>
        <edition value="WOS.SCI"/>
      </EWUID>
      <pub_info coverdate="NOV-DEC 2001" has_abstract="N" issue="6" pubmonth="NOV-DEC" pubtype="Journal" pubyear="2001" sortdate="2001-11-01" vol="16">
        <page begin="14" end="18" page_count="5">14-18</page>
      </pub_info>
      <titles count="6">
        <title type="source">IEEE INTELLIGENT SYSTEMS</title>
        <title type="source_abbrev">IEEE INTELL SYST</title>
        <title type="abbrev_iso">IEEE Intell. Syst.</title>
        <title type="abbrev_11">IEEE IN SYS</title>
        <title type="abbrev_29">IEEE INTELL SYST</title>
        <title type="item">Challenges for intelligent systems in biology</title>
      </titles>
      <names count="1">
        <name daisng_id="745091" seq_no="1" dais_id="10190039" reprint="Y" role="author">
          <display_name>Altman, RB</display_name>
          <full_name>Altman, RB</full_name>
          <wos_standard>Altman, RB</wos_standard>
          <first_name>RB</first_name>
          <last_name>Altman</last_name>
          <email_addr>russ.altman@stanford.edu</email_addr>
        </name>
      </names>
      <doctypes count="1">
        <doctype>Editorial Material</doctype>
      </doctypes>
      <publishers>
        <publisher>
          <address_spec addr_no="1">
[...]

PR #268, opened 2017-10-10
WosRecords - select records by database prefix

darrenleeweber:
WIP related to #264 

```ruby
wos_client = WosClient.new(Settings.WOS.AUTH_CODE, :info);
wos_queries = WosQueries.new(wos_client);
records = wos_queries.search_by_name('Altman, Russ');
records.count
#=> 464
records.by_database['WOS'].count
#=> 411
> records.by_database['MEDLINE'].count
#=> 53
# it seems this record set only contains WOS & MEDLINE, 53 should be correct
```

### DIFF
--- a/lib/wos_records.rb
+++ b/lib/wos_records.rb
@@ -27,6 +27,13 @@ class WosRecords
     @doc = Nokogiri::XML(@records) { |config| config.strict.noblanks }
   end
 
+  # Select records by the database prefix in the UID
+  # @param database [String] the UID database prefix (defaults to 'WOS')
+  # @return [Nokogiri::XML::NodeSet]
+  def by_database(database = 'WOS')
+    rec_nodes.select { |rec| record_uid(rec).start_with? database }
+  end
+
   # Iterate over the REC nodes
   # @yield rec [Nokogiri::XML::Element]
   def each
--- a/lib/wos_records.rb
+++ b/lib/wos_records.rb
@@ -27,11 +27,19 @@ class WosRecords
     @doc = Nokogiri::XML(@records) { |config| config.strict.noblanks }
   end
 
-  # Select records by the database prefix in the UID
-  # @param database [String] the UID database prefix (defaults to 'WOS')
-  # @return [Nokogiri::XML::NodeSet]
-  def by_database(database = 'WOS')
-    rec_nodes.select { |rec| record_uid(rec).start_with? database }
+  # Group records by the database prefix in the UID
+  #  - where a database prefix is missing, groups records into 'MISSING_DB'
+  # @return [Hash<String => WosRecords>]
+  def by_database
+    db_recs = rec_nodes.group_by do |rec|
+      uid_split = record_uid(rec).split(':')
+      uid_split.length > 1 ? uid_split[0] : 'MISSING_DB'
+    end
+    db_recs.each_key do |db|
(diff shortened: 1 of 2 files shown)


lib/wos_records.rb
```diff
@@ -27,6 +27,13 @@ def initialize(records: nil, encoded_records: nil)
     @doc = Nokogiri::XML(@records) { |config| config.strict.noblanks }
   end
 
+  # Select records by the database prefix in the UID
+  # @param database [String] the UID database prefix (defaults to 'WOS')
+  # @return [Nokogiri::XML::NodeSet]
+  def by_database(database = 'WOS')
+    rec_nodes.select { |rec| record_uid(rec).start_with? database }
```
atz: Probably should include the `:` delimiter in your matching.  I.E. `M != MEDLINE`.

atz: Since this method doesn't have any consumer yet, can you describe what you are expecting as a use pattern?  

If we expect consumers to iterate through both (or all) databases, might you prefer to `@return [Hash<String => Array<Nokogiri::XML::Element>>]`?  Specifically, just doing:
```ruby
rec_nodes.group_by { |rec| record_uid(rec).split(':').first }
```
Returning a each database's slice in a Hash:
```ruby
{
"WOS" => [#(Element:0x3fc69535de14 { name = "REC", children = [ #(Element:0x3fc69535c398 { name = "UID", children = [ #(Text "WOS:123")] })] })],
"MEDLINE" => [#(Element:0x3fc694cb59f4 { name = "REC", children = [ #(Element:0x3fc69536441c { name = "UID", children = [ #(Text "MEDLINE:456")] })] })]
}
```

darrenleeweber: Good suggestions, thanks for considerate review.  Also, the immutability of the data in the WosRecords is a problem, because the NodeSet enumerable methods used in this PR return an object that still links back to the parent document.  It might be better to follow the pattern in the duplicate-related methods to first `doc.dup` and then select/filter the nodes.  Along those lines, another option is to return a new instance of the WosRecords class for each database node set.  Maybe, following your suggestion, it could be {db_key: <WosRecords>}.

darrenleeweber: Pushed up changes, might be better.

lib/wos_records.rb
```diff
@@ -27,6 +27,22 @@ def initialize(records: nil, encoded_records: nil)
     @doc = Nokogiri::XML(@records) { |config| config.strict.noblanks }
   end
 
+  # Group records by the database prefix in the UID
+  #  - where a database prefix is missing, groups records into 'MISSING_DB'
+  # @return [Hash<String => WosRecords>]
+  def by_database
+    nodes = doc.dup.search('REC')
```
atz: I'm fairly certain that this `.dup` is no longer required, since every value in the Hash returned is a new `WosRecords` object based on the serialized representation (of already serialized elements).  Nothing attached to the old doc is returned.  Can you construct a test related to the question of immutability that would fail without the dup?  I might be missing something, but I can't imagine one right now.

Side note: `WosRecords` might be more useful if it could (also) be initialized on the Array of records (here `db_recs[db]`).

darrenleeweber: Pushed up an amended commit to remove the `doc.dup`; leaving the init change to another PR (could be good if we find additional uses for it outside this particular method).

PR #269, opened 2017-10-10
WebOfScienceSourceRecord creation

### LINKED ISSUE #184: Create new Web of Science source record model(s) and corresponding database table(s)
supercedes SciencewireSourceRecord

Needs to parse new doc result to doc-hash

and create the corresponding database for persistence

linked to #202

depends on #264

darrenleeweber:
Potential fix #184

### DIFF
--- a/app/models/web_of_science_source_record.rb
+++ b/app/models/web_of_science_source_record.rb
@@ -1,0 +1,45 @@
+class WebOfScienceSourceRecord < ActiveRecord::Base
+
+  validates :active, :database, :source_data, :source_fingerprint, :uid, presence: true
+
+  after_initialize :check_source_data
+
+  # @return doc [Nokogiri::XML::Document] XML document
+  def doc
+    @doc ||= Nokogiri::XML(source_data)
+  end
+
+  def active
+    # assume records are active until we discover a deprecation attribute
+    super || self.active = true
+  end
+
+  def database
+    super || begin
+      uid_split = uid.split(':')
+      self.database = uid_split.length > 1 ? uid_split[0] : nil
+    end
+  end
+
+  def source_fingerprint
+    super || self.source_fingerprint = Digest::SHA2.hexdigest(source_data)
+  end
+
+  # @return xml [String] XML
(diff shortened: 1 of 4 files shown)


app/models/web_of_science_source_record.rb
```diff
@@ -0,0 +1,46 @@
...
+    def extract_database
+      uid_split = uid.split(':')
+      self.database = uid_split.length > 1 ? uid_split[0] : nil
+    end
+
+    def extract_source_fingerprint
+      self.source_fingerprint = Digest::SHA2.hexdigest(source_data)
+    end
+
+    def extract_uid
```
atz: I don't think we need separate methods for `extracting` attributes.  Just wrap the (public) ActiveRecord accessor methods, like:
```ruby
def uid
  super || self.uid = doc.search('UID').text
end
```
Then we should be able to rely on the rails validator as normal (in a lazy fashion) without the `after_initialize` callback.

darrenleeweber: I'll try this, haven't seen this pattern before (been a long time since I've created an AR model). 
 If we don't use an `after_initialize`, I wonder if we need a `before_validate` or `before_save` callback to extract the field data?

darrenleeweber: Trying this out, but not sure this will work with the `database` field (or any other `nil` valued field), because when it should be `nil` (or NULL), it will always keep resetting itself and possibly updating the record when it shouldn't, i.e. consider this when it should be `nil`:
```ruby
    def database
      super || begin
        uid_split = uid.split(':')
        self.database = uid_split.length > 1 ? uid_split[0] : nil
      end
    end
```

darrenleeweber: rubocop complains about `found == in conditional`

darrenleeweber:
The spec doesn't actually use the `spec/factories/web_of_science_source_records.rb` so it might need to be removed.

darrenleeweber:
Squashed it all, should be working OK.

PR #270, opened 2017-10-11
Rake task for comparison of ScienceWire vs. WoS results

### LINKED ISSUE #213: Create a script for a report that given an author name produces a list of WoS IDs that are returned given the current Sciencewire API and the new WoS API
This is useful to determine if the new API is returning comparable data to the current Sciencewire API

For example, pick Russ Altman and Rob Dunbar, run a full dumb query with Sciencewire API, and produce a list of Sciencewire IDs (and hopefully WoSIDs).  Run the same query against the new WoS Client and produce the list of WosIDs.  Ticket is complete when those IDs are available.

atz:
Based only on author name (string) and default institution.

Fixes #213 

Relies on questionable workaround of matching ScienceWire IDs out of the local database via `publication_identifiers` table.  See https://github.com/sul-dlss/sul_pub/issues/213#issuecomment-335643165 for why that is problematic.

Output is like:
```bash
$ rake sw:wos_publications_for_name[Mangiafico,Peter,A]
Querying ScienceWire for #<ScienceWire::AuthorName:0x007fcc2decbcc8 @last="Mangiafico", @first="Peter", @middle="A">
... [List of SW IDs]

Querying WebOfScience for Mangiafico,Peter,A
... [List of WoS IDs]

0 ScienceWire IDs
0 WebOfScience IDs
	0 MEDLINE IDs: 0 of 0 matching PubIDs in DB have SW ID
	0 WOS IDs: 0 of 0 matching PubIDs in DB have SW ID

Total INTERSECTION of ScienceWire IDs (0)
	0 MEDLINE
	0 WOS

Unmatched ScienceWire IDs: 
... [List of unmatched SW IDs]
```

### DIFF
--- a/lib/tasks/sw.rake
+++ b/lib/tasks/sw.rake
@@ -2,17 +2,15 @@ namespace :sw do
   def harvester
     @harvester ||= begin
       h = ScienceWireHarvester.new
-
       Signal.trap('USR1') do
         harvester.debug = true
       end
-
       h
     end
   end
 
   # rake sw:fortnightly_harvest[1, 5]
-  desc 'harvest from sciencewire by email or known sciencewire pub ids'
+  desc 'Harvest from sciencewire by email or known sciencewire pub ids'
   task :fortnightly_harvest, [:starting_author_id, :ending_author_id] => :environment do |_t, args|
     args.with_defaults(starting_author_id: 1, ending_author_id: -1)
     starting_author_id = (args[:starting_author_id]).to_i
@@ -21,7 +19,7 @@ namespace :sw do
     # CapAuthorshipMailer.welcome_email("harvest complete").deliver
   end
 
-  desc 'harvest (with multiple processes) from sciencewire by email or known sciencewire pub ids'
+  desc 'Harvest (with multiple processes) from sciencewire by email or known sciencewire pub ids'
   task parallel_harvest: :environment do
     harvester.harvest_pubs_for_all_authors_parallel
   end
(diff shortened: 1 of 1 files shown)


lib/tasks/sw.rake
```diff
@@ -60,6 +58,38 @@ namespace :sw do
...
   end
 
+  desc 'Compare publications (IDs) returned by APIs for an Author identified by their name [last,first,middle]'
+  task :wos_publications_for_name, [:last, :first, :middle] => :environment do |_t, args|
+    fail "last name argument is required" unless args[:last].present?
+    fail "first name argument is required" unless args[:first].present?
+    author_name = ScienceWire::AuthorName.new(args[:last], args[:first], args[:middle])
+    attribs = ScienceWire::AuthorAttributes.new(author_name, '', [], ScienceWireHarvester.new.default_institution)
+    puts "Querying ScienceWire for #{author_name.inspect}"
+    ids = ScienceWire::HarvestBroker.new(nil, ScienceWireHarvester.new).ids_from_dumb_query(attribs)
```
darrenleeweber: Is the `ids_from_dumb_query` now a private method that requires a `send` here?

darrenleeweber: Could this rake task use one instance of the `ScienceWireHarvester.new` (it's currently instantiating two of them).

lib/tasks/sw.rake
```diff
@@ -60,6 +58,38 @@ namespace :sw do
...
+    puts "\n" << ids.sort if ids.count > 0
+    puts "\nQuerying WebOfScience for #{author_name.full_name}"
+    records = WosQueries.new(WosClient.new(Settings.WOS.AUTH_CODE)).search_by_name(author_name.full_name)
+    uids = records.uids.sort
+    puts uids
+    puts "\n#{ids.count} ScienceWire IDs"
+    puts "#{uids.count} WebOfScience IDs"
+    wos_ids     = uids.select { |id| id =~ /^WOS:/ }.map { |id| id[4..-1] } # match and strip prefix
+    medline_ids = uids.select { |id| id =~ /^MEDLINE:/ }.map { |id| id[8..-1] } # match and strip prefix
```
darrenleeweber: The select and regex is working here, no problem, just wondering if it could be a Regex.match() but I don't care.  It might be more efficient to instantiate the Regex once, but I dunno for sure.  e.g.
```ruby
> id = 'WOS:1234'
=> "WOS:1234"
> wos_regex = Regexp.new(/^WOS:(.*)/)
=> /^WOS:(.*)/
> wos_regex.match(id)
=> #<MatchData "WOS:1234" 1:"1234">
> wos_regex.match(id)[1]
=> "1234"
```
It might be used like:
```ruby
ids.map {|id| match = wos_regex.match(id); match[1] if match }.compact
```

lib/tasks/sw.rake
```diff
@@ -60,6 +58,38 @@ namespace :sw do
...
+    swids_from_medline = medline_hits.map { |pub_id| pub_id.publication.sciencewire_id }.compact
+    counts = Hash.new { 0 }
+    counts[:medline_intersection] = (ids & swids_from_medline).count
+    counts[:wos_intersection]     = (ids & swids_from_wos).count
+    puts "\t#{medline_ids.count} MEDLINE IDs: #{swids_from_medline.count} of #{medline_hits.count} matching PubIDs in DB have SW ID"
+    puts "\t#{wos_ids.count} WOS IDs: #{swids_from_wos.count} of #{wos_hits.count} matching PubIDs in DB have SW ID"
+    puts "\nTotal INTERSECTION of ScienceWire IDs (#{counts.values.sum})"
+    puts "\t#{counts[:medline_intersection]} MEDLINE"
+    puts "\t#{counts[:wos_intersection]} WOS"
+    puts "\nUnmatched ScienceWire IDs: " << ids - swids_from_medline - swids_from_wos if ids.count > counts.values.sum
```
darrenleeweber: As noted in standup, maybe add some code-comments somewhere here to explain why the `PublicationIdentifier` is required and how reliable it is (in general) at resolving the identifiers into a SciWire-id.  It may not be obvious that the WoS-id is not the same thing as a SciWire-id, hence the use of the `PublicationIdentifier` to make that connection for existing pubs.  Also we could note that there is a related rake task to find WosItemID values that are missing for existing SciWire pubs (it could be run before running this script to update the data in the `PublicationIdentifier`).

lib/tasks/sw.rake
atz: That hasn't been merged yet, so isn't yet required.  Could use the same Harvester, but I think it is useful that they don't have to be the same (`default_institution` could be a class method).  I'll say again I think `ScienceWire::HarvestBroker.new`'s 2nd param should take `ScienceWireHarvester.new` as a default also.

PR #277, opened 2017-10-13
Last little tweak to task

atz:
More counts and an extra newline character.

### DIFF
--- a/lib/tasks/sw.rake
+++ b/lib/tasks/sw.rake
@@ -84,10 +84,11 @@ namespace :sw do
     counts[:wos_intersection]     = (ids & swids_from_wos).count
     puts "\t#{medline_ids.count} MEDLINE IDs: #{swids_from_medline.count} of #{medline_hits.count} matching PubIDs in DB have SW ID"
     puts "\t#{wos_ids.count} WOS IDs: #{swids_from_wos.count} of #{wos_hits.count} matching PubIDs in DB have SW ID"
-    puts "\nTotal INTERSECTION of ScienceWire IDs (#{counts.values.sum})"
+    puts "\nTotal INTERSECTION with ScienceWire IDs (#{counts.values.sum} of #{ids.count}):"
     puts "\t#{counts[:medline_intersection]} MEDLINE"
     puts "\t#{counts[:wos_intersection]} WOS"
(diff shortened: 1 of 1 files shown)


lib/tasks/sw.rake
```diff
@@ -84,10 +84,11 @@ namespace :sw do
...
     puts "\t#{medline_ids.count} MEDLINE IDs: #{swids_from_medline.count} of #{medline_hits.count} matching PubIDs in DB have SW ID"
     puts "\t#{wos_ids.count} WOS IDs: #{swids_from_wos.count} of #{wos_hits.count} matching PubIDs in DB have SW ID"
-    puts "\nTotal INTERSECTION of ScienceWire IDs (#{counts.values.sum})"
+    puts "\nTotal INTERSECTION with ScienceWire IDs (#{counts.values.sum} of #{ids.count}):"
     puts "\t#{counts[:medline_intersection]} MEDLINE"
     puts "\t#{counts[:wos_intersection]} WOS"
-    puts "\nUnmatched ScienceWire IDs: " << (ids - swids_from_medline - swids_from_wos).join("\n") if ids.count > counts.values.sum
+    unmatched = ids - swids_from_medline - swids_from_wos
+    puts "\n#{unmatched.count} Unmatched ScienceWire IDs:\n" << unmatched.join("\n") unless unmatched.empty?
   end
```
darrenleeweber: It's not a blocker but might read a little better without a capitalized `Unmatched` on line 91.  Not sure that `INTERSECTION` needs to be all caps on line 87.  Another idea (again, not a blocker) is to name the `counts` keys so that an iteration on the counts hash could return everything without so many `puts`, like a generic
```ruby
counts.map {|k,v| "#{k}: #{v}" }.join("\n")
```
If some of the keys should have a "\t" then perhaps add it to the key (unless it needs to be accessed later for additional calculations, where the "\t" gets in the way and this could mean using String keys instead of Symbol)

darrenleeweber: Or maybe use JSON to pretty print the `counts`.

atz: The tabs are for outputs that are a strict subset of the previous unindented line's total.

PR #279, opened 2017-10-16
bundle update

darrenleeweber:
Housekeeping - testing whether a global `bundle update` is feasible.


darrenleeweber:
Abandoning this global bundle update.  Enabled dependabot to run more atomic bundle updates.

PR #296, opened 2017-10-18
Sensible defaults

atz:


darrenleeweber:
If f0a61fc55823d5a82657cedcdca74096ec916613 is in a separate PR, I can easily see, understand and approve that.  I don't fully understand the other commit (yet) and can't approve it until I understand it, so that's blocking approval of this PR as a whole.  If you want a quick turn around on the commit that I can easily approve, split it out, or wait until I get enlightened about the other one too.

app/controllers/publications_controller.rb
```diff
@@ -105,21 +105,21 @@ def sourcelookup
...
       metadata = {
         _created: Time.zone.now.iso8601,
         description: description,
         format: 'BibJSON',
         license: 'some licence',
-        query: query,
+        query: env['ORIGINAL_FULLPATH'].to_s,
         records:  records.count.to_s
       }
       metadata[:page] = page || 1
```
darrenleeweber: In general, changing the method param signature, esp. removing a param entirely to replace it with an ENV value, is a refactor that diminishes the utility of the method.  What if we later want to call it with something else that's not in the ENV?  The `query` param is gone, we can't do it any more.  I'm not a fan of removing params, I would be open to converting all the params to keyword-params and setting a default value for one or more of them.  As it is, I'm not very familiar with this section of code and need more time to see it in context and do some repo analysis on it (tomorrow).

darrenleeweber: Looking into this a bit more....
```
$ git grep 'wrap_as_bibjson_collection'
app/controllers/publications_controller.rb:          bibjson = wrap_as_bibjson_collection(description, matching_records, page, per)
app/controllers/publications_controller.rb:        bibjson = wrap_as_bibjson_collection(description, matching_records)
app/controllers/publications_controller.rb:    def wrap_as_bibjson_collection(description, records, page = nil, per_page = nil)
```
So it has very isolated use in this class only, so a private method it is and should be (just checking).

darrenleeweber:
```
$ git grep 'ORIGINAL_FULLPATH'
app/controllers/publications_controller.rb:        query: env['ORIGINAL_FULLPATH'].to_s,
```
WTF is this thing?  Is rack/rails setting this or what?

atz: Right, `wrap_as_bibjson_collection` is a private method, so changing the signature is fine.  Both places it was called invoked the same ENV, so there is no difference and no use case for passing anything else.  For a refactor, it doesn't matter what it does, it matters that it is the same thing.   

More broadly, that ENV value is coming from the webserver, like `REMOTE_USER`.  It is not manually set.

atz:
Any developer should be able to read a PR that touches only eleven lines total.  This is not onerous.  

Adding a default parameter is a commonplace change, especially when there is a clear preferred value.  We know we can add the default, because at first the argument is required, so every caller is providing it.  By definition nothing changes.  Then every caller that was sending the value that is now default can drop it (literally all invocations outside the class were sending the same value).

app/controllers/publications_controller.rb
darrenleeweber: The defaults for `, page = nil, per_page = nil` are rubbish and should use the values in the body of the method.  Can we add this to this change?  What about changing the params to be keyword params, so when we look at the method calls, it's easier to know what is being set?  (BTW, congrats on restraining these changes, this controller is begging for tons of changes, I hate it.)

darrenleeweber: `wrap_as_bibjson_collection(description, query, records, page = nil, per_page = nil)` might become:
```ruby
def wrap_as_bibjson_collection(description, query: ENV['ORIGINAL_FULLPATH'].to_s, records:, page: 1, per_page: 'all')
```

darrenleeweber:
Comments like "Any developer should be able to read a PR that touches only eleven lines total. This is not onerous." are an insult.  Of course I can read it!

app/controllers/publications_controller.rb
atz: I might agree with you on the `page` and `per_page` defaults, but why retain a parameter that is literally never used?  The point of this refactor is to reflect actual uses, not hypothetical ones.

atz:
@darrenleeweber Pulled up `page` and `per_page` default values into the method sig.

PR #299, opened 2017-10-18
add a rake task that will merge publications from one author into another

peetucket:
used when there are two known duplicate author rows that refer to the same person

### DIFF
--- a/lib/tasks/cleanup.rake
+++ b/lib/tasks/cleanup.rake
@@ -3,8 +3,8 @@ namespace :cleanup do
   # Use case: a single author has two author rows with publications associated with each.
   # You want to merge one author into the author, carrying any existing publications but not duplicating them.
   # This happens when two profiles are created initially because CAP was not able to match the physician information
-  #  to the faculty information until after two profiles were created.  They "merged" them on the CAP side, but the
-  #  publications were not merged on the SUL-PUB side.  This manifests itself as unexpected behavior (missing pubs, etc.)
+  # to the faculty information until after two profiles were created.  They "merged" them on the CAP side, but the
+  # publications were not merged on the SUL-PUB side.  This manifests itself as unexpected behavior (missing pubs, etc.).
   #
   # This rake task takes in two cap_profile_ids and will merge all of the publications *from* DUPE_CAP_PROFILED_ID's profile *into* PRIMARY_CAP_PROFILE_ID's profile.
   # It will then deactivate DUPE_CAP_PROFILED_ID's profile (which should now have no publications associated with it) to prevent harvesting into it.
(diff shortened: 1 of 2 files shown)


lib/tasks/cleanup.rake
```diff
@@ -0,0 +1,87 @@
...
+  # This happens when two profiles are created initially because CAP was not able to match the physician information
+  #  to the faculty information until after two profiles were created.  They "merged" them on the CAP side, but the
+  #  publications were not merged on the SUL-PUB side.  This manifests itself as unexpected behavior (missing pubs, etc.)
+  #
+  # This rake task takes in two cap_profile_ids and will merge all of the publications *from* DUPE_CAP_PROFILED_ID's profile *into* PRIMARY_CAP_PROFILE_ID's profile.
+  # It will then deactivate DUPE_CAP_PROFILED_ID's profile (which should now have no publications associated with it) to prevent harvesting into it.
+
+  # bundle exec RAILS_ENV=production PRIMARY_CAP_PROFILE_ID=123 DUPED_CAP_PROFILED_ID=456 rake cleanup:merge_profiles # will merge all publications from cap_profile_id 456 into 123, without duplication
+  # bundle exec RAILS_ENV=production PRIMARY_CAP_PROFILE_ID=123 DUPED_CAP_PROFILED_ID=456 DRY_RUN=true rake cleanup:merge_profiles # a dry run will show totals and IDs but not change the database
+  task :merge_profiles => :environment do |_t, args|
```
atz: Why take `args` and then not use them?  This task should not rely on ENV vars, but have two required arguments as part of its signature.

peetucket: is there a nice way to do named args, so you can clearly which id is which (if you invert them for example, you will get a very wrong result)

darrenleeweber:
I'd like to see a documented issue on what the data problem is to be addressed by this PR.  As a quick-fix, this work may be necessary, but a well documented issue should enable us to address the problem long-term, if not eliminate it entirely.

lib/tasks/cleanup.rake
darrenleeweber: SO is your best friend, e.g.

peetucket:
Problem attempting to work around (this PR is not a solution, but helps fix the data problem after the fact): https://github.com/sul-dlss/sul_pub/issues/314

lib/tasks/cleanup.rake
peetucket: It's not clear that lets you do is named parameters so you can't make a mistake by sending params in the wrong order (i.e. the "primary" and the "duped" being flipped would lead to the exact opposite result of what you want).

atz: @peetucket It should be obvious by how you name the params.  For example, part of the `rake -T` output is:
```bash
rake pubmed:pmid_profile_id_import[path_to_report,cap_profile_id]       
```
That is a result of `pubmed.rake` declaring:
```ruby
task :pmid_profile_id_import, [:path_to_report,:cap_profile_id] ...
```

peetucket: changed ENV variables to args

atz:
Are we are punting on the question of testability?  The hardcoding of `dry_run` flag feels bogus.  That is exactly the important kind of feature that should be tested.  Otherwise, you don't have any guarantees that the "dry run" isn't actually leaking changes into your production data.  That important state is one that literally never exists in the codebase (until you manually alter it in production).  

Otherwise OK w/ it.

peetucket:
Yeah, ok - I can ditch the dry_run stuff, should just trust that it works.  Let me spend at least a bit of time to try to get a test in there.

peetucket:
Added a couple simple tests - getting the rake infrastructure up was straightforward, though had to do a bunch of mocking to avoid setting up fixtures.

spec/tasks/cleanup_rake_spec.rb
```diff
@@ -0,0 +1,71 @@
+require 'spec_helper'
```
atz: This is redundant to `.rspec`'s 
```
--require spec_helper
```

peetucket:
removed require statement

atz:
Added a PR onto this PR.  

The only other thing I would want is for the specs to run silently without outputting all those `put` statements in the middle of the suite.  Usually that is accomplished by writing via a Logger object that can be mocked or directed at `/dev/null` during tests.  

There may be other tricks, but stubbing `Kernel.puts` or mocking `$stdout` seem like clearly bad choices.

peetucket:
mocking puts in a before block for just this spec seems to have the intended impact of cleaning up the spec output without inadvertently suppressing any legitimate output

PR #303, opened 2017-10-19
Added Hashie dependency

Maatary:


Gemfile
```diff
@@ -49,6 +49,8 @@ gem 'turnout'
 gem 'whenever', require: false
 gem 'yaml_db'
 
+gem 'hashie'
+
```
darrenleeweber: Add gems in alphabetical order in the main block of gems.

darrenleeweber:
The travis PR failure was a flapping spec, restarted it.

Gemfile
darrenleeweber: From the current production bundle:
```
[pub@sul-pub-prod current]$ bundle show --paths | grep hashie
/opt/app/pub/sul-pub/shared/bundle/ruby/2.3.0/gems/hashie-3.4.4
```
Should we add any constraints on the version to bundle?

darrenleeweber:
I found where Hashie was lost in the grape update
- https://github.com/sul-dlss/sul_pub/commit/7c10b453748a97a4e111e8fe27cf7cae0d2e2315#diff-e79a60dc6b85309ae70a6ea8261eaf95

darrenleeweber:
I think I'm going to revert the grape update and create a new ticket to pin grape and update grape later.

darrenleeweber:
Replacing this with a revert of the grape update, i.e.
https://github.com/sul-dlss/sul_pub/pull/307

PR #318, opened 2017-10-24
Links AMR client and associated request template

### LINKED ISSUE #193: Write a client that accepts WoSIDs and returns PMIDs using LinksAMR to determine which WoSIDs have PMIDs
Problem: PMID not provided in WoS API search results anymore

Solution: Use the Links Article Match Retrieval Service API

API Documentation: http://ipscience-help.thomsonreuters.com/LAMRService/WebServicesOverviewGroup/overview.html

### LINKED ISSUE #247: Update shared_config for authentication for Links AMR api
Currently we have the base64 encoded version, but we also need the username/password for use with the Links AMR API.  Likely good idea to remove the existing auth token since it can be derived from the username/password by doing a ```Base64.encode64('username:password')```

Also requires changes to the auth the auth code is retrieved and sent to the WoS Client, see #252

### LINKED ISSUE #252: Generate authorization token for the WoS Client using the username/password in the shared_config
Current WoS client is instantiated with the auth token in shared_configs...once that is removed and replaced with a username/password, we will need to generate the token 

e.g. ```WosClient.new(Settings.WOS.AUTH_CODE, :debug)``` won't work anymore

but you can generate the AUTH_CODE from the username/password with ```Base64.encode64('username:password')```

#247 is a pre-req

atz:
Example usage:
```ruby
client = Clarivate::LinksClient.new
=> #<Clarivate::LinksClient:0x007fdc8c54e640 @host="https://ws.isiknowledge.com", @password="xxxxx", @username="StanfordU_SW">
resp = client.links(['000346594100007', '000081515000015'])
  Rendered pages/clarivate_links.xml.erb (0.1ms)
=> {
"000081515000015"=>{:pmid=>"10435530", :ut=>"000081515000015", :doi=>"10.1118/1.598623"}, 
"000346594100007"=>{:ut=>"000346594100007", :doi=>"10.1002/2013GB004790"}
}
```
TODO: 
- tests
- uncap limit of incoming ids, batch 50 per request

Fixes #193 
Fixes #252 
Fixes #247

### DIFF
--- a/lib/clarivate/links_client.rb
+++ b/lib/clarivate/links_client.rb
@@ -1,0 +1,69 @@
+module Clarivate
+  # Links AMR (Article Match Retrieval) Service
+  # @see http://ipscience-help.thomsonreuters.com/LAMRService/WebServicesOverviewGroup/overview.html Service documentation
+  class LinksClient
+    attr_reader :username, :password, :host
+
+    # @param [String] username
+    # @param [String] password
+    # @param [String] host
+    def initialize(username: nil, password: nil, host: 'https://ws.isiknowledge.com')
+      @host = host
+      @username = username
+      @password = password
+      (@username, @password) = Base64.decode64(Settings.WOS.AUTH_CODE).split(':', 2) unless username
+    end
+
+    # limited to 50 ids
+    # @param [Array<String>] ids
+    # @return [Hash<String => Hash>]
+    def links(ids)
+      raise ArgumentError, "1-50 ids required" unless ids.present? && ids.count <= 50
+      response = connection.send(:post) do |req|
+        req.path = '/cps/xrpc'
+        req.body = request_body(ids.uniq)
+      end
+      ng = Nokogiri::XML(response.body) { |config| config.strict.noblanks }.remove_namespaces!
+      pairs = ng.xpath('response/fn/map/map').map do |node|
+        [node.attr('name').split('_').last, vals_to_hash(node.children.xpath('val'))]
+      end
+      pairs.to_h
+    end
+
+    private
+
+      # @return [Faraday::Connection]
+      def connection
+        @connection ||= Faraday.new(url: host) do |faraday|
+          faraday.use Faraday::Response::RaiseError
+          faraday.adapter :httpclient
+        end
+      end
+
+      # @return [ActionView::Base] used to render as though in an rails controller
+      def renderer
+        @renderer ||= ActionView::Base.new(ActionController::Base.view_paths, {})
+      end
+
+      # @return [String]
+      def request_body(ids)
+        renderer.render(
+          file: 'pages/clarivate_links.xml',
+          layout: false,
+          locals: {
+            client: self,
+            return_fields: %w(ut doi pmid),
+            cites: ids
+          }
+        )
+      end
+
+      # @param [Nokogiri::XML::Nodeset] vals
+      # @return [Hash<Symbol => String>] namespace to value
+      # Nodeset for [<val name="ut">UT value</val>, <val name="doi">DOI</val>]
+      # becomes { ut: 'UT value', doi: 'DOI'}
+      def vals_to_hash(vals)
+        vals.map { |val| [val.attr("name").downcase.to_sym, val.text.strip] }.to_h
+      end
+  end
+end
--- a/lib/clarivate/links_client.rb
+++ b/lib/clarivate/links_client.rb
@@ -19,13 +19,13 @@ module Clarivate
     # @return [Hash<String => Hash>]
     def links(ids)
       raise ArgumentError, "1-50 ids required" unless ids.present? && ids.count <= 50
-      response = connection.send(:post) do |req|
+      response = connection.post do |req|
         req.path = '/cps/xrpc'
         req.body = request_body(ids.uniq)
       end
       ng = Nokogiri::XML(response.body) { |config| config.strict.noblanks }.remove_namespaces!
       pairs = ng.xpath('response/fn/map/map').map do |node|
-        [node.attr('name').split('_').last, vals_to_hash(node.children.xpath('val'))]
+        [node.attr('name').split('_', 2).last, vals_to_hash(node.children.xpath('val'))]
       end
       pairs.to_h
     end
--- a/app/views/pages/clarivate_links.xml.erb
+++ b/app/views/pages/clarivate_links.xml.erb
@@ -1,0 +1,25 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<request xmlns="http://www.isinet.com/xrpc41" src="app.id=API Demo">
+  <fn name="LinksAMR.retrieve">
+    <list>
+      <map>
+        <val name="username"><%= client.username %></val>
+        <val name="password"><%= client.password %></val>
+      </map>
+      <map>
+        <list name="WOS">
+          <% return_fields.each do |val| %>
+            <val><%= val %></val>
+          <% end %>
+        </list>
+      </map>
+      <map>
+        <% cites.first(50).each do |id| %>
+          <map name="cite_<%= id %>">
+            <val name="ut"><%= id %></val>
+          </map>
+        <% end %>
+      </map>
+    </list>
+  </fn>
+</request>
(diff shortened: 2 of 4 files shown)


lib/clarivate/links_client.rb
```diff
@@ -0,0 +1,65 @@
...
+        renderer.render(
+          file: 'pages/clarivate_links.xml',
+          layout: false,
+          locals: {
+            client: self,
+            return_fields: %w(ut doi pmid),
+            cites: ids
+          }
+        )
+      end
```
darrenleeweber: Move the XML template into `lib/clarivate/clarivate_links.xml` and change the `file:` to get it from there.

lib/clarivate/links_client.rb
```diff
@@ -0,0 +1,65 @@
...
+
+      # @param [<Nokogiri::XML::Nodeset>] vals
+      # @return [Hash<Symbol => String>] namespace to value
+      # Nodeset for [<val name="ut">UT value</val>, <val name="doi">DOI</val>]
+      # becomes { ut: 'UT value', doi: 'DOI'}
+      def vals_to_hash(vals)
+        vals.map { |val| [val.attr("name").downcase.to_sym, val.text.strip] }.to_h
+      end
```
darrenleeweber: Why all the massage for `val.attr("name").downcase.to_sym`?  Simply `val.attr("name")` should be OK, so the client doesn't mask anything in the data.  Do we need the `ut` in the vals - can we assume that their response map is always reliable?  Can we assume that there can only be one `ut` and if we have it in the hash-key, that's enough?

lib/clarivate/links_client.rb
```diff
@@ -0,0 +1,65 @@
...
+
+      def request_body(ids)
+        renderer.render(
+          file: 'pages/clarivate_links.xml',
+          layout: false,
+          locals: {
+            client: self,
+            return_fields: %w(ut doi pmid),
+            cites: ids
+          }
```
darrenleeweber: Do we need the `ut` in the `return_fields`?  Can we assume that their response map is always reliable?  Can we assume that there can only be one `ut` and if we have it in the hash-key, that's enough?

lib/clarivate/links_client.rb
atz: Disinclined to do that.  Since I'm using the rails renderer, I'm following rails convention.

lib/clarivate/links_client.rb
atz: Because I wanted the Hash to be keyed by symbols (`val.attr("name")` is a string), and I wanted to be strict about what goes in.

lib/clarivate/links_client.rb
atz: No reason not to have it.  Also makes sense if we start querying by different input types.

lib/clarivate/links_client.rb
```diff
@@ -0,0 +1,65 @@
...
+        req.path = '/cps/xrpc'
+        req.body = request_body(ids.uniq)
+      end
+      ng = Nokogiri::XML(response.body) { |config| config.strict.noblanks }.remove_namespaces!
+      pairs = ng.xpath('response/fn/map/map').map do |node|
+        [node.attr('name').split('_').last, vals_to_hash(node.xpath('//val'))]
+      end
+      pairs.to_h
+    end
+
```
darrenleeweber: What's the logic in separating the `@host` and the `req.path` - can we put the path into the connection object on init (in the `connection` method)?  If the `path` is a constant, a private `PATH = x.freeze` is the place for it; if it might be a config option, `Settings.WOS.LINKS_PATH` could be the place for it.  Consider the config `Settings.WOS.LINKS_HOST` to store the default `host` location (most of the service API configs do this, IIRC).

We're assuming that the LINKS has the same credentials as the SOAP-API here and, AFAIK, that's correct.

lib/clarivate/links_client.rb
```diff
@@ -0,0 +1,65 @@
...
+    private
+
+      # @return [Faraday::Connection]
+      def connection
+        @connection ||= Faraday.new(url: host) do |faraday|
+          faraday.use Faraday::Response::RaiseError
+          faraday.adapter :httpclient
+        end
+      end
```
darrenleeweber: Add a client logger in case we need to debug failures.

lib/clarivate/links_client.rb
atz: Since it is actually working, that is demonstrably correct.  Don't care about path being a constant, since it can be overridden if desired (e.g. using a proxy).  Also, as far as this level of the code is concerned, it doesn't need to live in config for any reason.  The specs literally declare what the endpoint is, so it is fine to implement it that way.  If we need to override it, *then* we can worry about where to store the override value.  Pretty sure that will never happen though.

lib/clarivate/links_client.rb
darrenleeweber: Do we know whether anything could be lost by a `downcase` and whether there is any inconsistency in the API key strings?  (I'm considering millions of these strings going through a `downcase` that may not be necessary.)

lib/clarivate/links_client.rb
atz: If you need to debug, override `@connection` or do it in Faraday middleware.

lib/clarivate/links_client.rb
darrenleeweber: `# @param [<Nokogiri::XML::Nodeset>] vals` - probably works, I dunno; maybe the `<>` are not necessary, i.e.
`# @param [Nokogiri::XML::Nodeset] vals` is it (looks like a copy-n-paste of a .class).

lib/clarivate/links_client.rb
darrenleeweber: It will be much easier to maintain if the code is located in the same place.  This client does not need to follow a "rails-convention" - there is literally no MVC about it.  I do feel strongly about this and would require this to merge the PR.

lib/clarivate/links_client.rb
darrenleeweber: See the `NotificationManager` for the patterns established for this app.

atz: There are plenty of trash patterns in the app to be avoided.

lib/clarivate/links_client.rb
atz: I guess you could stop reviewing now then.

lib/clarivate/links_client.rb
atz: Ah, good catch.  Leftover from when I was first taking an `Array<...>`.

lib/clarivate/links_client.rb
darrenleeweber: Integration with https://github.com/sul-dlss/sul_pub/blob/master/lib/notification_manager.rb will be required to merge any API client.

atz: That won't be happening in this PR.  For one thing the use pattern is completely different: this client is *ancillary*, for data enhancement.  Right now, we project the initial version of the links client will only be executing inside the WoS client, that already has its own logging.  The case for a separate log has not been made.

Secondly, even if we determined we wanted additional logging infrastructure, it certainly isn't required *now*.  This provides a baseline for other work to proceed.  It isn't supposed to be a complete integration, based on the scope of tickets represented.

lib/clarivate/links_client.rb
darrenleeweber: Changes requested in #325 - a new PR on this branch.

lib/clarivate/links_client.rb
darrenleeweber: In #325 the fields are an input parameter to the query, rather than have them fixed.

darrenleeweber:
I want to point out that my request to use in-line XML is related to previous advice from a prior tech-lead on this project, which has shaped my expectation of what is expected for this app.  For instance, I wanted to normalize XML to ensure that it is well formed and easily composed, which I did here:

darrenleeweber:
After lengthy discussion about this PR, we have resolved that @atz will investigate opportunities to ensure that his adoption of the template-XML pattern can be applied to other areas of this code base.  Also, @atz will follow up elsewhere with code to enable logging, as he is adamant that it will not be done here.  Lastly, we disagree on a couple minor points that could be addressed elsewhere.  With these points clarified in discussion, the team has resolved to move forward with this PR.

PR #320, opened 2017-10-24
WosQueries - refactor to provide institutions parameter

### LINKED ISSUE #271: Adjust WoS queries to use institution name already configured in settings
The default institution of Stanford in the new WoS queries class (here: https://github.com/sul-dlss/sul_pub/blob/master/lib/wos_queries.rb#L158) is using a string instead of the value in the settings.yml file (here:  https://github.com/sul-dlss/sul_pub/blob/master/config/settings.yml#L58)

or it should expose an attr writer for a harvester or rake task to set that value using settings or user data

darrenleeweber:
Fix #271 
- this allows the consumer of the WosQueries object to provide any array of institutions
- the responsibility for setting the correct institutions lies with the consumer (Harvester, rake task etc.)
- this allows an author query without any institutions (previously impossible)

### DIFF
--- a/lib/wos_queries.rb
+++ b/lib/wos_queries.rb
@@ -70,9 +70,10 @@ class WosQueries
   end
 
   # @param name [String] a CSV name pattern: {last name}, {first_name} [{middle_name} | {middle initial}]
+  # @param institutions [Array<String>] a set of institutions the author belongs to
   # @return [WosRecords]
-  def search_by_name(name)
-    message = search_by_name_params(name)
+  def search_by_name(name, institutions = [])
+    message = search_by_name_params(name, institutions)
     retrieve_records(:search, message)
   end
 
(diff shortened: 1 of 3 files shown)


lib/wos_queries.rb
```diff
@@ -263,9 +258,11 @@ def search_by_doi_params(doi)
     end
 
     # @param name [String] a CSV name pattern: {last name}, {first_name} [{middle_name} | {middle initial}]
+    # @param institutions [Array<String>] a set of institutions the author belongs to
     # @return [Hash] search query parameters
-    def search_by_name_params(name)
-      user_query = "AU=(#{name_query(name)}) AND AD=(#{institutions.join(' OR ')})"
+    def search_by_name_params(name, institutions)
+      user_query = "AU=(#{name_query(name)})"
+      user_query += " AND AD=(#{institutions.join(' OR ')})" unless institutions.empty?
```
atz: Previously the default was Stanford, now the default is no constraint.  Right?  This changes the behavior of the task `wos_publications_for_name` in `sw.rake`.

darrenleeweber: Oh, if this class is already used in a rake task, I'll have to change that rake task in this PR.  Thanks, I'll do that and look for any other uses of it (probably none yet).

darrenleeweber: Done.

atz:
Resolved merge conflict in one file.

PR #324, opened 2017-10-25
WebOfScience module

darrenleeweber:
This could conflict with #320 once it is merged (or vice versa), TODO:
- [x] check/resolve conflicts with #320 
- [x] find/update all uses of `Wos*` classes
  - mostly done, but `lib/tasks/sw.rake` is already using this stuff, it has to be refactored (after #320)

Integrate new WOS code into a `WebOfScience` module.

This PR creates the `WebOfScience` module and moves classes into it and applies the module namespace everywhere the prior `Wos*` classes were called.  Would be good to make that move before going too far with using these new classes (e.g. in harvesting).  The new file structure is:

```
$ ll lib/web_of_science/
total 32K
-rw-r--r-- 1 dlweber users 2.5K Oct 24 16:41 client.rb
-rw-r--r-- 1 dlweber users 9.6K Oct 24 17:01 queries.rb
-rw-r--r-- 1 dlweber users 5.2K Oct 24 16:50 record.rb
-rw-r--r-- 1 dlweber users 5.0K Oct 24 16:51 records.rb
$ ll spec/lib/web_of_science/
total 28K
-rw-r--r-- 1 dlweber users 2.3K Oct 24 16:48 client_spec.rb
-rw-r--r-- 1 dlweber users 5.2K Oct 24 16:59 queries_spec.rb
-rw-r--r-- 1 dlweber users 6.6K Oct 24 16:50 record_spec.rb
-rw-r--r-- 1 dlweber users 7.6K Oct 24 17:03 records_spec.rb
```

Looked for any remaining `Wos` and found nothing relevant, using
```
$ git grep 'Wos'
app/data/bibtex_to_ingest/August19 17 Batch/elambin.txt:author = {Young O. and Lambin E. F. and Alcock F. and Haberl H. and Karlsson S. I. and McConnell W. J. and Myint T. and Pahl-Wostl C. and Polsky C. and Ramakrishnan P. S. and Scouvart M. and Schroeder H. and Verburg P. H.},
lib/web_of_science/records.rb:      # @param rec [Nokogiri::XML::Element] a Wos 'REC' element
lib/web_of_science/records.rb:      # @return UID [String] a Wos 'UID' value
script/missing_wos_ids.rb:class MissingWosId
script/missing_wos_ids.rb:  # Identify publications without a WosID
script/missing_wos_ids.rb:c = MissingWosId.new
```

PR #325, opened 2017-10-25
Clarivate::LinksClient - revised

darrenleeweber:
@atz - this is one of the changes that I requested in #318 - please note that this PR will merge into your branch, not master.

Changes:
- the `request_body` to use string interpolation
  - similar to https://github.com/sul-dlss/sul_pub/blob/master/lib/science_wire/query/publication_query_by_author_name.rb
  - this change releases this code from any dependency on rails, should we decide to extract this client code into a generic harvesting gem (other SUL-PUB and RIALTO discussions have alluded to this and most of the client code in this app tries to remain independent of rails; Tom has indicated tentative support for a generic harvesting project)

- the `links` method should accept a `fields` argument (let the consumer determine what they want)

- the change to the Hash returned is simply an efficiency
  - considering that this class could be working with millions of responses during the harvesting
  - assuming that Clarivate API has already normalized the key values, returning those values will provide data that is consistent with their API documentation

- the use of constants colocates the HOST/PATH values and freezes them
  - easier maintenance and the freeze might be more efficient

Checked that it works (and revised  it a bit to get it working):
```ruby
> client = Clarivate::LinksClient.new

> resp = client.links(['doi', 'pmid', 'ut'], ['000346594100007', '000081515000015'])
=> {"000081515000015"=>{"pmid"=>"10435530", "doi"=>"10.1118/1.598623", "ut"=>"000081515000015"}, "000346594100007"=>{"doi"=>"10.1002/2013GB004790", "ut"=>"000346594100007"}}

> resp = client.links(['pmid'], ['000346594100007', '000081515000015'])
=> {"000081515000015"=>{"pmid"=>"10435530"}, "000346594100007"=>{}}
```

Note: it might also be better to place this code in `lib/web_of_science`, but that depends on #324


atz:
Why do you insist on string keys?  That isn't important.  Why do you want to vary the fields being returned *now*?  That isn't yet necessary.  (And if we were going to do it, the order of params would be reversed with `fields` having a reasonable default, since `ids` CANNOT have a default.)

I am confused how you think this is helping.

atz:
The cost of manipulating keys is completely negligible.  For 100,000 strings here's a local benchmark cost of `downcase` and `to_sym` (which are implemented in C, after all):
```ruby
a = []
=> []
100000.times { a << %w(doi ut pmid).sample }
=> 100000
puts Benchmark.measure { a.map(&:downcase).map(&:to_sym) }
  0.020000   0.000000   0.020000 (  0.028296)
```
2/100th of a second.  Since this is happening after a series of REST calls with latency on a MAX of 150 strings (not 100,000), I don't think we can regard that cost as significant.  In fact, in production, I would argue you could never reliably even **detect** such a small cost.

darrenleeweber:
Revised and moved this work to #328

Issue #272, opened 2017-10-11
Name queries in the WoS client should follow the same model used in Sciencewire Client


peetucket:
When performing name searches during harvesting, we should follow the same model used in sciencewire searches

in the current sciencewire querying stack, there is a class (https://github.com/sul-dlss/sul_pub/blob/master/lib/science_wire/author_name.rb) that is used to construct a name query so that we get more comprehensive results than simply searching for "Last, First" ... it adds other common variants with just first initial or first middle, etc., e.g. instead of "Mangiafico, Peter" you will search on "Mangiafico, Peter A or Mangiafico P or Mangiafico PA"

atz:
Ideally, we should literally use the same `ScienceWire::AuthorName`, amended or extended if necessary.

darrenleeweber:
This is what it does now
- https://github.com/sul-dlss/sul_pub/blob/master/lib/wos_queries.rb#L150-L161

This was copied from Peter's initial gist on how to work with the new SOAP API.  I presume it was derived from the original ScienceWire code, but I'll double check details.

PR #327, opened 2017-10-26
Agent module

darrenleeweber:
Extract author name/address/institution stuff out of the SW module, it doesn't belong there.  Preparation for further work on #272 and also for later removing the SW module entirely.  This PR will be solely a refactor, to make sure this is done right first.

- the module cannot be named `Author`, as in `Author::Name` because there is already an `Author` model in the rails app and ruby/rails is not smart enough to differentiate between the rails model/class and a separate module.

- to check that all the things are changed, I used:
```sh
$ git grep 'ScienceWire::Author'
# lots of stuff here, but it's all AuthorAttributes stuff that should remain so; hence
# the follow returns nothing, which means all the things are changed.
$ git grep 'ScienceWire::Author' | grep -v 'AuthorAttributes'
```

- unless some `git rebase` stuff is required, I think this is done and I need it (want it) to continue work on #272

### DIFF
--- a/app/models/author_identity.rb
+++ b/app/models/author_identity.rb
@@ -11,13 +11,13 @@ class AuthorIdentity < ActiveRecord::Base
   # Converts an AuthorIdentity object to an AuthorAttributes object for use by other classes
   def to_author_attributes
     ScienceWire::AuthorAttributes.new(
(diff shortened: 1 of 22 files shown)


darrenleeweber:
Squashed.

PR #328, opened 2017-10-26
WoS links client

### LINKED ISSUE #329: Links AMR client request template location
The RubyMine IDE complains about the location of the Links AMR request template, see screenshot attached.  It suggests that it violates some rails patterns about the location of view templates.

screenshot from 2017-10-25 22-46-06

### LINKED ISSUE #345: WOS Links Client should return empty hashes for no results
e.g.
```ruby
links_client = Clarivate::LinksClient.new;
links_client.links(['MEDLINE:24452614'])
# => {"MEDLINE:24452614"=>{:message=>"No Result Found"}}
```

It's better for this result to be an empty Hash, i.e.:
```
{"MEDLINE:24452614"=>{} }
```
The client will need to detect this 'message' and remove the data, so that consumers can rely on a `.blank?` or a `.empty?` to detect that no results were found.

Similary, returning the UID requested (UT) in the result is not helpful, e.g.
```ruby
links_client.links([wos_record_encoded.uid])
=> {"WOS:A1972N549400003"=>{:ut=>"A1972N549400003"}}
```
In effect, this is an empty result set, because it returns no additional identifiers.  It should also return an empty hash (i.e. do not request the UT field in the first place).

darrenleeweber:
Fix #329 
Fix #345 

Connected to #193 - which requires only PMID links; this PR allows a consumer of this links client to request only PMID links (or anything the links service allows).

This is some remaining changes that I requested in #318 (it has been adapted to preserve some work done in #325).  This is now consistent with the spirit of #318, with a couple of tweaks to open it up and make it more flexible to use.

There is some kind of IDE complaint about the location of the template-XML file that I don't understand, but it's captured in #329 and this PR does not attempt to fix it.  This PR does relocate the template file successfully (it eliminates `ActionController` in the process, now only depends on `ActionView` with fewer paths to search to find the template-XML file).

Changes:
- moved client into `lib/web_of_science` / `WebOfScience` module
- extract methods refactor for response-parsing
- the `links` method can accept a `fields` argument that has a default set of fields
  - do not request the 'ut' field so it's easier to detect empty results, fix #345 
- the `request_body` can use template-XML, with a `fields` argument
  - the `renderer` is simplified by passing full file path to the `renderer.render`
    - fixes https://github.com/sul-dlss/sul_pub/issues/329
  - the template-XML is easily co-located with the client class file
- the Hash returned is empty when the service has "No Results Found", fix #345 
- the change to the Hash returned is to preserve the mapping from values in fields to the response
  - we can assume the links-API has already normalized the key values and
  - the key values they return match the field names requested
  - returning those values will use field names that are consistent with their API documentation
  - maybe it could be a `HashWithIndifferentAccess`, but have not explored that yet
  - it will help to see how this is intended to be used in
    - https://github.com/sul-dlss/sul_pub/blob/wos-harvest/lib/web_of_science/harvester.rb
    - https://github.com/sul-dlss/sul_pub/blob/wos-harvest/lib/web_of_science/harvester.rb#L74-L86
    - the process of enhancing the identifiers will need to use the same key-value strings that are in the original records retrieved to get a successful merge
- the use of constants colocates the HOST/PATH values and freezes them
  - easier maintenance if/when they change the links connection
  - the freeze is a rubocop thing for constants

Checked that it works (and revised  it a bit to get it working):
```ruby
# client
> client = WebOfScience::LinksClient.new
=> #<WebOfScience::LinksClient:0x00559d6d822e30 @host="https://ws.isiknowledge.com", @password="snipped", @username="snipped">

# default fields
> client.links(['000346594100007', '000081515000015'])
  Rendered lib/web_of_science/links_request.xml.erb (2.8ms)
=> {"000081515000015"=>{"pmid"=>"10435530","doi"=>"10.1118/1.598623"}, "000346594100007"=>{ "doi"=>"10.1002/2013GB004790"}}


# I only want these fields
> resp = client.links(['000346594100007', '000081515000015'], fields: ['doi', 'pmid'])
=> {"000081515000015"=>{"pmid"=>"10435530", "doi"=>"10.1118/1.598623"}, "000346594100007"=>{"doi"=>"10.1002/2013GB004790"}}

# ---
# If only PMID values are requested
# Match up field names with the response data to filter responses

> fields = ['pmid']
=> ["pmid"]

> resp = client.links(['000346594100007', '000081515000015'], fields: fields)
=> {"000081515000015"=>{"pmid"=>"10435530"}, "000346594100007"=>{}}

> resp.map {|_k,v| v.keys }
=> [["pmid"], []]

> resp.map {|_k,v| v.keys & fields }
=> [["pmid"], []]

> resp.map {|_k,v| (v.keys & fields).empty? }
=> [false, true]

> resp.select {|_k,v| ! (v.keys & fields).empty? }
=> {"000081515000015"=>{"pmid"=>"10435530"}}

> resp.select {|_k, v| v.keys.include? 'pmid' }
=> {"000081515000015"=>{"pmid"=>"10435530"}}
```


lib/web_of_science/links_client.rb
```diff
@@ -0,0 +1,85 @@
...
+      @username = username
+      @password = password
+      (@username, @password) = Base64.decode64(Settings.WOS.AUTH_CODE).split(':', 2) unless username
+    end
+
+    # Retrieve identifier 'fields' for the record 'ids'
+    # @param [Array<String>] ids
+    # @param [Array<String>] fields (defaults to ['ut', 'doi', 'pmid'])
+    # @return [Hash<String => Hash>]
+    def links(ids, fields = %w(ut doi pmid))
```
mjgiarlo: Would you consider using keyword arguments here, or are you concerned with backwards compatibility?

lib/web_of_science/links_client.rb
```diff
@@ -0,0 +1,85 @@
...
+      ids.uniq!
+      raise ArgumentError, '1-50 ids required' if ids.empty? || ids.count > 50
```
mjgiarlo: If you care about `ids` being `nil`, you could consider changing `#empty?` to `#blank?` though I don't know if this is likely to be problematic in this codebase.

mjgiarlo: If you want to handle all of the problematic cases (not passed an array, passed an empty array, passed an array with > 50 elements, you could change this to:

```ruby
# ...
raise ArgumentError, '1-50 ids required' unless valid_identifier_list?(ids)
# ...

def valid_identifier_list?(ids)
  return false unless ids.is_a? Enumerable
  ids.count.between?(1, 50)
end
```

lib/web_of_science/links_client.rb
```diff
@@ -0,0 +1,85 @@
...
+      raise ArgumentError, 'fields cannot be empty' if fields.empty?
```
mjgiarlo: Consider using `#blank?` instead of `#empty?` to guard against `nil`s here?

lib/web_of_science/links_client.rb
```diff
@@ -0,0 +1,85 @@
...
+          view_paths = ActionView::PathSet.new([File.dirname(__FILE__)])
+          ActionView::Base.new(view_paths, {})
+        end
+      end
+
+      # @param [Array<String>] fields
+      # @return [String]
+      def request_body(ids, fields)
+        renderer.render(
+          file: 'links_request.xml',
```
mjgiarlo: Is this a reference to a file in this codebase?

lib/web_of_science/links_client.rb
```diff
@@ -0,0 +1,85 @@
...
+            cites: ids
+          }
+        )
+      end
+
+      # @param response [Faraday::Response]
+      def response_parse(response)
+        ng = Nokogiri::XML(response.body) { |config| config.strict.noblanks }.remove_namespaces!
+        pairs = ng.xpath('response/fn/map/map').map do |node|
+          [node.attr('name').split('_', 2).last, vals_to_hash(node.children.xpath('val'))]
```
mjgiarlo: There is a lot going on here. Worth breaking parts of this into smaller private methods?

lib/web_of_science/links_client.rb
```diff
@@ -0,0 +1,85 @@
...
+        end
+        pairs.to_h
+      end
+
+      # @param [Nokogiri::XML::Nodeset] vals
+      # @return [Hash<String => String>] namespace to value
+      # Nodeset for [<val name="ut">UT</val>, <val name="doi">DOI</val>]
+      # becomes { 'ut' => 'UT', 'doi' => 'DOI'}
+      def vals_to_hash(vals)
```
mjgiarlo: I appreciate breaking this out into its own method.

lib/web_of_science/links_client.rb
```diff
@@ -0,0 +1,85 @@
+module WebOfScience
```
mjgiarlo: Rename looks good to me.

lib/web_of_science/links_client.rb
atz: You mean copying it from the original?  This is a renamed file with some unnecessary tweaks.

lib/web_of_science/links_client.rb
atz: Yes.  It is moved in this PR.

darrenleeweber:
Obviously this is contentious; I'm now asking for an independent review, as Joe and I clearly do not agree on it.

lib/web_of_science/links_client.rb
darrenleeweber: I'm open to using keyword args, although it might require additional changes to any existing consumers.  There should be very few at this stage, it's a new class in this app.

lib/web_of_science/links_client.rb
darrenleeweber: OK, open to that.  Also note that it runs an `id.uniq` on them so it makes sense to get the uniq set before counting them.

lib/web_of_science/links_client.rb
darrenleeweber: OK

lib/web_of_science/links_client.rb
darrenleeweber: This file is in the same directory as this class in this PR.  The render paths can only traverse this directory in this PR.

darrenleeweber:
```ruby
        wos_ids = %w(000081515000015 000346594100007)
        wos_client = WebOfScience::Client.new(Settings.WOS.AUTH_CODE)
        wos_queries = WebOfScience::Queries.new(wos_client)
        wos_records = wos_queries.retrieve_by_id(wos_ids)
        wos_records.first.identifiers
        # => {"issn"=>"0886-6236", "eissn"=>"1944-9224", "doi"=>"10.1002/2013GB004790", "WosUID"=>"WOS:000346594100007", "WosItemID"=>"000346594100007"}

        wos_records.first.wos_item_id
        #=> "000346594100007"
        links_client = Clarivate::LinksClient.new
        links_client.links([wos_records.first.wos_item_id])
        #=> {"000346594100007"=>{:ut=>"000346594100007", :doi=>"10.1002/2013GB004790"}}
```
Can we merge these two hashes?  No.  Do they use the same keys?  No.  Do we need the `ut` data, no.  These are good reasons to leave the API response data as much alone as possible, assuming that we can trust them to have already normalized these important identifiers.  In other words, I believe they have already normalized the identifier names and documented them for their service(s).

lib/web_of_science/links_client.rb
darrenleeweber: OK, done.

lib/web_of_science/links_client.rb
darrenleeweber: OK, done.

lib/web_of_science/links_client.rb
mjgiarlo: OK. Mostly I was wondering whether it would make sense to freeze this string as a constant (`TEMPLATE_PATH` or some such?), but that's not that big a deal. Thanks, you two.

peetucket:
I agree with all of the changes proposed in this PR, with the exception of the renaming/movement of the Links API client into the Web of Science namespace.  This means the movement of the template from "views" into "lib" folder should still occur, but shouldn't be renamed in the process.  I apologize for this taking so long to review and that it requires further revision - hopefully these requested revisions will not take long.  I will then review and merge.

PR #331, opened 2017-10-26
Every publication_identifier should be attached to a publication

atz:
Otherwise it is not identifying anything!

`inverse_of` is a recommended practice for associations.  Rails tries to auto-detect heuristically, but it is more reliable and efficient to declare the inverse.  See `Bi-directional associations` here:
   https://apidock.com/rails/ActiveRecord/Associations/ClassMethods

Verified there are no records currently (on prod or cap-qa) that would run afoul of this constraint:
```ruby
PublicationIdentifier.where(publication_id:nil).count
=> 0
```
But let's keep it that way.

### DIFF
--- a/app/models/publication_identifier.rb
+++ b/app/models/publication_identifier.rb
@@ -1,3 +1,3 @@
 class PublicationIdentifier < ActiveRecord::Base
-  belongs_to :publication
+  belongs_to :publication, required: true, inverse_of: :publication_identifiers
 end


darrenleeweber:
I've forgotten some AR association details, so reading up on the `Bi-directional associations` here:
https://apidock.com/rails/ActiveRecord/Associations/ClassMethods

It says, "for belongs_to associations has_many inverse associations are ignored" - does this mean it will ignore the inverse because our `Publication#has_many -> PublicationIdentifier`

If it doesn't ignore it, does using it mean that it will kinda aggressively load additional model data (vs. lazy-load the model data)?  I understand that it's trying to use the same model instance across the association, but I don't understand the performance/memory demands entirely.  For example, if we did a complete traversal on all the Publication records, when we get each record instance, would it also already load all the associated PublicationIdentifier records for it?

atz:
My understanding is that `inverse_of` would be ignored if defined on the Publication `has_many :publication_identifiers` statement.  It does not affect what is selected/retrieved for a given query.

In theory, rails was already correctly guessing this `inverse_of`, but it is better to avoid making it guess and be sure.

Issue #317, opened 2017-10-23
Come up with a design proposal for the WoS harvester


peetucket:
consider generalization into a harvester that can work with multiple sources

darrenleeweber:
Connected to #254

darrenleeweber:
Related documents:
- https://docs.google.com/drawings/d/15hOFHC54KDeEHs-4L9i-Q3AQ9AgUeOP1z3caqSnXmvs/edit

darrenleeweber:
A schematic of some elements involved in the harvest (doesn't entirely capture my thoughts on the process, but it helps a little to picture the elements and data flow through the system).

[sul_pub_data_flow.pdf](https://github.com/sul-dlss/sul_pub/files/1422706/sul_pub_data_flow.pdf)

darrenleeweber:
I've made a quick sketch of a WOS harvester in
- https://github.com/sul-dlss/sul_pub/blob/wos-harvest/lib/web_of_science/harvester.rb
- If this first sketch is an OK first approximation, we might proceed with PR #333

PR #333, opened 2017-10-27
WebOfScience::Harvester

### LINKED ISSUE #254: Create new WoS Harvest classes
see #253 for pre-req analysis

### LINKED ISSUE #272: Name queries in the WoS client should follow the same model used in Sciencewire Client

### LINKED ISSUE #301: Write code to determine when to create a new publication in the harvester
Publications harvested from WoS only need a new publication row if they haven't already previously been harvested from Sciencewire.

see #260 for the proposed algorithm and analysis

When we find a matching publication, store the link to the existing publication somehow along with the WOS source record but do not update the pub_hash and provenance

When we don't find an existing publication, we will create it in #316

### LINKED ISSUE #326: Implement logging around Links AMR client
This depends on the question of where Links Client integration happens.  Keep in mind that the Links client as presently conceived as ancillary to WoSClient (where it has its own logging).  Logging may happen at the point of integration (e.g., in WoSClient) or might be added internal to the client.

### LINKED ISSUE #392: Integrate linksAMR client with harvesting
During a harvesting run, for each author we need to look up which records have PMIDs for a given set of WosIDs

darrenleeweber:
Fix #272 to use an `Agent::AuthorName` and `Agent::AuthorInstitution` for WOS searches
Fix #254 to code the business logic for the WOS harvester
Fix #326 to log possible exceptions in the LinksClient calls
Fix #301 to code logic for de-duping new WOS-records to create new Pubs
Fix #392 integrate with LinksAmr
Connected to #317 - design discussions on the WOS harvester

This PR focuses on
- using the AuthorName and AuthorInstitution to issue an author query
- saving WebOfScienceSourceRecords
- integrating links-API data into `WebOfScience::Record#identifiers`
  - [x] depends on #328 - rebase on master when it's merged
  - [x] depends on #370 - rebase on master when it's merged

This PR _does not_ attempt to:
- create new `Publication` records
- create new `PublicationIdentifier` records
- these two goals depend on mapping the PubHash data

TODO:
- [x] WIP to refactor after pairing discussion today (Nov 6th)
- [x] add or enhance specs
- [x] split out some commits into separate PRs
- [x] cleanup commit history
- [x] refactor - extract class for `ProcessRecords`
- [x] identify ways to split up the work
  - create pending tests and ticket work to get that stuff done
  - create Publication records (e.g. mapping data to pub_hash)
  - create PublicationIdentifier records

### DIFF
--- a/lib/web_of_science/harvester.rb
+++ b/lib/web_of_science/harvester.rb
@@ -1,0 +1,145 @@
+module WebOfScience
+
+  # Application logic to harvest publications from Web of Science;
+  # This is the bridge between the WebOfScience API and the SUL-PUB application.
+  # This class is responsible for processing WebOfScience API response data
+  # to integrate it into the application data models.
+  class Harvester
+
+    # Harvest all publications for an author
+    # @param author [Author]
+    # @return [Array<String>] WosUIDs that create Publications
+    def process_author(author)
+      raise(ArgumentError, 'author must be an Author') unless author.is_a? Author
+      process_records author, records_for_author(author)
+    end
+
+    # Harvest DOI publications for an author
+    # @param author [Author]
+    # @param dois [Array<String>] DOI values (not URIs)
+    # @return [Array<String>] WosUIDs that create Publications
+    def process_dois(author, dois)
+      raise(ArgumentError, 'author must be an Author') unless author.is_a? Author
+      raise(ArgumentError, 'dois must be Enumerable') unless dois.is_a? Enumerable
+      # TODO: normalize the dois using altmetrics identifier gem, as in PR #246
+      dois.reject! { |doi| publication_identifier?('doi', doi) }
+      return [] if dois.empty?
+      records = wos_queries.search_by_doi(dois.shift)
+      dois.each { |doi| records.merge_records wos_queries.search_by_doi(doi) }
+      process_records author, records
+    end
+
+    # Harvest PMID publications for an author
+    # @param author [Author]
+    # @param pmids [Array<String>] PMID values (not URIs)
+    # @return [Array<String>] WosUIDs that create Publications
+    def process_pmids(author, pmids)
+      raise(ArgumentError, 'author must be an Author') unless author.is_a? Author
+      raise(ArgumentError, 'pmids must be Enumerable') unless pmids.is_a? Enumerable
+      # TODO: normalize the pmids using altmetrics identifier gem, as in PR #246
+      pmids.reject! { |pmid| publication_identifier?('pmid', pmid) }
+      return [] if pmids.empty?
+      # TODO: define the flow for retrieving this data
+      # PMID ->  links-API to get UID -> WOS API to get record (likely MEDLINE record)
+      #      -> (if that fails) PubMed API for PubMed record
+      # TODO: process_records author, wos_queries.retrieve_by_id(pmids) # Does this work?
+      author.id # this is only here to avoid a rubocop error
+      raise(NotImplementedError, 'There is no processor defined for PMID publications')
+    end
+
+    # Harvest WOS-UID publications for an author
+    # @param author [Author]
+    # @param uids [Array<String>] WOS-UID or WOS-ItemId values (not URIs)
+    # @return [Array<String>] WosUIDs that create Publications
+    def process_uids(author, uids)
+      raise(ArgumentError, 'author must be an Author') unless author.is_a? Author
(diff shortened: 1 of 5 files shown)


atz:
Since the end result touches 2 classes, you probably want some squashing when you feel comfortable with the code state.

atz:
time to rebase

lib/web_of_science/harvester.rb
```diff
@@ -0,0 +1,145 @@
...
+      raise(ArgumentError, 'author must be an Author') unless author.is_a? Author
+      raise(ArgumentError, 'pmids must be Enumerable') unless pmids.is_a? Enumerable
+      # TODO: normalize the pmids using altmetrics identifier gem, as in PR #246
+      pmids.reject! { |pmid| publication_identifier?('pmid', pmid) }
+      return [] if pmids.empty?
+      # TODO: define the flow for retrieving this data
+      # PMID ->  links-API to get UID -> WOS API to get record (likely MEDLINE record)
+      #      -> (if that fails) PubMed API for PubMed record
+      # TODO: process_records author, wos_queries.retrieve_by_id(pmids) # Does this work?
+      author.id # this is only here to avoid a rubocop error
```
peetucket: so this method is not actually intended to be implemented and returns nothing?  is that why there is a "raise" at the end?  and curious as to what the rubocop error is?

lib/web_of_science/process_records.rb
```diff
@@ -0,0 +1,134 @@
...
+      # Save and select new WebOfScienceSourceRecords
+      # @param [Array<WebOfScience::Record>]
+      # @return [Array<WebOfScience::Record>]
+      def save_wos_records(records)
+        # IMPORTANT: add nothing to PublicationIdentifiers here, or new_records will reject them
+        return [] if records.empty?
+        # We only want the 'pmid' for "WOS" records ("MEDLINE" records have one already)
+        process_links(records.select { |rec| rec.database == 'WOS' })
+        records.select do |rec|
+          saved = WebOfScienceSourceRecord.new(source_data: rec.to_xml).save!
```
peetucket: isn't it enough to do something like this instead of using the intermediate `saved` variable?  `records.select {|rec| WebOfScienceSourceRecord.new(source_data: rec.to_xml).save! }`

lib/web_of_science/process_records.rb
```diff
@@ -0,0 +1,134 @@
...
+    private
+
+      attr_reader :author
+      attr_reader :records
+
+      # ----
+      # Record filters and data flow steps
+
+      # @return [Array<String>] WosUIDs that create a new Publication
+      def create_publications
```
peetucket: i still find this set of method names not totally clear as to what they are intending to do (I know now because we discussed it, but I can see it being unclear for someone else who comes to the code) .... but don't have better suggestions, I don't think we should dwell on it here but maybe iterate on it later as we continue to work on this class

lib/web_of_science/process_records.rb
darrenleeweber: Thanks, aware of this.  This section will be changed to add identifiers to the src-record, but that TODO item depends on doing the work to add a migration to the src-record to add the fields for the identifiers.  Unlike other src-models, this one needs outside assistance to save some of the identifiers from the links-API.  I'm aware of this and left the intermediate variable and TODO note there intentionally as a reminder and sign-post that this is where a pending revision will occur in the near sprints.

lib/web_of_science/process_records.rb
darrenleeweber: OK, one interation is not enough it seems, but in the absence of better ideas let's move on with this for now.

lib/web_of_science/harvester.rb
darrenleeweber: Let's ticket the work on PMID harvesting via the new WOS-API when we are starting out with a PMID and we do not have a WOS record yet.  The process flow for this needs to be worked out.  Nevertheless, this is the placeholder for this, to declare an intention to support this public API.

peetucket: https://github.com/sul-dlss/sul_pub/issues/408

PR #334, opened 2017-10-28
Stronger validation for Contribution model

atz:
Existing validations were ad hoc and required the caller to interrogate potential values before instantiating the model. Afterwards model state was not validatable.

ActiveRecord validations offer a more comprehensive and reliable approach. This PR guarantees `publication` and `author` associations (already true for prod data). It also validates values for `visibility` and `status` in the conventional way, demonstrated in tests. See:
  http://guides.rubyonrails.org/active_record_validations.html

Note: this does NOT make `visibility` and `status` values required, because production data has already allowed `nil` values, even for `featured` (a boolean field):

```ruby
pry(main)> Contribution.group(:status).count
=> {nil=>4, "approved"=>318802, "denied"=>134889, "new"=>302397,
"unknown"=>138209}
pry(main)> Contribution.group(:visibility).count
=> {nil=>9, "private"=>129817, "public"=>764475}
pry(main)> Contribution.group(:featured).count
=> {nil=>4, false=>864760, true=>29539}
```

We should discuss what we want to do about that (if anything).

I had to reposition the declaration of constants to be before declaration of validations, but otherwise this is pretty lightweight revision to the model. 

Tests were cleaned up to avoid unnecessary intermediate assignment of hashes.  Lastly, a method unused anywhere except tests was removed.

### DIFF
--- a/app/models/contribution.rb
+++ b/app/models/contribution.rb
@@ -1,17 +1,23 @@
 class Contribution < ActiveRecord::Base
+  # Allowed values for visibility
+  VISIBILITY_VALUES = %w(public private).freeze
+  # Allowed values for status
+  STATUS_VALUES = %w(approved denied new unknown).freeze
 
-  def cap_profile_id
-    (author.cap_profile_id if author) || self[:cap_profile_id]
-  end
+  belongs_to :publication, required: true, inverse_of: :contributions
+  belongs_to :author, required: true, inverse_of: :contributions
 
-  belongs_to :publication
-  belongs_to :author
-  # has_one :publication_identifier, :foreign_key => "publication_id"
   has_one :publication_identifier, -> { where("publication_identifiers.identifier_type = 'PublicationItemId'") },
           class_name: 'PublicationIdentifier',
           foreign_key: 'publication_id',
           primary_key: 'publication_id'
-  # has_one :population_membership, :foreign_key => "author_id"
+
+  validates :visibility, inclusion: { in: VISIBILITY_VALUES }, allow_nil: true # TODO: disallow nil
+  validates :status, inclusion: { in: STATUS_VALUES }, allow_nil: true         # TODO: disallow nil
+
+  def cap_profile_id
+    (author.cap_profile_id if author) || self[:cap_profile_id]
+  end
 
   def self.authorship_valid?(authorship)
     author_valid?(authorship) && valid_fields?(authorship)
@@ -35,17 +41,11 @@ class Contribution < ActiveRecord::Base
       visibility_valid?(contrib)
   end
 
-  # Allowed values for visibility
-  VISIBILITY_VALUES = %w(public private).freeze
-
   # @return [Boolean]
   def self.visibility_valid?(contrib)
     contrib = contrib.with_indifferent_access
     VISIBILITY_VALUES.include? contrib[:visibility].to_s.downcase
   end
-
-  # Allowed values for status
-  STATUS_VALUES = %w(approved denied new unknown).freeze
 
   # @return [Boolean]
   def self.status_valid?(contrib)
--- a/app/models/contribution.rb
+++ b/app/models/contribution.rb
@@ -15,6 +15,15 @@ class Contribution < ActiveRecord::Base
   validates :visibility, inclusion: { in: VISIBILITY_VALUES }, allow_nil: true # TODO: disallow nil
   validates :status, inclusion: { in: STATUS_VALUES }, allow_nil: true         # TODO: disallow nil
 
+  after_initialize :init
+
+  # apply some default values and coercions (lowercasing)
+  def init
+    self.featured   = false if featured.nil? # can't use ||= with boolean
+    self.status     = status.downcase if status
+    self.visibility = visibility.downcase if visibility
+  end
+
   def cap_profile_id
     (author.cap_profile_id if author) || self[:cap_profile_id]
   end
--- a/app/models/publication.rb
+++ b/app/models/publication.rb
@@ -89,11 +89,13 @@ class Publication < ActiveRecord::Base
   end
 
   def self.find_by_doi(doi)
-    PublicationIdentifier.includes(:publication).where(identifier_type: 'doi', identifier_value: doi).map(&:publication)
+    Publication.includes(:publication_identifiers)
+               .find_by("publication_identifiers.identifier_type": 'doi', "publication_identifiers.identifier_value": doi)
   end
 
-  def self.find_by_pmid_in_pub_id_table(pmid)
-    PublicationIdentifier.includes(:publication).where(identifier_type: 'pmid', identifier_value: pmid).map(&:publication)
+  def self.find_by_pmid_pub_id(pmid)
+    Publication.includes(:publication_identifiers)
+               .find_by("publication_identifiers.identifier_type": 'pmid', "publication_identifiers.identifier_value": pmid)
   end
 
   # @return [Publication] new object
--- a/app/models/publication.rb
+++ b/app/models/publication.rb
@@ -43,31 +43,7 @@ class Publication < ActiveRecord::Base
     autosave: true,
     dependent: :destroy,
     after_add: :pubhash_needs_update!,
-    after_remove: :pubhash_needs_update! do
-    # See also: update_any_new_contribution_info_in_pub_hash_to_db
-    def build_or_update(author, contribution_hash = {})
-      # Assign or update these contribution attributes
-      # Ensure the contribution attributes contain the right author
-      contribution_hash[:author_id] = author.id
-      contribution_hash[:cap_profile_id] = author.cap_profile_id if author.cap_profile_id.present?
-      contrib = find_or_initialize_by(author_id: author.id) do |new_contrib|
-        new_contrib.assign_attributes(contribution_hash)
-        new_contrib.publication = proxy_association.owner
-      end
-      if contrib.persisted?
-        # Update an existing contribution
-        contrib.update(contribution_hash)
-      else
-        contrib.save
-        # SELF is an array of Contributions
-        self << contrib unless include? contrib
-      end
-      # The `proxy_association.owner` is a Publication instance, so
-      # set a trigger that will force it to update it's pub_hash data.
-      proxy_association.owner.pubhash_needs_update!
-      contrib
-    end
-  end
+    after_remove: :pubhash_needs_update!
 
   serialize :pub_hash, Hash
 
@@ -316,9 +292,8 @@ class Publication < ActiveRecord::Base
       Array(pub_hash[:authorship]).each do |contrib|
         hash_for_update = contrib.slice(:status, :visibility, :featured)
         # Find or create an Author of the contribution
-        author_id = contrib[:sul_author_id]
         cap_profile_id = contrib[:cap_profile_id]
-        author = Author.find_by_id(author_id)
+        author = Author.find_by_id(contrib[:sul_author_id])
         if cap_profile_id.present?
           author ||= Author.find_by_cap_profile_id(cap_profile_id)
           author ||= begin
@@ -335,11 +310,7 @@ class Publication < ActiveRecord::Base
         hash_for_update[:cap_profile_id] = author.cap_profile_id if author.cap_profile_id.present?
         contrib = contributions.where(author_id: author.id).first_or_initialize
         contrib.assign_attributes(hash_for_update)
-        if contrib.persisted?
-          contrib.save
-        else
-          contributions << contrib unless contributions.include? contrib
-        end
+        contrib.save if contrib.changed?
       end
       true
     end
--- a/app/models/publication.rb
+++ b/app/models/publication.rb
@@ -133,20 +133,20 @@ class Publication < ActiveRecord::Base
   def update_from_pubmed
     return false if pmid.blank?
     pm_source_record = PubmedSourceRecord.find_by_pmid(pmid)
+    return false unless pm_source_record
     pm_source_record.pubmed_update
     rebuild_pub_hash
-    save
   end
 
   def update_from_sciencewire
     return false if sciencewire_id.blank?
     sw_source = SciencewireSourceRecord.find_by_sciencewire_id(sciencewire_id)
+    return false unless sw_source
     sw_source.sciencewire_update
     rebuild_pub_hash
-    save
   end
 
-  # @return [self]
+  # @return [Boolean] true if .save is successful
   def rebuild_pub_hash
     if sciencewire_id
       sw_source_record = SciencewireSourceRecord.find_by_sciencewire_id(sciencewire_id)
@@ -156,7 +156,7 @@ class Publication < ActiveRecord::Base
       self.pub_hash = pubmed_source_record.source_as_hash
     end
     sync_publication_hash_and_db
-    self
+    save
   end
 
   def delete!
--- a/app/models/publication.rb
+++ b/app/models/publication.rb
@@ -50,13 +50,14 @@ class Publication < ActiveRecord::Base
       # Ensure the contribution attributes contain the right author
       contribution_hash[:author_id] = author.id
       contribution_hash[:cap_profile_id] = author.cap_profile_id if author.cap_profile_id.present?
-      contrib = where(author_id: author.id).first_or_initialize
+      contrib = find_or_initialize_by(author_id: author.id) do |new_contrib|
+        new_contrib.assign_attributes(contribution_hash)
+        new_contrib.publication = proxy_association.owner
+      end
       if contrib.persisted?
         # Update an existing contribution
         contrib.update(contribution_hash)
       else
-        # Add a new contribution
-        contrib.assign_attributes(contribution_hash)
         contrib.save
         # SELF is an array of Contributions
         self << contrib unless include? contrib
@@ -311,11 +312,7 @@ class Publication < ActiveRecord::Base
 
     def update_any_new_contribution_info_in_pub_hash_to_db
       Array(pub_hash[:authorship]).each do |contrib|
-        hash_for_update = {
-          status: contrib[:status],
-          visibility: contrib[:visibility],
-          featured: contrib[:featured]
-        }
+        hash_for_update = contrib.slice(:status, :visibility, :featured)
         # Find or create an Author of the contribution
         author_id = contrib[:sul_author_id]
         cap_profile_id = contrib[:cap_profile_id]
--- a/app/models/publication.rb
+++ b/app/models/publication.rb
@@ -74,7 +74,7 @@ class Publication < ActiveRecord::Base
                .find_by("publication_identifiers.identifier_type": 'pmid', "publication_identifiers.identifier_value": pmid)
   end
 
-  # @return [Publication] new object
+  # @return [Publication] new object, unsaved
   def self.build_new_manual_publication(pub_hash, original_source_string, provenance = Settings.cap_provenance)
     existingRecord = UserSubmittedSourceRecord.find_or_initialize_by_source_data(original_source_string)
     if existingRecord && existingRecord.publication
@@ -88,18 +88,16 @@ class Publication < ActiveRecord::Base
   def update_manual_pub_from_pub_hash(incoming_pub_hash, original_source_string, provenance = Settings.cap_provenance)
     incoming_pub_hash[:provenance] = provenance
     self.pub_hash = incoming_pub_hash.dup
-    r = user_submitted_source_records.first || UserSubmittedSourceRecord.find_or_initialize_by_source_data(original_source_string)
+    match = UserSubmittedSourceRecord.find_by_source_data(original_source_string)
+    match.publication = self if match # we may still throw this out w/o saving
+    r = user_submitted_source_records.first || match || user_submitted_source_records.build
     r.assign_attributes(
       is_active: true,
       source_data: original_source_string,
       title: title,
       year: year
     )
-    if r.new_record?
-      self.user_submitted_source_records = [r]
-    else
-      r.save
-    end
+    self.user_submitted_source_records = [r] if match # match is the only USSR not found/built via association
     update_any_new_contribution_info_in_pub_hash_to_db
     pubhash_needs_update! if persisted?
     self
@@ -119,11 +117,13 @@ class Publication < ActiveRecord::Base
     self
   end
 
+  # The expecation is that every time this method gets called, a save is about to happen,
+  # either because it is part of before_save callbacks or because the caller does it explicitly.
+  # Subparts are relieved from the burden of directly calling save themselves.
   # @return [self]
   def sync_publication_hash_and_db
     rebuild_authorship
-    sync_identifiers_in_pub_hash_to_db
-    add_all_identifiers_in_db_to_pub_hash
+    sync_identifiers_in_pub_hash
     set_sul_pub_id_in_hash if persisted?
     update_formatted_citations
     @pubhash_needs_update = false
@@ -250,31 +250,23 @@ class Publication < ActiveRecord::Base
 
   private
 
-    def add_all_identifiers_in_db_to_pub_hash
-      publication_identifiers.reload if persisted?
-      pub_hash[:identifier] = publication_identifiers.map(&:identifier)
-    end
-
-    def sync_identifiers_in_pub_hash_to_db
+    # doesn't actually write the Pub to DB, presumed to be part of before_save callback or explicit save
+    def sync_identifiers_in_pub_hash
       incoming_types = Array(pub_hash[:identifier]).map { |id| id[:type] }
       publication_identifiers.each do |id|
         next if id.identifier_type =~ /^legacy_cap_pub_id$/i # Do not delete legacy_cap_pub_id
-        id.delete unless incoming_types.include? id.identifier_type
+        publication_identifiers.delete(id) unless incoming_types.include?(id.identifier_type)
       end
 
       Array(pub_hash[:identifier]).each do |identifier|
         next if identifier[:type] =~ /^SULPubId$/i
-        i = publication_identifiers.find { |x| x.identifier_type == identifier[:type] } || PublicationIdentifier.new
-        i.assign_attributes certainty: 'confirmed',
-                            identifier_type: identifier[:type],
-                            identifier_value: identifier[:id],
-                            identifier_uri: identifier[:url]
-        if i.persisted?
-          i.save
-        else
-          publication_identifiers << i unless publication_identifiers.include? i
-        end
+        i = publication_identifiers.find { |x| x.identifier_type == identifier[:type] } # find includes not yet saved pub ids
+        i ||= publication_identifiers.find_or_initialize_by(identifier_type: identifier[:type])
+        i.certainty        = 'confirmed'
+        i.identifier_value = identifier[:id]
+        i.identifier_uri   = identifier[:url]
       end
+      pub_hash[:identifier] = publication_identifiers.map(&:identifier)
     end
 
     def add_any_pubmed_data_to_hash
--- a/lib/pubmed_harvester.rb
+++ b/lib/pubmed_harvester.rb
@@ -1,4 +1,3 @@
-
 class PubmedHarvester
   # Searches locally, then ScienceWire, then Pubmed for a publication by PubmedId
   # @param [String] pmid PubmedId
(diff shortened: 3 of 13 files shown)


app/models/contribution.rb
```diff
@@ -60,10 +60,6 @@ def self.featured_valid?(contrib)
     contrib[:featured].to_s =~ /true|false/i ? true : false
   end
 
-  def self.find_or_create_by_author_and_publication(author, publication)
-    find_or_create_by(author_id: author.id, publication_id: publication.id)
-  end
-
```
darrenleeweber: Interesting, cannot find where this was called, e.g. on `master`,
```
$ git grep 'find_or_create_by_author_and_publication'
app/models/contribution.rb:  def self.find_or_create_by_author_and_publication(author, publication)
spec/models/contribution_spec.rb:  describe '.find_or_create_by_author_and_publication' do
spec/models/contribution_spec.rb:      described_class.find_or_create_by_author_and_publication(
spec/models/contribution_spec.rb:      contrib = described_class.find_or_create_by_author_and_publication(
```

spec/models/contribution_spec.rb
```diff
@@ -12,17 +10,11 @@
...
-  let(:authorship) do
-    # parse the authorship_json because this is what the sul-pub API does
-    JSON.parse(authorship_json)
-  end
+  # return JSON because this is what the sul-pub API receives
+  let(:authorship_json) { pub_with_contrib.pub_hash[:authorship].first.to_json }
+  let(:authorship) { JSON.parse(authorship_json) } # parse authorship_json like sul-pub API
 
-  let(:subject) { pub_with_contrib.contributions.first }
+  subject { pub_with_contrib.contributions.first }
```
darrenleeweber: This comment is not a requirement, but an FYI and optional change; consider http://blog.davidchelimsky.net/blog/2012/05/13/spec-smell-explicit-use-of-subject/; that is, we might call it `contribution` instead of `subject`?  I think this comes up in `reek` analysis.  I stopped using `subject` a while ago, but it's not a rule to use it or not use an explicit subject.  I don't fully understand the smell, but IIRC it has something to do with rspec using the `subject` internally and possible conflicts with that.

app/models/contribution.rb
darrenleeweber:
```
$ git grep 'find_or_create_by'
app/api/sul_bib/authorship_api.rb:          pub = Publication.find_or_create_by_pmid(pmid)
app/api/sul_bib/authorship_api.rb:          pub = Publication.find_or_create_by_sciencewire_id(sw_id)
app/models/contribution.rb:  def self.find_or_create_by_author_and_publication(author, publication)
app/models/contribution.rb:    find_or_create_by(author_id: author.id, publication_id: publication.id)
app/models/publication.rb:  def self.find_or_create_by_pmid(pmid)
app/models/publication.rb:  def self.find_or_create_by_sciencewire_id(sw_id)
lib/tasks/pubmed.rake:        pub=Publication.find_or_create_by_pmid(pmid)
spec/api/sul_bib/authorship_api_spec.rb:          expect(Publication).to receive(:find_or_create_by_pmid).and_return(nil)
spec/api/sul_bib/authorship_api_spec.rb:          expect(Publication).to receive(:find_or_create_by_sciencewire_id).and_return(nil)
spec/models/contribution_spec.rb:  describe '.find_or_create_by_author_and_publication' do
spec/models/contribution_spec.rb:    it 'calls find_or_create_by_author_id_and_publication_id' do
spec/models/contribution_spec.rb:      expect(described_class).to receive(:find_or_create_by)
spec/models/contribution_spec.rb:      described_class.find_or_create_by_author_and_publication(
spec/models/contribution_spec.rb:      contrib = described_class.find_or_create_by_author_and_publication(
```

darrenleeweber: Seems to be a convenience method that is not used anywhere.
```
$ git grep 'contribution.find_or_create_by'  # nothing
$ git grep 'Contribution.find_or_create_by'  # nothing
$ git grep 'contrib*.find_or_create_by'  # nothing
$ git grep 'cont*.find_or_create_by'  # nothing
```

darrenleeweber: Agreed - seems to be OK to remove it.

spec/models/contribution_spec.rb
```diff
@@ -53,12 +45,21 @@
...
+  context 'ActiveRecord validation' do
+    let(:params) { { publication_id: pub_with_contrib.id, author_id: subject.author.id, visibility: 'private', status: 'approved' } }
+    it 'constrains field values' do
+      expect { described_class.create!(params) }.not_to raise_error
+      expect { described_class.create! }.to raise_error ActiveRecord::RecordInvalid
+      expect { described_class.create!(params.merge(visibility: 'bad'))   }.to raise_error ActiveRecord::RecordInvalid
+      expect { described_class.create!(params.merge(status: 'illegal'))   }.to raise_error ActiveRecord::RecordInvalid
+      expect { described_class.create!(params.merge(author: nil))         }.to raise_error ActiveRecord::RecordInvalid
+      expect { described_class.create!(params.merge(publication_id: nil)) }.to raise_error ActiveRecord::RecordInvalid
     end
```
darrenleeweber: My IDE always complains about more than one expectation per spec, but this is all related stuff.  I suppose the best practice is to have one expect per spec, so they can't interfere or short-circuit one another.  For example, if line 52 were to fail, nothing else would execute.  I think that when you know all the expectations build on one another, like stepping stones, it makes sense to combine them and it doesn't matter if the spec short-circuits.  In this spec, I'm wondering if they are dependent or independent specs.  I could see an argument for separating the specs for each field.  If your open to separating them, that might be better.

darrenleeweber:
The spec failures indicate this change breaks the API.  That's not a bad thing, in itself.  Obviously it's important to fix those specs, so that seems like it requires additional changes in this PR to the API code.  Only for this reason, updating the review to request these API changes to fix those specs.  Then, I should re-review it.

spec/models/contribution_spec.rb
atz: I much prefer these as is.  It would be different if they were performing operations on a common object, but they are independent block executions on the constructor that raise exceptions, so there is literally no way for them to side-effect each other.  They are testing permutations of the same conditions, so giving them separate descriptions is unnecessary, imo.

atz:
At least some of the problem is just that specs were using `status` values like `x` or `y`.

spec/models/contribution_spec.rb
darrenleeweber: Do you care if the spec execution stops at some point and does not test the others?  If it does stop, hopefully the line number is sufficient to be able to track which one fails.  This thing is no big deal, IMO, just noting an option to pursue best practice that recommends one-expect per spec.

spec/models/contribution_spec.rb
atz: Yeah, I read that years ago, and I use named subjects where the specs are large and need differentiation.  Specifying `subject` is intended by rspec, and there is no internal conflict presented.   The anti-pattern is overriding `subject` with a `let` statement like the tests currently do:
```ruby
let(:subject) { pub_with_contrib.contributions.first }
```
That is weird.

darrenleeweber: Yep, replacing `let(:subject)` with just `subject` is better, even better is `subject(:name)` whenever any spec needs to explicitly get it's mits on `subject` (I mean `name`).

atz:
Ignoring CodeClimate complaining about a method that I renamed.

app/models/contribution.rb
```diff
@@ -35,18 +41,12 @@ def self.valid_fields?(contrib)
       visibility_valid?(contrib)
   end
 
-  # Allowed values for visibility
-  VISIBILITY_VALUES = %w(public private).freeze
-
```
darrenleeweber: So `visibility_valid?` and `status_valid?` did a bunch of data munging for symbol/string compatibility.  Are these methods still required now that the model has validations?

app/models/publication.rb
```diff
@@ -50,13 +50,14 @@ def build_or_update(author, contribution_hash = {})
       # Ensure the contribution attributes contain the right author
       contribution_hash[:author_id] = author.id
       contribution_hash[:cap_profile_id] = author.cap_profile_id if author.cap_profile_id.present?
-      contrib = where(author_id: author.id).first_or_initialize
+      contrib = find_or_initialize_by(author_id: author.id) do |new_contrib|
+        new_contrib.assign_attributes(contribution_hash)
+        new_contrib.publication = proxy_association.owner
+      end
```
darrenleeweber: What does `new_contrib.publication = proxy_association.owner` do when the `find_or_initialize_by` returns an existing record and the publication association is already established?

app/models/contribution.rb
atz: Currently, yes, because the `authorship_api` still ties into them.  Conceivably, those uses could be converted once this is in place.

app/models/publication.rb
```diff
@@ -89,11 +89,13 @@ def self.find_or_create_by_sciencewire_id(sw_id)
...
+    Publication.includes(:publication_identifiers)
+               .find_by("publication_identifiers.identifier_type": 'doi', "publication_identifiers.identifier_value": doi)
   end
 
-  def self.find_by_pmid_in_pub_id_table(pmid)
-    PublicationIdentifier.includes(:publication).where(identifier_type: 'pmid', identifier_value: pmid).map(&:publication)
+  def self.find_by_pmid_pub_id(pmid)
+    Publication.includes(:publication_identifiers)
+               .find_by("publication_identifiers.identifier_type": 'pmid', "publication_identifiers.identifier_value": pmid)
   end
```
darrenleeweber: Why not rename it to `find_by_pmid`?

In general, the revised syntax is not much clearer to read, it's more complex to read IMO.  But, if it works, fine.

lib/pubmed_harvester.rb
```diff
@@ -1,12 +1,11 @@
-
 class PubmedHarvester
   # Searches locally, then ScienceWire, then Pubmed for a publication by PubmedId
   # @param [String] pmid PubmedId
   # @return [Array<Hash>] publication in BibJson format.  Usually for Publication sourcelookup requests
   #   If no publication found, returns an empty Array
   def self.search_all_sources_by_pmid(pmid)
     result = Publication.where(pmid: pmid).to_a # TODO: index manual and batch with pmid?
-    result = Publication.find_by_pmid_in_pub_id_table(pmid) if result.empty?
+    result = [Publication.find_by_pmid_pub_id(pmid)].compact if result.empty?
```
darrenleeweber: This is where things get really weird.

spec/models/publication_spec.rb
```diff
@@ -392,16 +392,15 @@
...
-    it 'returns an array of Publications that have this doi' do
+    it 'returns one Publication that has this doi' do
       publication.pub_hash = { identifier: [{ type: 'doi', id: '10.1016/j.mcn.2012.03.008', url: 'https://dx.doi.org/10.1016/j.mcn.2012.03.008' }] }
       publication.send(:sync_identifiers_in_pub_hash_to_db)
-      res = Publication.find_by_doi '10.1016/j.mcn.2012.03.008'
-      expect(res.size).to eq(1)
-      expect(res.first.id).to eq(publication.id)
+      res = Publication.find_by_doi('10.1016/j.mcn.2012.03.008')
+      expect(res.id).to eq(publication.id)
     end
```
darrenleeweber: Replace `publication.send(:sync_identifiers_in_pub_hash_to_db)` with `publication.save`

app/models/contribution.rb
```diff
@@ -15,6 +15,15 @@ class Contribution < ActiveRecord::Base
...
+  # apply some default values and coercions (lowercasing)
+  def init
+    self.featured   = false if featured.nil? # can't use ||= with boolean
+    self.status     = status.downcase if status
+    self.visibility = visibility.downcase if visibility
+  end
+
   def cap_profile_id
     (author.cap_profile_id if author) || self[:cap_profile_id]
   end
```
darrenleeweber: Squash this with validations commit.

app/models/publication.rb
```diff
@@ -43,31 +43,7 @@ class Publication < ActiveRecord::Base
...
-        # SELF is an array of Contributions
-        self << contrib unless include? contrib
-      end
-      # The `proxy_association.owner` is a Publication instance, so
-      # set a trigger that will force it to update it's pub_hash data.
-      proxy_association.owner.pubhash_needs_update!
-      contrib
-    end
-  end
+    after_remove: :pubhash_needs_update!
```
darrenleeweber: - Removing this seems to intersect with a prior commit that changed the contrib, maybe drop that prior commit?
- It's difficult to see that moving this to the api code has captured all the logic accurately, like what about checking if the cap_profile_id is present or not?
- The `after_remove` shouldn't be touched here - looks like accidental whitespace change.

app/models/publication.rb
```diff
@@ -335,11 +310,7 @@ def update_any_new_contribution_info_in_pub_hash_to_db
         hash_for_update[:cap_profile_id] = author.cap_profile_id if author.cap_profile_id.present?
         contrib = contributions.where(author_id: author.id).first_or_initialize
         contrib.assign_attributes(hash_for_update)
-        if contrib.persisted?
-          contrib.save
-        else
-          contributions << contrib unless contributions.include? contrib
-        end
+        contrib.save if contrib.changed?
       end
```
darrenleeweber: Some of this logic avoids duplicating contributions in the pub_hash, a past bug that was not easy to find and fix.  It's difficult to see how this change impacts that behavior, it's important that any prior bug fixes are not broken by the change.

spec/models/publication_spec.rb
```diff
@@ -152,7 +152,6 @@
     it 'should sync existing authors in the pub hash to contributions in the db' do
       publication.pub_hash = { authorship: [{ status: 'new', sul_author_id: author.id }] }
       publication.send(:update_any_new_contribution_info_in_pub_hash_to_db)
-      publication.save
       expect(publication.contributions.size).to eq(1)
```
darrenleeweber: Why removing the `publication.save` in several places?

spec/models/publication_spec.rb
```diff
@@ -332,21 +329,6 @@
...
-      expect(c.status).to eq('unknown')
-    end
-
-    it 'should update a contribution record if the association exists' do
-      publication.contributions.build_or_update author
-      c = publication.contributions.build_or_update(author, status: 'new')
-      expect(c.author).to eq(author)
-      expect(c.status).to eq('new')
-    end
-  end
```
darrenleeweber: If it's not in the method any more, the equivalent behavior still needs to be tested somehow.

app/models/publication.rb
atz: Right.  I think I preserved everything we want out of this.  Obviously, by definition, if the object isn't changed then there is nothing (new or edited) to save.

app/models/publication.rb
atz: Correct.  It makes this `find_by` the same as *other* `find_by` methods (which don't require documentation).

lib/pubmed_harvester.rb
atz: Less weird than having `find_by` methods in an AR class that don't return 1 record or nil, imo.

spec/models/publication_spec.rb
atz: It isn't part of what is being tested here.  I.E., `publication.contributions.size` and `publication.contributions.last` don't rely on having been saved (and shouldn't!)

spec/models/publication_spec.rb
atz: This is covered by the API tests.

app/models/publication.rb
atz: The `do` part of the `after_remove` line was removed (i.e. the start of the block now removed).

atz: The lines moved to here:
https://github.com/sul-dlss/sul_pub/pull/334/files#diff-c6ef41c2e0bc46fa80c88d0c5ccb026bR27

Correct me if I'm wrong, but we're assigning the contribution the author's ID (as `author_id`), so my sense is it would be counterproductive to simultaneously retain some *other* author's `cap_profile_id`.  This author is taking control of the contrib, so it should be tied to their respective IDs (or nobody's).

app/models/publication.rb
atz: @darrenleeweber Since `Publication` already has `pmid` field, `find_by_pmid` is a dynamic method that already exists!  This method's usefulness is that it searches `publication_identifiers` instead of `publications` table directly.

spec/api/sul_bib/publications_api_spec.rb
```diff
@@ -154,6 +154,8 @@ def validate_authorship(pub_hash, submission)
 
   describe 'POST /publications' do
     context 'when valid post' do
+      let(:submission) { JSON.parse(valid_json_for_post) }
```
darrenleeweber: :+1:

spec/api/sul_bib/publications_api_spec.rb
```diff
@@ -192,23 +193,18 @@ def validate_authorship(pub_hash, submission)
 
       it 'creates a matching pub_hash in the publication record from the posted bibjson' do
         post_valid_json
-        pub_hash = Publication.last.reload.pub_hash
-        submission = JSON.parse(valid_json_for_post)
-        validate_authorship(pub_hash, submission)
+        validate_authorship(Publication.last.pub_hash, submission)
       end
```
darrenleeweber: No `reload` required, huh.  Puzzled, not sure how that impacts the pub_hash content.  (Seems like maybe a case for another let to get the pub_hash content.)

spec/api/sul_bib/publications_api_spec.rb
```diff
@@ -75,7 +75,7 @@
   end
```
darrenleeweber: Could be squashed with other commit on this spec file.

darrenleeweber:
Note to self - I really have to do some homework on https://github.com/sul-dlss/sul_pub/pull/334/commits/1a76e20c1eea04a6386bb496cbf63025af1f62c8 because it seems to change the return value from a `Publication` to a `Boolean` by adding the `save` in the rebuild-pubhash method.  Also need to check if anywhere there is a need to update the pubhash without saving.

spec/api/sul_bib/publications_api_spec.rb
atz: Yeah, you can watch in console when `Publication.last` hits the database.  Of course it has to check every time you call it, since other processes might have written new records.  Point being the record is freshly retrieved, so `reload` doesn't help.

app/models/publication.rb
darrenleeweber: IIRC (not sure), basically we don't trust the CAP-team to call our API with the same params all the time and, maybe, sometimes they have our `author.id` (as something like `sul_author_id`) and sometimes they don't but they send the `cap_profile_id`; and sometimes they only send one or the other; also if we only get a `cap_profile_id` and we can't find an author.... we try to place nicely.  I dunno, maybe I'm being paranoid or there have been weird bugs fixed at some point.  Maybe I just need to get dirty with this in my console to see what's going on.

lib/pubmed_harvester.rb
darrenleeweber: Yea, `find_by` does imply it should return a record or nil.  Just a pity the callers now have to do some weird array wrapper stuff or the methods could be renamed to avoid the `find_by`, which the new rubocop is complaining about too.  It's sorta convenient to throw this stuff onto Publication to sorta make it behave like it has a `pmid` or a `doi` field when it doesn't really.  Although I don't love the caller-array munging so much, it's fine.  I suppose thinking about rubocop fixes could lead somewhere like `Publication.get_by_pmid` or `Publication.by_pmid` or  `Publication.for_pmid` and similar equally weird stuff.  Otherwise, it all disappears and callers have to do the joins themselves.  Dunno if a `through` could be mangled either because the PublicationIdentifiers don't have any `pmid` or `doi` fields either; maybe a through could have a block on it, dunno.

app/models/publication.rb
darrenleeweber: The bit that has always bugged me is the
```
contributions << contrib unless contributions.include? contrib
```
IIRC, I added the `unless contributions.include? contrib` to fix a dup bug ages ago when I first started work on sul_pub.

spec/models/publication_spec.rb
darrenleeweber: I know this is actually in the pub spec, so it's sorta OK, but I can't help wondering if the `send` could be removed and replaced with the `save` instead.  Would be cool if the spec respected the callbacks too.  In this case, seems like the spec is just asking pub to save a contribution and it should be doing that in a before block or something, unless the spec is really targeting that private method specifically.   (Can't see all the context from this github view.)  If it's generally testing the callbacks as a whole, it should use `save`, but if it's really focused on just that private method, OK, use the send.  If it's the latter, seems like it should be using an rspec change pattern, like
```
expect { pub-with-authorship-to-save-a-contribution }.to change { Contribution.count }
```

spec/models/publication_spec.rb
darrenleeweber: OK, didn't see that right away.

app/models/publication.rb
atz: Those values aren't coming straight from query params though, they are coming out of a valid `Author` object in the database.  So we can trust them.

app/models/publication.rb
atz: I get that.  Since that time we started using `contributions.where(author_id: author.id).first_or_initialize`, so there are now two possibilities:
- If we have a match, it would already be in the association, so we are updating instead of inserting
- If we built a new contrib, it is *also* already in the association.  It can't be a dup because it **must** have a different author_id.

I think the workaround was there because duplication happened when new objects (i.e. `contrib.persisted? == false`) were manually being added to the association.  Now we avoid manually inserting and ActiveRecord handles it.  If you can imagine a test case for `update_any_new_contribution_info_in_pub_hash_to_db` that would trigger duplication, I'd be all for it though!

spec/models/publication_spec.rb
atz: No, the `send` IS the method being tested.  This is a unit test on the private function that is part of the save callbacks.  `.save` is tested elsewhere.

atz:
I touch all those same lines in this PR.

app/models/publication.rb
darrenleeweber: Oh, it does have a `pmid` field already.  Very weird that this util method even exists then.  Maybe all the callers of this method should be using `Publication.find_by_pmid` instead.  Since it's just one place, might be a good case for removing this method entirely and shove the responsibility for looking up the PublicationIdentifiers in the pubmed_harvester - could make that code a whole lot clearer too.  Can't find any other uses....
```
$ git grep 'find_by_pmid_in_pub_id_table'
app/models/publication.rb:94:  def self.find_by_pmid_in_pub_id_table(pmid)
lib/pubmed_harvester.rb:9:    result = Publication.find_by_pmid_in_pub_id_table(pmid) if result.empty?
```

darrenleeweber:
When we're nearly done with this, would be good squash some commits on the `spec/api/sul_bib/publications_api_spec.rb `

darrenleeweber:
Might be good too to squash a couple commits on the Contribution model and, if possible, cull the old validate methods - should be able to call `cont.valid?` after this work is done.  Not sure if anything is trying to validate API params for authorship before doing an update, but, if there is, it will have to handle an AR exception I guess.

atz:
I'd like to defer the API validation migration to a follow-up, but I'm open to some squashing.

app/models/publication.rb
```diff
@@ -250,31 +250,23 @@ def authoritative_pmid_source?
...
-
-    def sync_identifiers_in_pub_hash_to_db
+    # doesn't actually write the Pub to DB, presumed to be part of before_save callback or explicit save
+    def sync_identifiers_in_pub_hash
       incoming_types = Array(pub_hash[:identifier]).map { |id| id[:type] }
       publication_identifiers.each do |id|
         next if id.identifier_type =~ /^legacy_cap_pub_id$/i # Do not delete legacy_cap_pub_id
-        id.delete unless incoming_types.include? id.identifier_type
+        publication_identifiers.delete(id) unless incoming_types.include?(id.identifier_type)
       end
```
darrenleeweber: Not entirely clear about what class we get from `publication_identifiers` (I guess it is an enumerable of `PublicationIdentifier`).  This change seems to remove the AR-record from the enumerable, but not from the db (if I understand this right).  If that's the intention, maybe `publication_identifiers.reject!` declares it up front and simplifies the block body.  Seems like a big change not to delete the db-record though, need to wrap my head around this when I've had some :coffee:

app/models/publication.rb
```diff
@@ -250,31 +250,23 @@ def authoritative_pmid_source?
...
-          publication_identifiers << i unless publication_identifiers.include? i
-        end
+        i = publication_identifiers.find { |x| x.identifier_type == identifier[:type] } # find includes not yet saved pub ids
+        i ||= publication_identifiers.find_or_initialize_by(identifier_type: identifier[:type])
+        i.certainty        = 'confirmed'
+        i.identifier_value = identifier[:id]
+        i.identifier_uri   = identifier[:url]
       end
+      pub_hash[:identifier] = publication_identifiers.map(&:identifier)
     end
```
darrenleeweber: It seems like the `Array` iteration was previously modifying the `publication_identifiers` in the loop, but now it doesn't and nothing directly modifies the loop variable `identifier`.  If I read it correctly, now the iteration is a no-op.  I'm not yet clear on this, but this whole thing looks a bit like the `PublicationIdentifier.pubhash_update` method recently added.  Maybe it could be rewritten like:
```ruby
pub_hash[:identifier]).each do |identifier|
  # find or create the matching PublicationIdentifier record - say `pub_id`
  pub_id.pubhash_update
end
```
It might work better as a class method on `PublicationIdentifier` - something to be responsible for taking an pub_hash[:identifier] hash and doing the find, build or CRUD for it. Dunno.

app/models/publication.rb
atz: No, the problem is that the data doesn't match up between `publications` and `publication_identifiers` 100% of the time.  This method is intentionally relying on the latter table.

atz: I'm open to moving it out, though I don't think it is necessarily part of this PR.

app/models/publication.rb
atz: Yes, you are following the right thread, but missing some context.  It removes it from the association, and then it takes *whatever action* is prescribed by the association itself.  In this case it is declared with 
```ruby
dependent: :destroy
```
Another option, `:nullify`, would keep the record. But this one should destroy it.

spec/models/publication_spec.rb
atz: This is setup and does not require being saved.  `.save` is not the point of this test.

app/models/publication.rb
atz: This line here is moved in consolidating one of the other `sync`-related methods, from here:
https://github.com/sul-dlss/sul_pub/pull/334/files/c41775a9a3e21751a4ae9775c6dc91e8c501ed78..8337b69867e95934b6476f15fcfa175368353a7a#diff-54a25b14a10ce4b168bc13af6acd3190L255

darrenleeweber: - At ln 269, do we need a `publication_identifiers.reload` in there?
- I _think_ I get it; the loop is updating or creating PublicationIdentifier records so they match the pub_hash data; then after the loop all the pub_hash identifiers are replaced with the db-records content.

app/models/publication.rb
darrenleeweber: Split this concern off into https://github.com/sul-dlss/sul_pub/issues/441

app/models/publication.rb
```diff
@@ -250,31 +250,23 @@ def authoritative_pmid_source?
 
   private
 
-    def add_all_identifiers_in_db_to_pub_hash
-      publication_identifiers.reload if persisted?
-      pub_hash[:identifier] = publication_identifiers.map(&:identifier)
-    end
```
darrenleeweber: OK, seems to be entirely internal
```
$ git grep 'add_all_identifiers_in_db_to_pub_hash'
app/models/publication.rb:147:    add_all_identifiers_in_db_to_pub_hash
app/models/publication.rb:274:    def add_all_identifiers_in_db_to_pub_hash
```

app/models/publication.rb
darrenleeweber:
```ruby
publication_identifiers.select {|id| id.identifier_type !~ /^legacy_cap_pub_id$/i }.reject do |id|
  ! incoming_types.include?(id.identifier_type)
end
```

app/models/contribution.rb
darrenleeweber: Created a follow-up issue for this - https://github.com/sul-dlss/sul_pub/issues/442

PR #336, opened 2017-10-28
WebOfScience::Record revised

darrenleeweber:
Connected to #254 

Revisions to `WebOfScience::Record` extracted from #333 to focus on just these changes in this PR.


spec/lib/web_of_science/record_spec.rb
```diff
@@ -1,23 +1,30 @@
...
   let(:wos_uid) { 'WOS:A1972N549400003' }
 
+  let(:medline_encoded_record) { File.read('spec/fixtures/wos_client/medline_encoded_record.html') }
+  let(:medline_decoded_record) do
+    coder = HTMLEntities.new
+    coder.decode(medline_encoded_record)
+  end
+  let(:medline_record_encoded) { described_class.new(encoded_record: medline_encoded_record) }
+  let(:medline_record_decoded) { described_class.new(record: medline_decoded_record) }
+  let(:medline_uid) { 'MEDLINE:24452614' }
```
atz: I don't see where these two statements are used.

darrenleeweber: That's right, they might not be used yet.  They will be useful for additional specs on MEDLINE records, e.g. extracting things for PubHash mapping.  Rather than add them later, they were added while including the MEDLINE fixture data.  Since they are `let` statements, they are no-op until they are used; and I expect they will be used.  If I'm wrong about that, I guess they could be removed soon enough.

darrenleeweber: I will add specs to use these fixture data.

peetucket:
LGTM

darrenleeweber:
The commit history is OK here, it's all fairly atomic changes and I have other things to do than rebase this stuff some more.

Issue #274, opened 2017-10-12
Map new WoS record format(s) to pub_hash


peetucket:
prereq #184

- [ ] confirm that the new WoS records have the minimum required fields to build pub_hash citations; examine a record that exists in both the Sciencewire and Web of Science APIs and compare citations that would be created from both source records and see how they compare

DLW:  A key piece of that puzzle is to determine what the minimal set of fields is to create a citation.  This involves review of
- https://github.com/sul-dlss/sul_pub/blob/master/app/models/pub_hash.rb#L56
- because that is where the pub_hash fields are extracted to create a citation

Maatary:
A first glimpse at the actual first step of the mapping from science wire to a pub_hash (this is not the complete set of information but the bulk)

in the ScienceWireHarvester

```
pub_hash = SciencewireSourceRecord.convert_sw_publication_doc_to_hash(incoming_sw_xml_doc)
```


in SciencewireSourceRecord


```
def self.convert_sw_publication_doc_to_hash(publication)
    doi = extract_doi(publication)
    issn = extract_issn(publication)
    pmid = extract_pmid(publication)
    swid = extract_swid(publication)
    wosid = extract_wosid(publication)

    doi_identifier = { type: 'doi', id: doi, url: "#{Settings.DOI.BASE_URI}#{doi}" } unless doi.blank?
    issn_identifier = { type: 'issn', id: issn, url: Settings.SULPUB_ID.SEARCHWORKS_URI + issn } unless issn.blank?

    identifiers = []
    identifiers << { type: 'PMID', id: pmid, url: "#{Settings.PUBMED.ARTICLE_BASE_URI}#{pmid}" } unless pmid.blank?
    identifiers << { type: 'WoSItemID', id: wosid, url: "#{Settings.SCIENCEWIRE.ARTICLE_BASE_URI}#{wosid}" } unless wosid.blank?
    identifiers << { type: 'PublicationItemID', id: swid } unless swid.blank?
    identifiers << doi_identifier unless doi.blank?

    record_as_hash = {}
    record_as_hash[:provenance] = Settings.sciencewire_source
    record_as_hash[:sw_id] = swid
    record_as_hash[:pmid] = pmid unless pmid.blank?
    record_as_hash[:issn] = issn unless issn.blank?
    record_as_hash[:identifier] = identifiers

    record_as_hash[:title] = publication.xpath('Title').text unless publication.xpath('Title').blank?
[...]

PR #343, opened 2017-10-30
sul_pub-274-Wos-pub_hash-mapping

Maatary:
The mapping from WOS API to the pub_hash structure

Connected to #274


peetucket:
notes in wiki

PR #344, opened 2017-10-30
Avoid n+1 SQL queries

atz:
When you know you are going back to the DB to fetch the associated object for each record, just include that table in the first query and avoid **N+1** requests.

### DIFF
--- a/app/models/publication.rb
+++ b/app/models/publication.rb
@@ -88,11 +88,11 @@ class Publication < ActiveRecord::Base
   end
 
   def self.find_by_doi(doi)
-    PublicationIdentifier.where(identifier_type: 'doi', identifier_value: doi).map(&:publication)
+    PublicationIdentifier.includes(:publication).where(identifier_type: 'doi', identifier_value: doi).map(&:publication)
   end
 
   def self.find_by_pmid_in_pub_id_table(pmid)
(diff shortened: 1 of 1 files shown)


darrenleeweber:
Seems OK, assuming this is supposed to be faster like this.  I'm not very familiar with this, is there an AR doc or recommendation on this?  I'd like to learn something about it.  (I guess I don't want it to result in a large JOIN before returning a result; I assume the SQL engines can optimize it anyway.  I'll take some time tomorrow to look at the SQL in the console.)

atz:
Yeah, this is pretty much a textbooks example of when to use `.includes`:
https://apidock.com/rails/ActiveRecord/QueryMethods/includes

darrenleeweber:
Interesting, looking at the console to see what this translates to in SQL.

### Before
```ruby
doi = "10.0561/100.00005052"
#=> "10.0561/100.00005052"
PublicationIdentifier.where(identifier_type: 'doi', identifier_value: doi).map(&:publication)
```
```sql
SELECT `publication_identifiers`.*
  FROM `publication_identifiers`
  WHERE `publication_identifiers`.`identifier_type` = 'doi' 
    AND `publication_identifiers`.`identifier_value` = '10.0561/100.00005052'
-- followed by another query to get the publication
SELECT  `publications`.* FROM `publications` WHERE `publications`.`id` = 259318 LIMIT 1
```

PR #348, opened 2017-10-31
A simple Harvester base class

atz:
Connected to #254

This is mostly a demonstration.  The main takeaway is starting from an Enumerable of Authors and not having to deal with a bunch of externalizable logic about author selection or batching. Note that `harvest` still operates on an Enumerable of `Author`s, so one given subclass might do:
```ruby
def harvest(authors)
  authors.each { |author| harvest_one(author) }
end
```
while another that hypothetically supports querying, say, 10 authors at once might (re-)batch.  
```ruby
def harvest(authors)
  authors.each_slice(10) { |slice| harvest_ten(slice) }
end
```
Neither has to worry about *any* author selection logic unless it wants to.  Any prevailing pattern that would need to be implemented can be done externally (e.g. single author is trivial), or in the base class thereby covering all subclasses.

If we want to carry this forward, I can add basic tests.  Otherwise just take it for what it is worth as a clean approach.

### DIFF
--- a/spec/lib/science_wire_harvester_spec.rb
+++ b/spec/lib/science_wire_harvester_spec.rb
@@ -1,4 +1,3 @@
-
 describe ScienceWireHarvester, :vcr do
   let(:author_without_seed_data) { create :author, emails_for_harvest: '' }
   let(:author_with_seed_email) { create :author }
--- a/lib/harvester.rb
+++ b/lib/harvester.rb
@@ -1,0 +1,3 @@
+module Harvester
+  require 'harvester/base'
+end
--- a/lib/harvester/base.rb
+++ b/lib/harvester/base.rb
@@ -1,0 +1,24 @@
+module Harvester
+  # An iota of abstraction for things a harvester must do.
+  # The starting point is always one or more Authors.
+  class Base
+    # @return [void]
+    def harvest_all
+      Author.where(active_in_cap: true, cap_import_enabled: true)
+            .find_in_batches(batch_size: batch_size)
+            .each { |batch| harvest(batch) }
+    end
+
+    # @param [Enumerable<Author>] authors
+    # @return [void]
+    def harvest(_authors)
+      raise "harvest must be implemented in subclass"
+    end
+
+    private
+
+      def batch_size
+        100
+      end
+  end
+end
--- a/spec/lib/harvester/base_spec.rb
+++ b/spec/lib/harvester/base_spec.rb
@@ -1,0 +1,20 @@
+describe Harvester::Base do
+  let(:authors) { FactoryGirl.create_list(:author, 5, cap_import_enabled: true) }
+  let(:subclass) { Class.new(described_class) }
+  let(:instance) { subclass.new }
+
+  describe '#harvest_all' do
+    it 'chunks calls to harvest based on batch_size' do
+      expect { authors }.to change { Author.count }.by(5)
+      expect(instance).to receive(:batch_size).and_return(2)
+      expect(instance).to receive(:harvest).exactly(3).times
+      instance.harvest_all
+    end
+  end
+
+  describe '#harvest' do
+    it 'throws exception from base class' do
+      expect { instance.harvest([]) }.to raise_error(RuntimeError)
+    end
+  end
+end


atz:
@darrenleeweber let me know if you want to carry this forward or not.

darrenleeweber:
I like the general idea, just undecided on the public API at the moment.  Can we work out the details on the harvester in #333 first and then see if we can sync up this thing with that thing (changing the public API of either of them to match up one way or the other).  Then we're in a position to work on specs and code coverage.  So, if it's OK, let's keep this open with a view to working on it soon?

atz:
I'm in a position to do that work now.  It is simple.  What I want to know is whether it would be useful or not.

darrenleeweber:
So far the public API on the WOS-harvester looks like:
```ruby
    def process_author(author)
    def process_dois(author, dois)
    def process_pmids(author, pmids)
    def process_uids(author, uids)
```
Is that something we can abstract to a base class?  If the base class supported calling these methods, then the WOS-harvester could inherit it right now.  Then we could get the base class `harvest_all` to be hooked up with `process_author` in the subclass.  Does that make sense?

atz:
Yeah, that is too concrete.  Not every subclass would care about dois/pmids/uids.  You are worrying about the next steps related to the existing class.

darrenleeweber:
OK, then, I'm sold.  Spec it.

atz:
Ha, good question.  Probably should just be undefined.  It's not like there is anything a caller can do with the information (besides log it, which the harvester should already do with greater detail).

darrenleeweber:
On thinking more about return values, prompted changes in the WOS-harvester; it now returns an array of WOS-UID that result in new Publication records.  Could be easily modified to return an array of Pub-id, but that's a little less meaningful for most of the WOS-harvest methods.  When it subclasses the `harvest(authors)` method, however, it could switch it up for the more general case of returning Pub-ids.

atz:
I don't think Code Coverage is helping by complaining about 0% coverage diff.  Especially when it misses that the two methods in the class actually are tested.

spec/lib/science_wire_harvester_spec.rb
```diff
@@ -1,8 +1,7 @@
-
 describe ScienceWireHarvester, :vcr do
   let(:author_without_seed_data) { create :author, emails_for_harvest: '' }
   let(:author_with_seed_email) { create :author }
-  let(:author) { create :author }
+  let(:author) { create :author, cap_import_enabled: false }
```
darrenleeweber: I can understand this change is intended to be a no-op after changing the default in the factory.  Just curious about why this cannot use the default in the factory, if you care to look into it?

darrenleeweber:
- Implicitly call the spec file (should match file pattern "*_spec.rb"):
```
bundle exec rspec spec/lib/harvester/
Run options: exclude {:"data-integration"=>true}

All examples were filtered out

Randomized with seed 921


Finished in 0.00043 seconds (files took 3.95 seconds to load)
0 examples, 0 failures

Randomized with seed 921

Coverage report generated for RSpec to /data/src/dlss/cap/sul_pub/coverage. 296 / 659 LOC (44.92%) covered.
Coverage has dropped by 18.29% since the last time (maximum allowed: 0.10%).
```

- Explicitly call the spec file
```
$ bundle exec rspec spec/lib/harvester/base.rb 
Run options: exclude {:"data-integration"=>true}

Randomized with seed 54457
..

Finished in 0.23465 seconds (files took 3.95 seconds to load)
2 examples, 0 failures

Randomized with seed 54457

Coverage report generated for RSpec to /data/src/dlss/cap/sul_pub/coverage. 330 / 746 LOC (44.24%) covered.
Coverage has dropped by 18.97% since the last time (maximum allowed: 0.10%).
```

spec/lib/science_wire_harvester_spec.rb
atz: Just because the test was specifically asking about `cap_import_enabled`.

darrenleeweber: I suppose the next question is whether or not the factory should have the same default value as the db-schema.  It looks like `db/schema.rb` sets no default and it allows `NULL`; in effect, I _think_ the default is a `falsey` or `nil` value.  Perhaps the factory should not explicitly set it?
```ruby
> Author.new.cap_import_enabled
=> nil
```

darrenleeweber: Seems like the value might be set in `Author.fetch_from_cap_and_create` via the cap authors poller, but it depends entirely on the data in the CAP-API retrieved.

PR #353, opened 2017-10-31
Rails deprecation: Use .reload instead of association(true)

atz:
See:
https://stackoverflow.com/questions/25188266/what-does-activerecord-do-with-arguments-passed-in-to-association-attribute-meth

This would help us in a potential Rails 5 upgrade.


darrenleeweber:
I see.  I want to approve but, but first ask if we can expand the scope to change all the things?

darrenleeweber:
In standup, we noted that this change can't be found anywhere else.

PR #355, opened 2017-11-01
Publication model has methods that should be private

atz:
NOTE: for the purposes of this revision, I am treating `cap_cutover.rake` as a vestigial/deprecated piece of code. Where it is the sole (external) caller for one method, I am not treating that as
a continuing "current" use.  I edited it so the task will still work the same (by using `.send`, similar to the specs), but I am allowing `Publication` to take the otherwise private method private.

In a separate PR, most of `cap_cutover.rake` is being removed.

### DIFF
--- a/app/models/publication.rb
+++ b/app/models/publication.rb
@@ -199,69 +199,6 @@ class Publication < ActiveRecord::Base
     self
   end
 
-  def sync_identifiers_in_pub_hash_to_db
-    incoming_types = Array(pub_hash[:identifier]).map { |id| id[:type] }
-    publication_identifiers.each do |id|
-      next if id.identifier_type =~ /^legacy_cap_pub_id$/i # Do not delete legacy_cap_pub_id
-      id.delete unless incoming_types.include? id.identifier_type
-    end
-
-    Array(pub_hash[:identifier]).each do |identifier|
-      next if identifier[:type] =~ /^SULPubId$/i
-
-      i = publication_identifiers.find { |x| x.identifier_type == identifier[:type] } || PublicationIdentifier.new
-
-      i.assign_attributes certainty: 'confirmed',
-                          identifier_type: identifier[:type],
-                          identifier_value: identifier[:id],
-                          identifier_uri: identifier[:url]
-
-      if i.persisted?
-        i.save
-      else
-        publication_identifiers << i unless publication_identifiers.include? i
-      end
-    end
-  end
-
-  def update_any_new_contribution_info_in_pub_hash_to_db
-    Array(pub_hash[:authorship]).each do |contrib|
-      hash_for_update = {
-        status: contrib[:status],
-        visibility: contrib[:visibility],
-        featured: contrib[:featured]
-      }
-
-      # Find or create an Author of the contribution
-      author_id = contrib[:sul_author_id]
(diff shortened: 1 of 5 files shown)


spec/models/publication_spec.rb
```diff
@@ -161,9 +161,9 @@
 
     it 'should update attributions of existing contributions to the database' do
       expect(publication.contributions.size).to eq(0)
-      publication.contributions.build_or_update(author, status: 'new')
+      publication.contributions.build_or_update(author, status: 'unknown')
```
darrenleeweber: What's happened here, any idea why this needs to change?

spec/lib/doi_search_spec.rb
```diff
@@ -23,7 +23,7 @@
           provenance: 'cap',
           identifier: [{ type: 'doi', id: '10.1111/j.1444-0938.2010.00524.x', url: 'https://dx.doi.org/10.1111/j.1444-0938.2010.00524.x' }]
         }
-        publication.sync_identifiers_in_pub_hash_to_db
+        publication.send(:sync_identifiers_in_pub_hash_to_db)
         publication.save
```
darrenleeweber: Rhetorical Q:  So does this mean it's a public method?  Maybe not, it's only in a spec.  No problem.

app/models/publication.rb
```diff
@@ -373,4 +298,76 @@ def authoritative_pmid_source?
   end
 
   alias authoritative_doi_source? sciencewire_pub?
+
+  private
```
darrenleeweber: Please shuffle methods in the private section alphabetically by method name.

darrenleeweber:
Generally 👍 on this, just requesting a method shuffle and more documentation on the description (so I can clearly see these are private methods and this PR documents that they are private).  That is, please add a git-grep on all these methods to document the description for this PR.  In that documentation, please note the specific exceptions for rake tasks that will be culled later.

spec/lib/doi_search_spec.rb
atz: Nothing should be public just for the purpose of calling it in tests.

spec/models/publication_spec.rb
atz: It doesn't have to, but I think it reflects the intent of the test more clearly.  I.E., a different contribution w/ a different value.

atz:
@darrenleeweber covered everything

spec/models/publication_spec.rb
darrenleeweber: OK

spec/lib/doi_search_spec.rb
darrenleeweber: Yes, agreed.  In this case, it's just a little bit odd that the spec is on a different class.  That's what prompted a bit of extra thought and a comment to myself there, like Ariadne's thread.  Perhaps I can make a pattern out of that -- if I prefix a comment with `Rhetorical Q:` or `Note to self:` then it's just my way of noting that I've looked at it thought about and arrived at a conclusion myself.

darrenleeweber:
I requested proof that these are not public methods on the master branch.  I want to see a git grep on each of these method names from the master branch.

darrenleeweber:
This is reverted, pending proof the methods are not public, including rake tasks.

PR #359, opened 2017-11-01
Remove unused class method on Contribution

atz:
Split off from #334


darrenleeweber:
Yep, I did some checks and agreed to this earlier in #334 [1], glad to see it split out here.

[1] https://github.com/sul-dlss/sul_pub/pull/334#discussion_r147798599

PR #360, opened 2017-11-01
Use valid status values in tests

atz:
At least where it matters (to possibly adding rails validations).  Split off from #334.

### DIFF
--- a/spec/models/publication_spec.rb
+++ b/spec/models/publication_spec.rb
@@ -1,4 +1,3 @@
-
 describe Publication do
   let(:publication) { FactoryGirl.create :publication }
   let(:author) { FactoryGirl.create :author }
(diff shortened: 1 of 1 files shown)


darrenleeweber:
Could these specs be improved with a default-global let for authorship data, or by changing or using a FactoryGirl.create for it?  In effect, something like:
```ruby
let(:authorship) { [{ status: 'new', sul_author_id: author.id }] }
```
If some specs need a different status, change the default let return data, eg 
- `authorship[:status] = 'unknown'`

If there is a common global default with cap_author instead, create that let too.

atz:
I don't see that as required for this PR, right?  I mean, that question is separate from whether the values are valid or not.

darrenleeweber:
Just depends on how much you want to clean up; e.g. there are some factories for this stuff
- spec/factories/contribution.rb
- spec/factories/publication.rb

If you don't want to clean up to use factories and just want a few valid values, this gets that done I suppose.  If it's only like an hour to use factories instead, that would be great.  If it's like 4 hours, let's skip it.  If you got bigger fish to fry, let's skip it.  Ping me in slack if you prefer a real-time chat.

PR #365, opened 2017-11-02
[WIP, TBD] WebOfScience::Record#identifiers_merge to integrate links data

darrenleeweber:
Extracted from #333 for a focused review, required in #333 

This is going to be replaced by #370

### DIFF
--- a/lib/web_of_science/record.rb
+++ b/lib/web_of_science/record.rb
@@ -51,6 +51,13 @@ module WebOfScience
       end
     end
 
+    # Update identifiers using a merge strategy to preserve the values already in the identifiers
+    # @param links [Hash<String => String>] other identifiers (from Links API)
+    # @return identifiers [Hash<String => String>]
+    def identifiers_merge(links)
+      identifiers.reverse_update links
+    end
+
     # @return names [Array<Hash<String => String>>]
     def names
       @names ||= begin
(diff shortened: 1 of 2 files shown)


atz:
This seems unnecessary.  Are you testing `identifiers.update`?  Because otherwise you are only testing "does Ruby work?"  Really dislike seeing structure put in place of fluency.  

I provided code that would avoid the unnecessary overwriting operations, but you might prefer to use `reverse_merge`, which is [built for exactly this case](https://apidock.com/rails/Hash/reverse_merge).  I think that makes it simple enough so as not to warrant an individual method.

darrenleeweber:
You could have pointed out this rails-y helper earlier 😏 

Looks like it should be `reverse_update`, i.e.
```ruby
 bundle exec rails c
Loading development environment (Rails 4.2.10)
[1] pry(main)> x = {a: 1}
=> {:a=>1}
[2] pry(main)> y = {a: 2, b: 2}
=> {:a=>2, :b=>2}
[3] pry(main)> x.reverse_merge y
=> {:a=>1, :b=>2}
[4] pry(main)> x
=> {:a=>1}
[5] pry(main)> x.reverse_update y
=> {:a=>1, :b=>2}
[6] pry(main)> x
=> {:a=>1, :b=>2}
```

darrenleeweber:
I'm going to close this, but not delete the branch right away.  It should be replace by #370 instead.

PR #370, opened 2017-11-02
WebOfScience::Identifiers

darrenleeweber:
Required for #333 

Connected to #254 to code the business logic for the WOS harvester
Connected to #301 to code logic for de-duping new WOS-records to create new Pubs

This replaces #365 (although #365 was OK, it left the identifiers open to exploitation/corruption, alternative truths and all that is evil, Darth).

This class is entirely responsible for WOS identifiers and managing any inconsistencies there might be between how WOS names their identifiers and how sul_pub names them.  It's purpose is to be the sole gatekeeper between WOS and sul_pub.  In that role, it is tasked with protecting the identifiers as immutable data, but allowing very specific merges from the links-API data.  It's also responsible for parsing identifiers based on the type of WOS-DB providing the record and transforming some identifiers into useful forms that match sul_pub identifiers.

### DIFF
--- a/lib/web_of_science/identifiers.rb
+++ b/lib/web_of_science/identifiers.rb
@@ -1,0 +1,104 @@
+require 'forwardable'
+
+module WebOfScience
+
+  # Immutable Web of Knowledge (WOK) identifiers
+  class Identifiers
+    extend Forwardable
+    include Enumerable
+
+    # Delegate enumerable methods to the mutable Hash.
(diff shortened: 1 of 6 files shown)


darrenleeweber:
I've requested review on this and nobody is accepting the work.  All the specs are good and code climate is good, even fixed and issue.  This is blocking more important work now.  I'm going to assume responsibility for moving this forward.

PR #374, opened 2017-11-03
[Dup 376?]  Cutover housecleaning

atz:
Shrinks `cap_cutover.rake` by 14 of 16 tasks.  Then follows up by:
- removing previously used code that is now unused
- making `private` any methods that were only publicly used in the rake file (including class methods)

Tests of newly private methods were revised to use `.send` if necessary.  Besides the cut to deadweight, the benefit is mainly to `Publication` and the source record models.


darrenleeweber:
Conflicts

darrenleeweber:
This is consolidated into #376

Issue #368, opened 2017-11-02
WebOfScience::Record#to_pub_hash


darrenleeweber:
From #350 

- [ ] implement `WebOfScience::Record#to_pub_hash` to work with common summary data

```ruby
def to_pub_hash
  {
    # summary data mapped to pub_hash
  }
end
```

Use https://github.com/sul-dlss/sul_pub/wiki/WOS-to-Pub_hash-Mapping

darrenleeweber:
Connected to https://github.com/sul-dlss/sul_pub/issues/383

darrenleeweber:
@Maatary - please pull the content from the google doc into the wiki.  I have to switch google accounts to be able to view it (and still not sure I can see it); this minor inconvenience is a total time sink and interrupts my dev-flow big time.

darrenleeweber:
[Wos-Pub_hash-Mapping.pdf](https://github.com/sul-dlss/sul_pub/files/1442811/Wos-Pub_hash-Mapping.pdf)

darrenleeweber:
@Maatary - did you find a WOS-UID for a conference record?  What is the UID?

darrenleeweber:
RE questions related to Links-API fields, this is the list of all fields
```ruby
ALL_FIELDS = %w(ut doi pmid title isbn issn issue vol year tpages sourceURL timesCited citingArticlesURL relatedRecordsURL).freeze
```
The rake task `wos:links["UID"]` can be used to search data, when we merge this rake task from #369

PR #375, opened 2017-11-03
[Obsolete -> 379, 380] WebOfScience::Record#to_pub_hash

Maatary:
#368
-Refactored Record toward the SRP pattern by moving out all the logic of WOS XML Document information extraction, in its own class named Wos_Extractor.

-Record now simply act as a the pure data type that it is, that is a Record. It provide methods to access the Record relevant information and the toPubHash Method to get a PubHash from it. Ideally that function should be moved in the consumer of the Record. The record can't know all its consumer in advance and provide a method for each. But a consumer as the PubHash class knows what it can be initialized from. Hence a class method in PubHash such that PubHash.fromRecord(record) that would instantiate the PubHash would be preferable. 

-As of now we extract 90% of what is required. In comment are the contentious element that are missing. The other fiels missing are the 2 PubMed fields that can be extracted when the Record is a Medline Record. This can be determine by the Database field.

Mapping can be found here https://github.com/sul-dlss/sul_pub/wiki/WOS-to-Pub_hash-Mapping

### DIFF
--- a/spec/lib/web_of_science/record_spec.rb
+++ b/spec/lib/web_of_science/record_spec.rb
@@ -224,56 +224,6 @@ describe WebOfScience::Record do
     end
   end
 
-=begin
-  describe '#summary' do
-    let(:summary) { wos_record_encoded.summary }
-
-    it 'works' do
-      expect(summary).to be_an Hash
-    end
-    it 'contains doctypes Array' do
-      expect(summary['doctypes']).to be_an Array
-    end
-    it 'contains names Array' do
-      expect(summary['names']).to be_an Array
-    end
-    it 'contains pub_info Hash' do
-      expect(summary['pub_info']).to be_an Hash
-    end
-    it 'contains publishers Array' do
-      expect(summary['publishers']).to be_an Array
-    end
-    it 'contains titles Hash' do
-      expect(summary['titles']).to be_an Hash
-    end
-  end
-=end
-=begin
-
-  describe '#summary_struct' do
-    let(:summary) { wos_record_encoded.summary_struct }
-
-    it 'works' do
-      expect(summary).to be_an OpenStruct
-    end
-    it 'contains doctypes Array' do
-      expect(summary.doctypes).to be_an Array
-    end
-    it 'contains names Array' do
-      expect(summary.names).to be_an Array
-    end
-    it 'contains pub_info OpenStruct' do
-      expect(summary.pub_info).to be_an OpenStruct
-    end
-    it 'contains publishers Array' do
-      expect(summary.publishers).to be_an Array
-    end
-    it 'contains titles OpenStruct' do
-      expect(summary.titles).to be_an OpenStruct
-    end
-  end
-=end
-
   describe '#titles' do
     it 'works' do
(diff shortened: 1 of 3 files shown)


darrenleeweber:
I want to see a PR that changes nothing, only adds the `toPubHash` method, which should be `to_pub_hash`, as noted explicitly in https://github.com/sul-dlss/sul_pub/issues/368

This PR exceeds the requirements explicitly noted in 368.  Please either close this PR and keep the branch so a new branch/PR can focus tightly on 368.  If necessary, we can revisit the purpose of extracting the extractors (but I don't see it right away).  

Please also take a look at #370 to see what will be happening there and we will reconsider what to do with the identifier URL data after #370 and this work is complete.  Or this work might need to wait for #370 and then modify the Identifiers class to add the URLs there.

spec/lib/web_of_science/record_spec.rb
```diff
@@ -224,6 +224,7 @@
     end
   end
 
+=begin
```
atz: I don't see any useful changes to this file.  What are you intending w/ these `=begin` and `=end` tags?

Maatary: That was just putting things in comment as per the IDEA

Maatary:
@darrenleeweber Yes it is a shell and that's is intended. It is just a structure that represent your data and you can take from it whatever you want. IT is your Record as a Ruby object full stop. You can take the field that you want from the Record, or get the Record as a PubHash. IMHO, that is the only thing that should be done with an abstraction that represent a record. This is a separation of concern and a single responsibility principle. Its sole responsibility is to provide data as is or as a PubHash. 

The responsibility of extracting data is something else. 

I want to remind the SRP = Single Responsibility Principle. A class should have only one reason to change. If the Structure of the XML Changes or if we want to get things differently, we won't have to touch the Record Class. We will only focus on the extractor. 

There is no need to jump between two classes if you consider the SRP. One class abstract a record and only does that. A record is only something you get information from. The other class abstract an Extractor. 

PS: 

This is why Functional programing is so popular because you would think like that by default. The reason being that FR push the combination of OO best practices( e.g. SRP) to their logical extinct by putting them in the language directly. Class like Record are called ADT.

darrenleeweber:
I want to defer the work on the extractor, if we need it.  I'm not entirely opposed it, just saying it's out of scope for the issue to be addressed by this work.
Whenever the record data changes, it's most likely all the `attr_reader` instance vars will also change on Record, along with the extractor.  Note that in #370 I have split out some of the responsibility to another class and delegated methods to that class.  Using method delegation might be cleaner than all the assignments to instance vars.  But anyway, that is out of scope.

Maatary:
> Whenever the record data changes, it's most likely all the attr_reader instance vars will also change on Record, along with the extractor.

That is not correct in my humble opinion. The data can be situated elsewhere in different section of the XML, or simply named differently. The bit of information is independent of the form in which the information comes. 

But anyway i get you

darrenleeweber:
In slack discussion, we agreed for me to take this on, as @Maatary is assigned to other LD4P work, as of today and next week.

PR #377, opened 2017-11-03
Avoid reloading twice

atz:
The first thing `add_all_identifiers_in_db_to_pub_hash` does is:
```
publication_identifiers.reload if persisted?
```

Since this script only runs on persisted objects, that means it will do the same reload every time anyway.


atz:
Uh, I think we can avoid a -0.2% coverage change blocking this.  Deleting one line should never be blocked by coverage.

PR #379, opened 2017-11-03
WebOfScience::Record#pub_hash

### LINKED ISSUE #368: WebOfScience::Record#to_pub_hash

### LINKED ISSUE #383: Construct PubHash "pages" values to match current Sciencewire format
If the WOS record has pages information for `start` and `end` page, with the same page, it could map into the following in a PubHash pages field:
```ruby
:pages=>"413-413"
```
Is this the desired result or should it be simplified to one page, i.e.
```ruby
:pages=>"413"
```

In the google doc of the mappings at https://docs.google.com/document/d/1x5sZVpZ_nREWBizyiTCVuP6YmJ6-AhA2yfCxrWZi-FU/edit @Maatary notes that the prior format for SW-API records was like
```
:pages=>"413-6"
```
So, that's even more complex, esp. when the pages span boundaries across the 10-units, like `418-420` would be `418-20` in that form.

darrenleeweber:
Fix #383 
Fix #368 

Isolating components of #375 
- this isolates only the `toPubHash` method from d86ae6f653ae189e11ed294c3b63468afc1ac926

TODO:
- [x] cherry-pick additional commits from #375 
- [x] squash commits into their final form
- [x] more specs

@Maatary 

An informal integration test was used to develop and revise this by checking how it handles some WOS records, e.g. using the `bundle exec rails c` console, the records should all parse without exceptions:
```ruby
altman_records = WOS.queries.search_by_name('Altman, Russ, B', ['Stanford University']);
altman_records.map(&:pub_hash)
# the next step is to create citations, e.g.
altman_records.map {|rec| ph = PubHash.new(rec.pub_hash); ph.to_apa_citation }
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -46,6 +46,7 @@ gem 'pry-rails'
 gem 'pubmed_search'
 gem 'savon', '~> 2.11'
 gem 'simple_form'
+gem 'StreetAddress', '~> 1.0', '>= 1.0.6'
 gem 'turnout'
 gem 'whenever', require: false
 gem 'yaml_db'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,6 +1,7 @@
 GEM
   remote: https://rubygems.org/
   specs:
+    StreetAddress (1.0.6)
     actionmailer (4.2.10)
       actionpack (= 4.2.10)
       actionview (= 4.2.10)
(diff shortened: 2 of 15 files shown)


darrenleeweber:
I've reviewed all the other commits in #375 and decided they add no more data fields; mostly manual rubocop fixes (could have used `rubocop -a` to fix them) and removing some specs that need to stay in this PR.

PR #381, opened 2017-11-03
lib/scripts/publication_identifier_normalization.rb

darrenleeweber:
@atz argued for this, but there's no sprint goals for it and I'm not adding specs for it.  If @atz insists on specs etc., please merge and go for it (if the team agrees to ticket and assign sprint work for it).


atz:
Understood that there is tech debt here around actually testing the new class.  That was inherent to the previous PR already merged.

PR #382, opened 2017-11-04
WOS provenance setting

darrenleeweber:
Required in #379 to populate the PubHash with a new source record provenance. 
- f291f9341a1da71fea30dcfbd8f7de946b316291

I assume we want to replace
```ruby
pub_hash[:provenance] = Settings.sciencewire_source
```
with say
```ruby
pub_hash[:provenance] = Settings.wos_source
```

What provenance do we want to use for Web of Science records?  This PR is a suggestion, I'm open to alternatives.  We seem to be using `wos` in our common vocab on this work.

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -85,15 +85,12 @@ sul_doc_types:
   book: book
   article: article
 
-cap_provenance: cap
-
-manual_source: manual
-
-sciencewire_source: sciencewire
-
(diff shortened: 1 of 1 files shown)


Maatary:
I think adding this ```Settings.wos_source``` or ```Settings.webOfscience_source```

PR #407, opened 2017-11-07
Adjust scheduling of harvesting rake tasks on qa and prod so that will not run concurrently

peetucket:
currently prod and cap-qa run their fortnightly harvests on the exact same day, just staggering the hour and minute start time -- this fixes the start hour and time but staggers them 5 days apart so they will not run concurrently

### DIFF
--- a/config/schedule.rb
+++ b/config/schedule.rb
@@ -6,12 +6,17 @@ end
 
 set :output, 'log/cron.log'
 
-# fortnightly sciencewire harvest at 5pm-ish
-every 2.weeks, at: stagger(17), roles: [:harvester] do
+# fortnightly sciencewire harvest at 5pm in qa, on the 5th and 20th of the month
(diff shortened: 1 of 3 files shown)


config/schedule.rb
```diff
@@ -1,17 +1,18 @@
...
+every "0 17 1,15 * *", roles: [:harvester_prod] do
   rake 'sw:fortnightly_harvest'
 end
 
-# poll cap for new authorship information nightly at 4am-ish
-every 1.day, at: stagger(4), roles: [:harvester] do
+# poll cap for new authorship information nightly at 4am-ish in both prod and qa
+every 1.day, at: stagger(4), roles: [:harvester_qa, :harvester_prod] do
   rake 'cap:poll[1]'
 end
```
darrenleeweber: Looks like the `stagger` method is still required here on the daily cron.

peetucket: indeed...added back and rebased

PR #376, opened 2017-11-03
Cutover pruning

atz:
Fix #356 

This rolls up the work from two separate branches (incl. #374) (thereby avoiding a merge conflict).

Issue #356, opened 2017-11-01
Review cap_cutover.rake for possible deletion


atz:
There is a fairly significant surface of code that exists only to service this seemingly obsolete rake file `cap_cutover.rake`, some of which already described itself as "very temporary"... several years ago.  Ideally, we can rid ourselves of it completely.  

Barring that, if there are individual tasks that are still presently useful, please identify them so we can: 
- know what is worth maintaining, and
- get rid of the rest.

peetucket:
Recommend we keep this file, but add a comment to the top:
```# As of November 2017 these tasks have not knowingly been run since at least March 2016.  It is possible they can be removed completely at a future refactor.```

*Some specific task note*

Tasks that seem to be related to some one time cutover, unlikely to ever be called again, can be deprecated:
* `pull_pubmed_for_cap` 
* `pull_sw_for_cap`
* `ingest_authors`
* `create_pubs_from_cap`
* `create_pubs_from_delta`
* `ingest_man_pubs`
* `man_pubs_from_delta`
* `fix_man_pubs_authors`
* `overwrite_profile_ids`
*  `build_pubs`, `build_missing_pubs`, `build_delta_pubs`, `update_pubs`, `update_authors`, `build_missing_pubs_one_item` --- these ultimately call a method `cutover_sync_hash_and_db` on the `Publication` model, which I don't think we need anymore either (it has a comment suggesting it was a temporary method)

No longer needed, this functionality now exists in the `update_pubs` task in the `sul.rake` file:
* `rewrite_authorship`

**Comment by atz** [comment]
Deprecation is a formal communication to consumers of code about changing pattern.  But we are the consumers of rake tasks.  What are we trying to communicate?  I don't mind adding some comments, but it seems weird that we are treating the users' behaviors as unknowable when they are us.  Either we use the things those (untested, unmaintained) tasks provide or we don't.  

If we don't need `cutover_sync_hash_and_db`, then we also don't need those 6 rake tasks.

More concretely, when I look at these tasks, they spawn logs.  Those logs have 9 names:
```
cap_build_delta_pubs.log
cap_build_pubs.log
cap_create_pmid_pubs.log
cap_import_man_pubs.log
cap_import_pubmed_source_records.log
cap_import_sciencewire_source_records.log
cap_update_pubs.log
fix_man_pubs_authors.log
rebuild_authorship_in_hash.log
```
Our prod logs directory has files going back to `2016 Jun 1`.  NONE of those files appear.  

[...]

PR #422, opened 2017-11-13
CapProfileIdRewriter now unused, removed

atz:
Split off from #376.  

Connected to #356


darrenleeweber:
As noted in #376, this script has been used when migrating db-dumps between deployment systems.  The purpose is to sync the `cap_profile_id` values between the CAP-API and the SUL-PUB author dbs.  The sync is required anytime the CAP-team migrate data between deployment systems (e.g., prod -> dev) or we do.  Unless there are arguments that this is not needed, I cast a :-1: on removing this script.

darrenleeweber:
Prior to #376 we had:
```
$ git grep 'CapProfileIdRewriter'
lib/cap_profile_id_rewriter.rb:3:class CapProfileIdRewriter
lib/tasks/cap_cutover.rake:391:    CapProfileIdRewriter.new.rewrite_cap_profile_ids_from_feed
```
After #376 the rake task is removed
```
$ git grep 'CapProfileIdRewriter'
lib/cap_profile_id_rewriter.rb:3:class CapProfileIdRewriter
```

darrenleeweber:
@peetucket - cast a tie-breaking vote to keep this file.

PR #426, opened 2017-11-14
Replace rake task to sync cap_profile_ids, removed in #376, connected…

darrenleeweber:
Replacing a rake task removed in #376
- slightly modified to protect prod environment

Connected to #356, #422

### DIFF
--- a/lib/tasks/cap_cutover.rake
+++ b/lib/tasks/cap_cutover.rake
@@ -59,4 +59,10 @@ namespace :cap_cutover do
     end
     Rails.logger.info total_running_count.to_s + ' in ' + distance_of_time_in_words_to_now(start_time)
   end
+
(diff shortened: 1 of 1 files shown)


atz:
Alternatively, that can just be done from console or: 
```bash
rails runner -e production 'CapProfileIdRewriter.new.rewrite_cap_profile_ids_from_feed'
```

There really isn't anything to the task.

darrenleeweber:
Yes, it could be run like that too.  With the task, it shows up in `rake -T` (which is where we go when we forget what we want to run).  It tries to protect a production system by excluding running it in `production`.  Of course, there are ways around that and a savy-dev should be able to find a way around it, if necessary.  The production "protection" is just for a user who might make a mistake.

atz:
I thought you described this as being run on stage.  Am I recalling correctly?  If so, that is weird, because stage has `RAILS_ENV=production`.

darrenleeweber:
Allow it to run on a production system, but only after user confirmation.

darrenleeweber:
The team voted to keep the file in #422 and this rake task is useful.

PR #427, opened 2017-11-14
Publication/PublicationIdentifier specs use factory data

### LINKED ISSUE #424: Specs on Publication/PublicationIdentifier can use factories
These specs should not call a Publication model private callback method, they can use a factory, i.e.
- https://github.com/sul-dlss/sul_pub/blob/master/spec/factories/publication_identifier.rb

```
spec/api/sul_bib/sourcelookup_spec.rb:69:        publication.sync_identifiers_in_pub_hash_to_db
spec/lib/doi_search_spec.rb:26:        publication.sync_identifiers_in_pub_hash_to_db
```

darrenleeweber:
Fix #424 

- [x] spec/api/sul_bib/sourcelookup_spec.rb
- [x] spec/lib/doi_search_spec.rb

After #376 these specs had to call a private callback method to create publication identifiers.  This is fixed in this PR by using factory data for the publication identifiers and calling the `Publication.save` that should trigger it's private callback methods.

### DIFF
--- a/spec/api/sul_bib/sourcelookup_spec.rb
+++ b/spec/api/sul_bib/sourcelookup_spec.rb
@@ -15,13 +15,7 @@ describe SulBib::API, :vcr do
     expect(response.status).to eq(200)
     JSON.parse(response.body)
   end
-  let(:test_doi) { '10.1016/j.mcn.2012.03.008' }
-  let(:sourcelookup_by_doi) do
-    params = { format: 'json', doi: test_doi }
-    get sourcelookup_path, params, capkey
-    expect(response.status).to eq(200)
-    JSON.parse(response.body)
-  end
+
   let(:valid_json_for_post) do
     {
       title: 'some title',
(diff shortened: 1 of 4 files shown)


spec/lib/doi_search_spec.rb
```diff
@@ -1,16 +1,21 @@
 
 describe DoiSearch do
-  let(:publication) { create :publication }
-  let(:author) { create :author }
+  let(:doi_value) { '10.1016/j.mcn.2012.03.007' }
+  let(:doi_identifier) do
```
atz: This is the same as the above `:doi_identifier`, so you can back out one to the common context and keep it in common.  

It might seem like there is an ordering issue (because `doi_identifier` depends on `doi_value`), but `let`  statements are not executed in the order declared (rather, in the order invoked in the example), so it would work fine.

spec/api/sul_bib/sourcelookup_spec.rb
```diff
@@ -50,24 +44,27 @@
...
+      end
+      let(:result) do
+        params = { format: 'json', doi: doi_value }
+        get sourcelookup_path, params, capkey
+        expect(response.status).to eq(200)
+        JSON.parse(response.body)
+      end
+
       it 'returns one document ' do
-        result = sourcelookup_by_doi
```
atz: Definitely good to get rid of this useless aliasing!

atz:
Generally positive, with just that one suggestion.

spec/lib/doi_search_spec.rb
darrenleeweber: Huh?  It's in the global context for this spec file already.  Maybe your referring to the spec above, but these are different spec files and the `doi_value` is different between the two files (the last digit is different).

atz: OK, that makes sense.

PR #429, opened 2017-11-14
Wos identifiers enhanced

### LINKED ISSUE #410: When a 'doi' and 'xref_doi' are both present in the identifiers results of a record and are the same, only store the 'doi'
Records sometimes have both a doi and a xref_doi identifier in the record with the same value.  In these cases, only store the doi in publication identifiers.

### LINKED ISSUE #430: Collect the WOS EISSN when it's available

darrenleeweber:
Fix #410 
Fix #430 

Extracted from #379 for easier review.

### DIFF
--- a/lib/web_of_science/identifiers.rb
+++ b/lib/web_of_science/identifiers.rb
@@ -17,7 +17,9 @@ module WebOfScience
     # @param rec [WebOfScience::Record]
     def initialize(rec)
       raise(ArgumentError, 'ids must be a WebOfScience::Record') unless rec.is_a? WebOfScience::Record
-      extract_ids(rec)
+      @rec = rec
+      extract_ids
+      extract_uid
       parse_medline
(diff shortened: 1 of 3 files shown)


lib/web_of_science/identifiers.rb
```diff
@@ -38,10 +40,30 @@ def doi
     end
 
     # @return [String|nil]
+    def doi_uri
+      "#{Settings.DOI.BASE_URI}#{doi}" if doi.present?
+    end
+
+    # @return [String|nil]
```
atz: YARD uses commas for types, not pipes, like `[String, nil]`:
http://www.rubydoc.info/gems/yard/file/docs/GettingStarted.md#Declaring_Types

darrenleeweber: Fixed

PR #431, opened 2017-11-14
Rubocop rake files

atz:


lib/tasks/pubmed.rake
```diff
@@ -44,79 +49,73 @@ namespace :pubmed do
...
-        message = "*****cap_profile_id #{id} not found"
-        puts message
-        logger.error message
-      end
-      if error_count > max_errors
-        message = "Halting: Maximum number of errors #{max_errors} reached"
-        logger.error message
-        raise message
       end
+      next unless error_count > max_errors
```
darrenleeweber: I don't really care about this, but I find it easier to read: `next if error_count < max_errors`

lib/tasks/pubmed.rake
```diff
@@ -44,79 +49,73 @@ namespace :pubmed do
...
+    end_time = Time.zone.now
+    message = "Total: #{total_authors}. Authors found: #{authors_found}. Authors not found: #{authors_not_found}.  Total publications: #{total_pubs}.  Publications updated: #{pubs_updated_count}.  Errored publications: #{error_count}.  Ended at #{end_time}.  Total time: #{distance_of_time_in_words(end_time, start_time)}"
     puts message
     logger.info message
   end
 
   desc 'Retrieve and print a single publication by PubMed-ID'
   task :publication, [:pmid] => :environment do |_t, args|
-    fail "pmid argument is required." unless args[:pmid].present?
+    raise "pmid argument is required." unless args[:pmid].present?
```
darrenleeweber: Prefer `if {x}.blank?` to `unless {x}.present?`

lib/tasks/pubmed.rake
atz: This isn't my code and wasn't flagged by rubocop.

lib/tasks/pubmed.rake
atz: This is a rubocop-applied revision, not something I am authoring now.  Either we can accept rubocop revisions or we can't.  There is no need to intermingle every possible desired improvement with easily approved automated and auto-prompted cleanup.

PR #433, opened 2017-11-16
WebOfScience::Record - use the wos_logger

darrenleeweber:
Extracted from #379 

This will enable logging exceptions when processing WOS record data.


atz:
I still think the `NotificationManager` pattern is backwards, but this is small enough I don't care much.

PR #434, opened 2017-11-16
WebOfScience::Records - #sample returns random record

darrenleeweber:
Extracted from #379 

This enables random sampling of records from WOS-records (the `:to_a` comes from `Enumerable` and `sample` is from `Array`; there's no need to test these ruby functions, it works nicely in the console).


atz:
When do we ever want a random record?

darrenleeweber:
I use it in the console all the time and I want the convenience of an easy accessor.

atz:
But there is no production code or console use case that actually uses this.  That makes it seem odd to include as production code.

PR #435, opened 2017-11-16
Wos xml parser

darrenleeweber:
Extracted from #379 

This greatly simplifies the WOS-record(s) parsing and extraction for XML data.  It decreates the memory footprint for those classes too, by hanging onto only the data it needs.

### DIFF
--- a/spec/lib/web_of_science/xml_parser_spec.rb
+++ b/spec/lib/web_of_science/xml_parser_spec.rb
@@ -5,8 +5,6 @@ describe WebOfScience::XmlParser do
   let(:wos_record_encoded) { WebOfScience::Record.new(encoded_record: wos_encoded_record) }
 
   shared_examples 'it has well formed XML' do
-    let(:html_char) { '&lt;' }
-
     it 'returns an XML String' do
       expect(xml_result).to be_an String
     end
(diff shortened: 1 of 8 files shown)


spec/lib/web_of_science/xml_parser_spec.rb
```diff
@@ -0,0 +1,34 @@
+require 'htmlentities'
+
+describe WebOfScience::XmlParser do
+  let(:wos_encoded_record) { File.read('spec/fixtures/wos_client/wos_encoded_record.html') }
+  let(:wos_record_encoded) { WebOfScience::Record.new(encoded_record: wos_encoded_record) }
+
+  shared_examples 'it has well formed XML' do
+    let(:html_char) { '&lt;' }
```
atz: This is only used once.  It doesn't seem to need `let` semantics.

darrenleeweber: OK, removed

spec/lib/web_of_science/xml_parser_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+    end
+  end
+
+  describe '#to_xml' do
+    let(:xml_result) { wos_record_encoded.to_xml }
+
+    it_behaves_like 'it has well formed XML'
+  end
+
+  describe '#parse_xml' do
```
atz: The method being tested here is `#parse`

darrenleeweber: Done

PR #438, opened 2017-11-16
bundle update dlss_cops rubocop parallel

darrenleeweber:
Fix "The rubocop dependency defined in Gemfile.lock has a known low severity security vulnerability in version range < 0.48.1 and should be updated."

This PR updates to:
```
$ bundle show | grep rubocop
  * rubocop (0.49.1)
  * rubocop-rspec (1.15.1)
$ bundle show | grep cops
  * dlss_cops (0.0.5)
```

The rubocop auto-config generator updated the TODO items (fixing them is not the goal of this PR):
```
187 files inspected, 2231 offenses detected
Created .rubocop_todo.yml.
```
Most, if not all, of the layout offences could be auto-corrected.  All the layout offences were first removed from the TODO file and then auto-correction was applied in one commit.  The auto-correct report was
```
187 files inspected, 108 offenses detected, 107 offenses corrected
```

Connected to an update on dlss_cops, see https://github.com/sul-dlss/dlss_cops/pull/13


atz:
Ugh, new rubocop means `172 files inspected, 1793 offenses detected`.  Some stuff seems helpful, but others I find unpalatable, like:
```
use (@author_count % 50).zero? instead of @author_count % 50 == 0
```
Some of it just just churn, based on rubocop-rails or rubocop-rspec or rubocop itself deciding that `%w[]` is better than `%w()` that it used to recommend.

Not really interesting in vetting ~1800 changes presently.

atz:
Some of these are just wrong in our context:
```ruby
lib/doi_search.rb:3:15: C: Rails/DynamicFindBy: Use find_by instead of dynamic find_by_doi.
    results = Publication.find_by_doi doi
```
`find_by_doi` is a method we define, not a dynamic rails method.  Now, that **is** its own problem, but it is certainly not going to be fixed by changing that line to `Publication.find_by(doi: doi)`.

We have a lot of one-off `find_by` methods that trigger similar warnings.

darrenleeweber:
Superceded by #564

Issue #404, opened 2017-11-07
Fix "method_lines" issue in lib/pub_hash.rb


dazza-codes:
Method `csl_doc` has 81 lines of code (exceeds 25 allowed). Consider refactoring.

https://codeclimate.com/github/sul-dlss/sul_pub/lib/pub_hash.rb#issue_5a0156e47642320001000157

Issue #402, opened 2017-11-07
Fix "method_complexity" issue in lib/pub_hash.rb


dazza-codes:
Method `create_csl_report` has a Cognitive Complexity of 16 (exceeds 5 allowed). Consider refactoring.

https://codeclimate.com/github/sul-dlss/sul_pub/lib/pub_hash.rb#issue_5a0156e47642320001000155

PR #440, opened 2017-11-17
PubHash to Csl::Citation

darrenleeweber:
This PR moves `PubHash` to `Csl::Citation`.

### Rationale

The `PubHash` class name is misleading;  the current `PubHash` class is not really responsible for any `pub_hash` data at all.  It is really responsible for mapping `pub_hash` data into citations, using the citation style language (CSL).  This PR renames and moves this class in way that clearly communicates this purpose.

This also paves the way for creating (or replacing) a new `PubHash` class that is actually responsible for the `pub_hash` data.  Ideally, it would be _the_ wrapper of that data for all purposes (and possibly _the_ MySQL “type” for that field in the publication table).  This PR does not yet create that class, it opens the namespace for a new `PubHash` class, if we want to create it.  In essence, this new class would initially be an extract class refactor on the `Publication` model (e.g. extract all the pub_hash accessors and possibly a lot of the "sync" functionality).  At the extreme, a new `PubHash` class could be responsible for dynamic serialization of the app API responses and, if that were done, the `Publication.pub_hash` field could be dropped entirely, along with all the complex sync callback operations.

### Follow up issues

The code-climate issues are old tech-debt that can be addressed in backlog issues #402, #403, and #404.  This PR will make this work easier, because the refactoring may involve extracting a class hierarchy to work with different doc-types.  These classes can live in the CSL module.

### Before/After greps

- Before
```
$ git checkout master 
Switched to branch 'master'
Your branch is up-to-date with 'origin/master'.
$ git grep 'PubHash.new'
app/models/publication.rb:229:    h = PubHash.new(pub_hash)
lib/pubmed_harvester.rb:41:      h = PubHash.new(pub_hash)
lib/science_wire_client.rb:295:        h = PubHash.new(pub_hash)
```

- After
```
$ git checkout pubhash-to-citation 
Switched to branch 'pubhash-to-citation'
Your branch is up-to-date with 'origin/pubhash-to-citation'.
$ git grep 'PubHash.new'
# nothing
$ git grep 'Csl::Citation.new'
app/models/publication.rb:229:    cite = Csl::Citation.new(pub_hash)
lib/pubmed_harvester.rb:41:      cite = Csl::Citation.new(pub_hash)
lib/science_wire_client.rb:294:        cite = Csl::Citation.new(pub_hash)
spec/models/publication_spec.rb:321:      cite = Csl::Citation.new({})
```


spec/models/publication_spec.rb
```diff
@@ -317,18 +317,25 @@
   end
 
   describe 'update_formatted_citations' do
-    it 'should update the apa, mla, and chicago citations' do
-      allow(publication).to receive(:pub_hash).and_return({})
-      apa = double
-      mla = double
-      chicago = double
-      expect_any_instance_of(PubHash).to receive(:to_apa_citation).and_return(apa)
```
atz: Definitely good to get rid of `expect_any_instance_of` 👍

PR #443, opened 2017-11-18
WOS record editors

darrenleeweber:
Split out of #379 

### Notes
I reviewed Russ Altman records to find examples of different roles, using:
```ruby
roles = altman_records.map {|rec| [rec.uid, rec.to_xml.scan(/(role=".*?")/) ] };
```
There are `author`, `book_editor` and `publisher` roles that apply to agents.  There is also a `researcher_id` role but that seems specific to WOS metadata management or something.

### DIFF
--- a/spec/lib/web_of_science/record_spec.rb
+++ b/spec/lib/web_of_science/record_spec.rb
@@ -60,30 +60,43 @@ describe WebOfScience::Record do
       expect(agents).to be_an Array
     end
     it 'contains name data' do
-      expect(agents.count).to eq 1
+      expect(agents).not_to be_empty
     end
-    it 'contains first_name' do
-      expect(agents.first).to include('first_name' => 'DC')
-    end
-    it 'contains last_name' do
-      expect(agents.first).to include('last_name' => 'WEBER')
-    end
-    it 'contains role' do
-      expect(agents.first).to include('role' => 'author')
+    it 'conforms to expected key-value pairs' do
+      expect(agent).to include('first_name' => String, 'last_name' => String, 'role' => String)
     end
     it 'missing attribute data is nil' do
-      expect(agents.first['reprint']).to be_nil
+      expect(agent).not_to include('reprint')
     end
   end
 
   describe '#authors' do
     let(:agents) { wos_record_encoded.authors }
+    let(:agent) { agents.first }
 
     it_behaves_like 'it is an array of names'
+    it 'contains author values' do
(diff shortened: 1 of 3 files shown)


spec/lib/web_of_science/record_spec.rb
```diff
@@ -60,30 +60,61 @@
       expect(agents).to be_an Array
     end
     it 'contains name data' do
-      expect(agents.count).to eq 1
+      expect(agents).not_to be_empty
     end
     it 'contains first_name' do
-      expect(agents.first).to include('first_name' => 'DC')
```
atz: Why is it better to test for any `String` value than the actual expected value?  

RSpec lets you test hashes very easily (in one `expect`), like:
```ruby
expect(agents.first).to include('first_name' => 'DC', 'last_name' => 'WEBER', 'role' => 'author')
```
The output will plainly identify which part or parts are missing.  Although I don't understand the imprecision, you could also write your new form as:
```ruby
expect(agents.first).to include('first_name' => String, 'last_name' => String, 'role' => String)
```
RSpec does the `be_a` match against the classname in exactly the same way.

atz: I see these are `shared_examples` and you are now testing divergent cases.  It looks like each _new_ case now tests `first_name`, `last_name` and `role` explicitly.  So the only value of using `shared_examples` for them is whether it is a non-empty `Array` or not and the `nil`-ness of reprint.  

You might want to break those minimal tests out to a different `shared_examples` set.  Then keep the explicit values for a stronger set (which would still cover `#authors` and `#names`).  Reducing the precision of the examples and then having to redo the same tests *with* precision seems like a loss.

darrenleeweber: The shared examples are used elsewhere in the spec on `names`, so they still serve the purpose of checking that the result of the parsing provides the expected key-value pairs in the Hash.  The more specific cases are checking details of the values in the key-value pairs, assuming that other specs have already validated that the key-value pairs exist and the value class/type is a String.  Seems OK to me.

darrenleeweber: As for style changes in the rspecs to check entire hashes - that is possible, for sure.  But if something breaks, it may not tell me why without looking at all the details of the hash.  If you want them collapses, I'm OK with doing it, just explaining the approach was to allow some pieces to pass and others to fail (or to isolate failure to a specific key-value pair).

darrenleeweber: OK, pushed up changes to consolidate a few specs.

PR #456, opened 2017-11-21
db filtering

### LINKED ISSUE #419: Limit results to WoS and Medline editions only
We are currently searching all databases using the WoK database identifier.  We want to limit results to only WoS and Medline results to prevent issues with parsing of other databases and potential duplication issues.  Clarivate tells us that using the <editions> filter does not work with limiting to specific databases.  Even though we only expect to get back WoS or MEDLINE records, we should filter against other types of records (e.g. CCC) just to be sure.

This could be a no-op -- if the code doesn't break when an unknown record is encountered, we could log this and ignore the record.  That is acceptable and this issue could then be closed. 

- [ ] can be done in the harvester/processor code (https://github.com/sul-dlss/sul_pub/blob/master/lib/web_of_science/process_records.rb)

atz:
Move `wos.rb` to `web_of_science.rb`
- our other classes are in the `WebOfScience` module (and subdir)
- added logger method
- updated only invocations of `WOS.` to match (in rake task)

Does the filtering and tests that log messages are sent.

Fixes #419

### DIFF
--- a/lib/web_of_science.rb
+++ b/lib/web_of_science.rb
@@ -1,5 +1,5 @@
 # Web of Science (WOS) utilities
-module WOS
+module WebOfScience
   # rubocop:disable Style/ClassVars
 
   # @return [WebOfScience::Harvester]
@@ -22,5 +22,8 @@ module WOS
     @@queries ||= WebOfScience::Queries.new(client)
   end
 
+  def self.logger
+    @@logger ||= Logger.new(Settings.WOS.LOG)
+  end
   # rubocop:enable Style/ClassVars
 end
--- a/lib/web_of_science/client.rb
+++ b/lib/web_of_science/client.rb
@@ -9,10 +9,11 @@ module WebOfScience
   # http://ipscience-help.thomsonreuters.com/wosWebServicesExpanded/WebServicesExpandedOverviewGroup/Introduction.html
   # It uses the savon gem for SOAP, see http://savonrb.com/version2/client.html
   class Client
-
     API_VERSION = '3.0'.freeze # Based on USER GUIDE July 7, 2015
     AUTH_WSDL = 'http://search.webofknowledge.com/esti/wokmws/ws/WOKMWSAuthenticate?wsdl'.freeze
     SEARCH_WSDL = 'http://search.webofknowledge.com/esti/wokmws/ws/WokSearch?wsdl'.freeze
+
+    delegate :logger, to: :WebOfScience
 
     def initialize(auth_code, log_level = :info)
       @auth_code = auth_code
@@ -75,12 +76,5 @@ module WebOfScience
       @search = nil
       @session_id = nil
     end
-
-    private
-
-      def logger
-        @logger ||= NotificationManager.wos_logger
-      end
   end
 end
-
--- a/lib/web_of_science/harvester.rb
+++ b/lib/web_of_science/harvester.rb
@@ -5,6 +5,7 @@ module WebOfScience
   # This class is responsible for processing WebOfScience API response data
   # to integrate it into the application data models.
   class Harvester
+    delegate :logger, to: :WebOfScience
 
     # Harvest all publications for an author
     # @param author [Author]
@@ -136,10 +137,6 @@ module WebOfScience
           wos_client = WebOfScience::Client.new(Settings.WOS.AUTH_CODE)
           WebOfScience::Queries.new(wos_client)
         end
-      end
-
-      def logger
-        @logger ||= NotificationManager.wos_logger
       end
   end
 end
--- a/lib/web_of_science/process_records.rb
+++ b/lib/web_of_science/process_records.rb
@@ -4,6 +4,7 @@ module WebOfScience
   # Process records retrieved by any means; this is a progressive filtering of the harvested records to identify
   # those records that should create a new Publication.pub_hash, PublicationIdentifier(s) and Contribution(s).
   class ProcessRecords
+    delegate :logger, to: :WebOfScience
 
     # @param author [Author]
     # @param records [WebOfScience::Records]
@@ -125,10 +126,6 @@ module WebOfScience
       def publication_identifier?(type, value)
         return false if value.nil?
         PublicationIdentifier.where(identifier_type: type, identifier_value: value).count > 0
-      end
-
-      def logger
-        @logger ||= NotificationManager.wos_logger
       end
   end
 end
--- a/lib/web_of_science/records.rb
+++ b/lib/web_of_science/records.rb
@@ -7,6 +7,8 @@ module WebOfScience
     extend Forwardable
     include Enumerable
 
+    delegate logger: :WebOfScience
+
     # @return [Integer] WOS record count
     def_delegators :rec_nodes, :count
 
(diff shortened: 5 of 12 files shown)


atz:
Not sure I care about the code-climate complaint here.

lib/web_of_science.rb
```diff
@@ -22,5 +22,8 @@ def self.queries
     @@queries ||= WebOfScience::Queries.new(client)
   end
 
+  def self.logger
+    @@logger ||= Logger.new(Settings.WOS.LOG)
+  end
```
darrenleeweber: This module doesn't need a logger because each of the classes that it wraps has their own logging.  This logger can be removed.

lib/web_of_science.rb
```diff
@@ -1,5 +1,5 @@
 # Web of Science (WOS) utilities
-module WOS
+module WebOfScience
```
darrenleeweber: OK, if you insist, but it's intended to be a console utility and typing `WOS` is easier than `WebOfScience`.

darrenleeweber: I suppose if your going to insist on this, then the rake task and namespace should also move from `wos` to `web_of_science` too.  So, either back out moving this WOS module or move the rake tasks too.

lib/web_of_science/client.rb
```diff
@@ -9,11 +9,12 @@ module WebOfScience
   # http://ipscience-help.thomsonreuters.com/wosWebServicesExpanded/WebServicesExpandedOverviewGroup/Introduction.html
   # It uses the savon gem for SOAP, see http://savonrb.com/version2/client.html
   class Client
-
     API_VERSION = '3.0'.freeze # Based on USER GUIDE July 7, 2015
     AUTH_WSDL = 'http://search.webofknowledge.com/esti/wokmws/ws/WOKMWSAuthenticate?wsdl'.freeze
     SEARCH_WSDL = 'http://search.webofknowledge.com/esti/wokmws/ws/WokSearch?wsdl'.freeze
 
+    delegate :logger, to: :WebOfScience
+
```
darrenleeweber: Does this require including the Forwardable module?

lib/web_of_science/harvester.rb
```diff
@@ -5,6 +5,7 @@ module WebOfScience
   # This class is responsible for processing WebOfScience API response data
   # to integrate it into the application data models.
   class Harvester
+    delegate :logger, to: :WebOfScience
```
darrenleeweber: Forwardable?  I prefer that it does not rely on rails magic for this to work.  The less rails magic the better in the business logic of the app.

lib/web_of_science/process_records.rb
```diff
@@ -4,6 +4,7 @@ module WebOfScience
   # Process records retrieved by any means; this is a progressive filtering of the harvested records to identify
   # those records that should create a new Publication.pub_hash, PublicationIdentifier(s) and Contribution(s).
   class ProcessRecords
+    delegate :logger, to: :WebOfScience
```
darrenleeweber: Forwardable.

spec/lib/notification_manager_spec.rb
```diff
@@ -66,17 +66,13 @@
...
-      expect(Logger).to receive(:new).with(Settings.WOS.LOG).once
-      described_class.error(exception, message, wos_client)
-    end
+    before { WebOfScience.class_variable_set(:@@logger, nil) }
+
     it 'logs errors' do
+      expect(WebOfScience).to receive(:logger).and_return(null_logger)
       expect(null_logger).to receive(:error).exactly(3)
       described_class.error(exception, message, wos_client)
     end
```
darrenleeweber: This seems to have changed the meaning of the spec.  It's no longer testing that the logger is a single instance of a Logger, let's put that back.  I don't know why calling the `NotificationManager.error` once should be logging 3 errors, that seems weird; shouldn't it log one error?

spec/lib/web_of_science/client_spec.rb
```diff
@@ -61,18 +61,12 @@
...
+      before { savon.expects(:close_session).returns(no_session_matches) }
+
+      it 'creates a logger and works' do
+        expect(WebOfScience).to receive(:logger).and_return(null_logger)
         expect(wos_client.session_close).to be_nil
       end
-      it 'creates a logger' do
-        expect(NotificationManager).to receive(:wos_logger).and_call_original
-        wos_client.session_close
-      end
```
darrenleeweber: Prefer to leave it as separate specs with only one expectation per spec.

lib/web_of_science.rb
darrenleeweber: Oh, I see the grand plan is move this responsibility from `NotificationManager` here.  Given the considerable consolidation of responsibility for these things into the `NotificationManager`, moving that responsibility to disparate locations in the app seems to add complexity rather than simplify the app.  In effect, it is not changing the way that `NotificationManager` is called or used to manage the generic `NotificationManager.error` calls; it just stripping the `NotficationManager` of control over how the logger is initialized.  Given all the delegation changes, it sorta makes sense to have the delegations contained in the module namespace.  So, OK, whatever.

lib/web_of_science/records.rb
```diff
@@ -28,19 +30,21 @@ def initialize(records: nil, encoded_records: nil)
...
-      end
-      db_recs
+      db_recs.map do |db, recs|
+        unless %w(WOS MEDLINE).include?(db)
+          logger.warn("Discarding #{recs.count} records from unknown database: #{db}")
+          next
+        end
+        [db, WebOfScience::Records.new(records: "<records>#{recs.map(&:to_xml).join}</records>")]
+      end.compact.to_h
     end
```
darrenleeweber: Please don't touch this method like this.  I don't like it at all.  This kind of thing should be the responsibility of the client using this method, not implemented like this as a nasty constraint on this method and the functionality of this class.  I really want this gone.

lib/web_of_science/records.rb
```diff
@@ -7,6 +7,8 @@ class Records
     extend Forwardable
     include Enumerable
 
+    delegate logger: :WebOfScience
```
darrenleeweber: OK, keep this but get it out of this commit and put it in an earlier commit where the other logging changes are made too.

lib/web_of_science.rb
atz: I am not concerned with the rake task namespace in this PR (or probably ever).

lib/web_of_science/client.rb
atz: No.  Quite annoyingly it is a slightly different syntax than `Forwardable`.  I used the `Forwardable` syntax in classes that already included it.  I used the stock syntax everywhere else.

lib/web_of_science/harvester.rb
atz: This is a rails app.  We rely on railsisms **throughout** the application, including the parts you imagine are now or someday could be separate.  I am disinclined to include additional dependencies in places where a perfectly functional (and more commonly used) solution is already available and, as demonstrated here, works perfectly well.  

If you want to push the issue, I will spin off a PR to remove every use of Forwardable.

spec/lib/notification_manager_spec.rb
atz: It isn't the `NotificationManager`'s concern anymore.  The responsibility is with the module.

darrenleeweber:
#476 now contains the logging.  The filtering in this PR needs to be moved elsewhere, into the harvest record processing.

PR #476, opened 2017-11-27
Wos logging

atz:
Split off from #456 (without the filtering implementation).

### DIFF
--- a/lib/notification_manager.rb
+++ b/lib/notification_manager.rb
@@ -18,7 +18,7 @@ class NotificationManager
       when CapAuthorsPoller, CapHttpClient
         log_exception(cap_logger, log_message, e)
       when WebOfScience::Client, WebOfScience::Harvester, WebOfScience::ProcessRecords, WebOfScience::Record
-        log_exception(wos_logger, log_message, e)
+        log_exception(WebOfScience.logger, log_message, e)
       else
         log_exception(Rails.logger, log_message, e)
       end
@@ -39,10 +39,6 @@ class NotificationManager
 
     def sciencewire_logger
       @@sciencewire_logger ||= Logger.new(Settings.SCIENCEWIRE.LOG)
-    end
-
-    def wos_logger
-      @@wos_logger ||= Logger.new(Settings.WOS.LOG)
     end
     # rubocop:enable Style/ClassVars
 
--- a/lib/web_of_science.rb
+++ b/lib/web_of_science.rb
@@ -1,5 +1,5 @@
 # Web of Science (WOS) utilities
-module WOS
+module WebOfScience
   # rubocop:disable Style/ClassVars
 
   # @return [WebOfScience::Harvester]
@@ -22,5 +22,8 @@ module WOS
     @@queries ||= WebOfScience::Queries.new(client)
   end
 
+  def self.logger
+    @@logger ||= Logger.new(Settings.WOS.LOG)
+  end
   # rubocop:enable Style/ClassVars
 end
--- a/spec/lib/notification_manager_spec.rb
+++ b/spec/lib/notification_manager_spec.rb
@@ -66,17 +66,13 @@ describe NotificationManager do
       described_class.error(exception, message, ScienceWireClient.new)
     end
   end
-  context '.wos_logger' do
+  context 'WebOfScience.logger' do
     let(:wos_client) { WebOfScience::Client.new(Settings.WOS.AUTH_CODE) }
 
-    before do
-      described_class.class_variable_set(:@@wos_logger, nil)
-    end
-    it 'creates a single logger' do
-      expect(Logger).to receive(:new).with(Settings.WOS.LOG).once
-      described_class.error(exception, message, wos_client)
-    end
+    before { WebOfScience.class_variable_set(:@@logger, nil) }
+
     it 'logs errors' do
+      expect(WebOfScience).to receive(:logger).and_return(null_logger)
       expect(null_logger).to receive(:error).exactly(3)
       described_class.error(exception, message, wos_client)
     end
--- a/spec/lib/web_of_science/client_spec.rb
+++ b/spec/lib/web_of_science/client_spec.rb
@@ -16,30 +16,27 @@ describe WebOfScience::Client do
 
   describe '#new' do
     it 'works' do
-      expect(wos_client).to be_an described_class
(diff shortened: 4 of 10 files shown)


lib/web_of_science.rb
```diff
@@ -22,5 +22,8 @@ def self.queries
     @@queries ||= WebOfScience::Queries.new(client)
   end
 
+  def self.logger
+    @@logger ||= Logger.new(Settings.WOS.LOG)
+  end
```
darrenleeweber: This commit combines moving the namespace plus adding a new logger - these should be separate commits.  Split this commit into separate commits and then squash the addition of the logger with another commit that removes it from the NotificationManager, that commit is then a clear, atomic move-method refactor.

lib/notification_manager.rb
```diff
@@ -40,10 +40,6 @@ def cap_logger
     def sciencewire_logger
       @@sciencewire_logger ||= Logger.new(Settings.SCIENCEWIRE.LOG)
     end
-
-    def wos_logger
-      @@wos_logger ||= Logger.new(Settings.WOS.LOG)
-    end
     # rubocop:enable Style/ClassVars
```
darrenleeweber: Merge this change with the addition of this method into the `WebOfScience` module, so it is clearer that this is a move-method refactor.

spec/lib/notification_manager_spec.rb
```diff
@@ -66,17 +66,13 @@
...
-      expect(Logger).to receive(:new).with(Settings.WOS.LOG).once
-      described_class.error(exception, message, wos_client)
-    end
+    before { WebOfScience.class_variable_set(:@@logger, nil) }
+
     it 'logs errors' do
+      expect(WebOfScience).to receive(:logger).and_return(null_logger)
       expect(null_logger).to receive(:error).exactly(3)
       described_class.error(exception, message, wos_client)
     end
```
darrenleeweber: Replace
- `expect(WebOfScience).to receive(:logger).and_return(null_logger)`
- with `expect(Logger).to receive(:new).with(Settings.WOS.LOG).once.and_return(null_logger)`

Not sure why the `null_logger` is receiving 3 errors, is that right?

lib/web_of_science.rb
atz: Why nitpick commit composition for changes you ostensibly want?  We are literally never going to use either state independently.  For this PR (and the previous one I already extracted these from), the **point** of the rename is to provide the logger method at the top `WebOfScience` module.  Breaking them apart gains nothing.

spec/lib/notification_manager_spec.rb
atz: Nope.  That is 100% NOT the concern of `NotificationManager` anymore.

spec/lib/web_of_science/client_spec.rb
```diff
@@ -61,18 +61,12 @@
...
+
+      it 'creates a logger and works' do
+        expect(WebOfScience).to receive(:logger).and_return(null_logger)
         expect(wos_client.session_close).to be_nil
       end
-      it 'creates a logger' do
-        expect(NotificationManager).to receive(:wos_logger).and_call_original
-        wos_client.session_close
-      end
     end
```
darrenleeweber: Best rspec pratices are to have only 1 expectation per spec.  This change is moving the code away from a best practice.  Apart from that, it is consistent with the intention to move the logger method (so that part of it is correct).

There is really only 1 line that needs to change in this spec.  At the previous line 66, the change would be to replace `NotificationManager` with `WebOfScience`.  That is the minimal change that works for the move-method refactor.

atz:
@darrenleeweber on #456 and in person, you already said you were "OK, with the logging and module name change".  How is it you now have numerous objections to exactly the same code?

spec/lib/web_of_science/client_spec.rb
atz: Read it again.  There is only one expectation.

spec/lib/notification_manager_spec.rb
darrenleeweber: Huh?  I'm not talking about `NotificationManager`, right.

darrenleeweber: BTW, I think it gets 3 errors because there are 3 retries in the Faraday client.

spec/lib/web_of_science/client_spec.rb
darrenleeweber: Huh?  The change collapses two specs into one, so it has two expectations, right?

darrenleeweber:
These changes will not conflict with other PRs, if/when this is merged.

lib/web_of_science.rb
darrenleeweber: I notice that renaming the module and moving the logger method are currently in two commits, but each commit contains a bit of both.  Either squash them all or separate the move module commit from the move method commit.

darrenleeweber: Yea, so just squash the first and last commits in this PR.

spec/lib/web_of_science_spec.rb
```diff
@@ -26,4 +26,10 @@
       expect(result).to be_an WebOfScience::Queries
     end
   end
+
+  describe '#logger' do
+    it 'works' do
+      expect(described_class.logger).to be_a Logger
+    end
+  end
```
darrenleeweber: The better spec is to test that it uses the `Settings.WOS.LOG` value, so moving that spec into this spec file from the NotificationManager could resolve the questions on how that changed.

lib/web_of_science.rb
atz: Squashed.

spec/lib/notification_manager_spec.rb
atz: I think that is right.  In any case, the number of times is unchanged.

darrenleeweber:
looks OK, but I want to hold off on a merge until a few other PRs are merged first.  This PR will require rebase for 3 or 4 open PRs because it changes the logger in all the WOS classes (not sure how many open PRs it will impact).  It will be much easier to rebase 476 than the other way around, the 476 rebase should be automatic and only needs to be done once after the other PRs are merged.

darrenleeweber:
Ah, OK, looks like we have already merged a few other PRs and only a couple remain.  Going ahead with this.

PR #477, opened 2017-11-28
WOS filter databases

### LINKED ISSUE #419: Limit results to WoS and Medline editions only

darrenleeweber:
Fix #419 

Selects only WOS & MEDLINE records for processing, using a configuration setting to specify what databases are allowed to be processed.

### DIFF
--- a/lib/web_of_science/process_records.rb
+++ b/lib/web_of_science/process_records.rb
@@ -55,7 +55,8 @@ module WebOfScience
       # @return [Array<WebOfScience::Record>]
       def match_wos_records(records)
         return [] if records.count.zero?
-        records.select { |rec| WebOfScienceSourceRecord.where(uid: rec.uid).count.zero? }
+        matching_uids = WebOfScienceSourceRecord.where(uid: records.map(&:uid)).pluck(:uid)
+        records.reject { |rec| matching_uids.include? rec.uid }
       end
 
       ## 3
--- a/lib/web_of_science/process_records.rb
+++ b/lib/web_of_science/process_records.rb
@@ -34,7 +34,8 @@ module WebOfScience
 
       # @return [Array<String>] WosUIDs that create a new Publication
       def create_publications
-        new_wos_records = match_wos_records # cf. WebOfScienceSourceRecord
+        filtered_records = filter_databases
+        new_wos_records = match_wos_records(filtered_records) # cf. WebOfScienceSourceRecord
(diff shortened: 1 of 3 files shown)


lib/web_of_science/process_records.rb
```diff
@@ -43,12 +44,21 @@ def create_publications
...
+        return records if Settings.WOS.ACCEPTED_DBS.empty?
+        records.select { |rec| Settings.WOS.ACCEPTED_DBS.include? rec.database }
       end
 
       ## 2
+      # Filter and select new WebOfScienceSourceRecords
+      # @return [Array<WebOfScience::Record>]
+      def match_wos_records(records)
+        return [] if records.count.zero?
+        records.select { |rec| WebOfScienceSourceRecord.where(uid: rec.uid).count.zero? }
```
atz: For 100 records, these does 100 queries on `WebOfScienceSourceRecord`'s table when it could just do one.

spec/lib/web_of_science/process_records_spec.rb
```diff
@@ -51,12 +51,10 @@
     # Happy paths
 
     it 'works' do
-      result = processor.execute
-      expect(result).not_to be_nil
+      expect(processor.execute).not_to be_nil
```
atz: 👍

lib/web_of_science/process_records.rb
darrenleeweber: Yes, that change could be made, although it wasn't the focus of this PR.  I'll go ahead and try to resolve that anyway in this PR.

darrenleeweber:
Pushed up more efficient filtering for existing WOS source records in f696ae07d5c8e95b7a84de9d5667bc2feea22f10

PR #478, opened 2017-11-29
Rake task to harvest from WebOfScience

### LINKED ISSUE #195: Create new harvesting rake tasks to use new harvest broker/request classes for WoS
- [x] create new rake tasks using new WoS harvest
- [ ] update the wiki
- [x] update the harvesting schedule via whenever

darrenleeweber:
Fix #195

### DIFF
--- a/lib/tasks/wos.rake
+++ b/lib/tasks/wos.rake
@@ -1,4 +1,9 @@
 namespace :wos do
+  desc 'Harvest from Web of Science, for all authors'
+  task harvest_authors: :environment do
+    WebOfScience.harvester.harvest_all
+  end
+
   desc 'Retrieve and print links for a publication by WOS-UID or WosItemId'
   task :links, [:wos_id] => :environment do |_t, args|
     raise 'wos_id argument is required.' if args[:wos_id].blank?
--- a/lib/web_of_science/harvester.rb
+++ b/lib/web_of_science/harvester.rb
@@ -7,6 +7,18 @@ module WebOfScience
   class Harvester
     delegate :logger, to: :WebOfScience
 
+    # @param [Enumerable<Author>] authors
+    # @return [nil]
+    def harvest(authors)
+      authors.each do |author|
+        begin
+          process_author(author)
+        rescue StandardError => e
+          NotificationManager.error(e, "WebOfScience harvest failed for author: '#{author.id}'", self)
+        end
+      end
+    end
+
     # Harvest all publications for an author
     # @param author [Author]
     # @return [Array<String>] WosUIDs that create Publications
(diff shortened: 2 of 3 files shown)


atz:
Deployment of this would be blocked by other work being required, right?  Or is that work already merged?

darrenleeweber:
Yes, the change in the schedule.rb file probably should not be deployed.  Maybe that change should be split out of this PR into a separate PR, so the other changes can be merged and they would be a deployment-no-op.

lib/tasks/wos.rake
```diff
@@ -1,4 +1,11 @@
 namespace :wos do
+  desc 'Harvest from Web of Science, for all authors'
+  task harvest_authors: :environment do
+    Author.where(active_in_cap: true, cap_import_enabled: true)
+          .find_in_batches(batch_size: 200)
+          .each { |authors| WOS.harvester.harvest(authors) }
```
atz: This just copies the code from the base class.  That seems unnecessary.

atz:
Can you give some detail about the problems you saw from trying to extend the harvester base class?  Like it went recursive or something?

darrenleeweber:
Yes, it went recursive.  For now, I'd rather we create an issue for that and try to resolve it in another PR, not here.  This is working OK.

lib/tasks/wos.rake
darrenleeweber: Yes.  Initially, when I tried to inherit from the base class, this task was just `WOS.harvester.harvest_all`.  It did not work because some specs went recursive.  I don't understand why and don't have time to resolve it, so this solution was adopted to get an expedient result that works.  Let's spin off a new issue to work on the WOS-harvester to use the base class (I don't have to time to replicate the problem and resolve it in PR).

atz:
OK, I retrieved your comment:
```
did try to use the `Harvester::Base` class but ran into some weird cyclic complexity issue in the specs and it could not load the class.  Dunno what that was about, so just worked around it for now.
```
That isn't directly related to recursion.  That's a stylistic complaint.  You should still use the base class.

lib/harvester/base.rb
```diff
@@ -2,17 +2,17 @@ module Harvester
...
-    # @return [void]
+    # @return [nil]
     def harvest_all
       Author.where(active_in_cap: true, cap_import_enabled: true)
             .find_in_batches(batch_size: batch_size)
             .each { |batch| harvest(batch) }
     end
 
     # @param [Enumerable<Author>] _authors
-    # @return [void]
```
atz: `void` is the correct return type here. 
https://github.com/lsegal/yard/blob/master/docs/GettingStarted.md#declaring-types

The value returned is not guaranteed or meaningful, and specifically is not guaranteed to be nil.

darrenleeweber: OK, thanks for that explanation and pointer to the doc about using `void`.  Seems weird that they recommend something commonly used in Java, for which the ruby equivalent is `nil`.  Seems to me like giving no type details at all would be better, in the case where it has no meaning or significance at all.  Anyhow, I can change this to keep the `void`.

darrenleeweber:
OK, fixed all the things.

Issue #452, opened 2017-11-20
For a MEDLINE record, extract mesh into pub_hash


peetucket:

darrenleeweber:
Example data in `MEDLINE:26776186`
```
<MeshHeadingList>
  <MeshHeading>
    <DescriptorName MajorTopicYN="N" UI="D019295">Computational Biology</DescriptorName>
    <QualifierName MajorTopicYN="N" UI="Q000379">methods</QualifierName>
    <QualifierName MajorTopicYN="N" UI="Q000706">statistics &amp; numerical data</QualifierName>
    <TreeCodeList>
      <TreeCode MajorTopicYN="N">H01.158.273.180</TreeCode>
      <TreeCode MajorTopicYN="N">L01.313.124</TreeCode>
      <TreeCode MajorTopicYN="N">Y33</TreeCode>
      <TreeCode MajorTopicYN="N">Y09</TreeCode>
    </TreeCodeList>
  </MeshHeading>
  <MeshHeading>
    <DescriptorName MajorTopicYN="N" UI="D057225">Data Mining</DescriptorName>
    <QualifierName MajorTopicYN="N" UI="Q000379">methods</QualifierName>
    <TreeCodeList>
      <TreeCode MajorTopicYN="N">L01.313.500.750.280.199</TreeCode>
      <TreeCode MajorTopicYN="N">L01.470.625</TreeCode>
      <TreeCode MajorTopicYN="N">Y33</TreeCode>
    </TreeCodeList>
  </MeshHeading>
  <MeshHeading>
    <DescriptorName MajorTopicYN="N" UI="D020000">Decision Support Systems, Clinical</DescriptorName>
    <QualifierName MajorTopicYN="Y" UI="Q000706">statistics &amp; numerical data</QualifierName>
    <TreeCodeList>
      <TreeCode MajorTopicYN="N">L01.313.500.750.300.190</TreeCode>
      <TreeCode MajorTopicYN="Y">Y09</TreeCode>
    </TreeCodeList>
  </MeshHeading>
  <MeshHeading>
    <DescriptorName MajorTopicYN="N" UI="D003955">Diagnostic Tests, Routine</DescriptorName>
    <QualifierName MajorTopicYN="N" UI="Q000706">statistics &amp; numerical data</QualifierName>
    <TreeCodeList>
      <TreeCode MajorTopicYN="N">E01.370.395</TreeCode>
      <TreeCode MajorTopicYN="N">Y09</TreeCode>
[...]

PR #480, opened 2017-11-29
WebOfScience create Publication (with PubMed additions), PublicationIdentifiers and Contributions

### LINKED ISSUE #316: Create a new publication record when we have a new record
After the matching logic in #301, if no match is found create a new publication with a pub_hash

Implementing create_publication into process_records class

- [ ] #368 -  Create pub_hash from WoS record using mapping described in #274 or MEDLINE record using mapping in #349
- [ ] #450 - Create new PublicationIdentifiers from rec.identifiers
- [ ] #451 - For WOS-record that has a PMID, fetch data from PubMed and merge new data into pub_hash
- [ ] associate publication with author (i.e. create contribution)

### LINKED ISSUE #450: Create new PublicationIdentifiers from rec.identifiers

### LINKED ISSUE #451: For WOS-record that has a PMID, fetch data from PubMed and merge new data into pub_hash

### LINKED ISSUE #453: After creating a publication, associate publication with author (i.e. create contribution)

darrenleeweber:
Fix #316 - create new Publications
Fix #450 - create new PublicationIdentifiers
Fix #451 - fetch and integrate PubMed data (for WOS-db records)
Fix #453 - create new Contributions

### Notes
- Work on #452 can be done to add MESH headings for MEDLINE-db records and, when that is done, no changes are required in process-records code, but the spec to check for the mesh-headings can be enabled.
- it would be nice if there was an easy way to tell VCR to use the same PubMed recording every time, but I don't have time now to figure that out.  Anyway, once they are recorded, it's done and should be OK.  The PubMed recordings are public data, no sensitive information needs to be filtered for them.


lib/web_of_science/process_records.rb
```diff
@@ -36,10 +36,10 @@ def execute
       # @return [Array<String>] WosUIDs that create a new Publication
       def create_publications
         filtered_records = filter_databases
```
atz: Would you be OK with skipping this unnecessary assignment and just using `filter_databases` as the param in the next line?

spec/lib/web_of_science/process_records_spec.rb
```diff
@@ -64,9 +64,43 @@
...
+    it 'creates new PublicationIdentifiers' do
+      expect { processor.execute }.to change { PublicationIdentifier.count }
+    end
+
+    it 'creates new Contributions' do
+      expect { processor.execute }.to change { Contribution.count }
+    end
+    it 'creates new contribution in the pub_hash[:authorship]' do
+      uids = processor.execute
+      pub = Publication.for_uid(uids.sample)
```
atz: Using `.sample` in test element selection makes the test suite unrepeatable and non-deterministic.  So even if it fails, you won't know how to repeat/fix that case.  If you want to test all the various objects, then test all of them.  Otherwise, pick any representative slice you want, but don't make it randomized per execution.  

This is not comparable to rspec's execution **order** being random.  That is fine.  The whole point of that is to emphasize how deterministic and isolated each example is.  This approach with `.sample` is the opposite.

darrenleeweber: Literally any one of these publications is representative, so using `sample` is fine here, it gets a representative Publication.  All of the Publication/Contribution models are known quantities.  All of the `uids` have been processed successfully, so any one of them is fine to test here.  No changes required.

lib/web_of_science/process_records.rb
darrenleeweber: This change could be made, but it's not essential to behavior, it's just a style preference AFAICT.

darrenleeweber: Anyway, done.

darrenleeweber:
OK, removed all the `.sample` from the specs.  It now tests all the contributions result in authorship.  It now tests all the results for MESH headings, when the source data has them (some of this is TODO work when the MEDLINE mapping works in #483 and will be done later, not in this PR).

lib/web_of_science/process_records.rb
```diff
@@ -89,21 +88,47 @@ def filter_by_identifiers(records)
...
+        record.uid
+      rescue StandardError => e
+        NotificationManager.error(e, "#{record.uid} failed to create Publication", self)
         nil
       end
 
+      # For WOS-record that has a PMID, fetch data from PubMed and enhance the pub.pub_hash with PubMed data
+      # @param [WebOfScience::Record] record
+      # @param [Publication] pub
+      # @return [Publication, nil]
```
atz: I agree with the TODO this seems like it might be more part of the publication domain, since it performs 2 updates to `pub`.  Return value specification seems not to matter since it is unused.  I don't see how it could ever return `nil` though (it raises instead).

darrenleeweber: - Actually, I _think_ I did that comparison and decided not to use it, because it doesn't do a reverse merge and that's what we need here to preserve the WOS record data and just update or add new data.  But, I'll take a second look at it just to be sure.
- Yea, maybe the return documentation is old, needs updating.

PR #483, opened 2017-12-02
WebOfScience::MapMesh

### LINKED ISSUE #452: For a MEDLINE record, extract mesh into pub_hash

darrenleeweber:
Fix #452

### DIFF
--- a/spec/lib/web_of_science/map_mesh_spec.rb
+++ b/spec/lib/web_of_science/map_mesh_spec.rb
@@ -1,0 +1,100 @@
+describe WebOfScience::MapMesh do
+  subject(:mapper) { described_class.new(record) }
+
+  let(:medline_xml) { File.read('spec/fixtures/wos_client/medline_record_26776186.xml') }
+  let(:record) { WebOfScience::Record.new(record: medline_xml) }
+  let(:pub_hash) { mapper.pub_hash }
+
+  describe '#new' do
+    it 'works with WOS records' do
+      expect(mapper).to be_an described_class
+    end
+    it 'raises ArgumentError with nil params' do
+      expect { described_class.new }.to raise_error(ArgumentError)
+    end
+    it 'raises ArgumentError with anything other than WebOfScience::Record' do
+      expect { described_class.new('could be xml') }.to raise_error(ArgumentError)
+    end
+  end
+
+  shared_examples 'medline_works' do
+    it 'works with MEDLINE records' do
+      expect(mapper).to be_an described_class
+    end
+  end
+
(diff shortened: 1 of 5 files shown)


spec/lib/web_of_science/map_mesh_spec.rb
```diff
@@ -0,0 +1,103 @@
...
+    end
+  end
+
+  shared_examples 'pub_hash' do
+    it 'works' do
+      expect(pub_hash).to be_an Hash
+    end
+  end
+
+  shared_examples 'mesh' do
```
atz: This whole `shared_examples` block is only used once, but it seems to have moved to a level of abstraction with `include(name: String, major: String, id: String)` that would only be useful if it is used many times.  IMO, it should just be folded into the one block that invokes it and have concrete values instead of `String`.  

My preference (despite rubocop-rspec's inane quibbling) would also be to combine the two expectations like:
```ruby
expect(tree_codes).to be_an Array
expect(tree_codes.first).to include(code: String, major: String)
```
`.first` is going to fail if the object isn't Array-like anyway, so there isn't much point keeping it separate.

We could collapse all these down to a single expectation:
```ruby
      expect(pub_hash[:mesh_headings]).to include a_hash_including(
          descriptor: a_collection_including(name: 'A', major: 'B', id: 'C'),
          qualifier: a_collection_including(name: 'Q', major: 'R', id: 'S'),
          treecode: a_collection_including(code: 'X', major: 'Y')
      )
```

atz: The output from rspec upon failure (e.g., after assigning `treecode[:major] = 'Z'`) looks like:
```diff
-[(a hash including {:descriptor => (a collection including {:name => "A", :major => "B", :id => "C"}), 
-:qualifier => (a collection including {:name => "Q", :major => "R", :id => "S"}), 
-:treecode => (a collection including {:code => "X", :major => "Y"})})]
+[{:descriptor=>{:name=>"A", :major=>"B", :id=>"C"},
+  :qualifier=>{:name=>"Q", :major=>"R", :id=>"S"},
+  :treecode=>{:code=>"X", :major=>"Z"}}]
```
Some people freak out that there is more info than before, but that seems perfectly readable to me.  One major advantage is you get to see the values that ARE there, not just that the ones you are looking for aren't there.  I.E., if the keys get renamed or turn from symbols to strings, the output will show that instead of just saying `expected 'Y', got nil`.

darrenleeweber: OK, done

PR #491, opened 2017-12-04
sul_pub-368-Refactoring

Maatary:
related to #368


lib/web_of_science/data/records.rb
```diff
@@ -23,10 +24,25 @@ class Records
...
+      # @return [Nokogiri::XML::Document]
+      # @raise RuntimeError when arguments are all nil
+      def parse(xml, encoded_xml)
+        xml ||= begin
+          raise 'encoded_xml is nil' if encoded_xml.nil?
+          coder = HTMLEntities.new
+          coder.decode(encoded_xml)
+        end
+        Nokogiri::XML(xml) { |config| config.strict.noblanks }
       end
```
darrenleeweber: This change will undo prior refactoring and break the code when it needs to work without an encoded record, if I understand this correctly.  I suggest this commit is dropped.

spec/lib/web_of_science/xml_parser_spec.rb
```diff
@@ -1,32 +0,0 @@
-require 'htmlentities'
```
darrenleeweber: I want this class and it's spec; I can't approve this.

spec/lib/web_of_science/services/hash_mappers/names_mapper_spec.rb
```diff
@@ -1,21 +1,15 @@
-describe WebOfScience::MapNames do
+describe WebOfScience::Services::HashMappers::NamesMapper do
```
darrenleeweber: What does this module complexity buy us?  The extra module depth doesn't seem warranted for a relatively small set of files in the existing module.  If this were a much larger module or app, maybe it would be necessary.  As it is, it seems like semantic wrapping (not sure if it has any functional significance).

spec/lib/web_of_science/services/hash_mappers/mapper_spec.rb
```diff
@@ -1,24 +0,0 @@
-describe WebOfScience::Mapper do
```
darrenleeweber: I don't understand why this one commit that seems to point to something done on MapNames is also removing the spec on the mapper base class.

spec/lib/web_of_science/services/hash_mappers/names_mapper_spec.rb
Maatary: The idea is that we would move every service type class in the Services module. So you could have a Services::Connection, Services::Harvester, and so on

atz: Yes, but what does that give us?

atz:
2000 line delta, over 20 commits is pretty messy.  I don't think this PR is workable in its current form.

darrenleeweber:
This work is being abandoned in the current work cycle.  The commit cleanup and conflicts are problematic.  The principles of the design are sound and valuable, the details of the commits and some minor quibbles with the depth of the modules are getting in the way.  Closing this PR as we are not going to pursue it, but leaving the branch available for later review.

PR #508, opened 2017-12-04
Harvesters ENABLED by settings

### LINKED ISSUE #507: Create feature flag for sciencewire client vs web of science client
Add a config boolean option to be used elsewhere in the code for deciding which client to use

darrenleeweber:
Fix #507

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -34,6 +34,7 @@ PUBMED:
 
 ## Sciencewire Auth Config
 SCIENCEWIRE:
+  enabled: true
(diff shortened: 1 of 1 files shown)


atz:
@darrenleeweber ALLCAPS?  That seems like it is a holdover from when these were `CONSTANTS`.  Do we want to stick w/ that?

darrenleeweber:
@atz - good question, I suppose it could be lower case.

darrenleeweber:
@atz - done, lower case `Settings.<x>.enabled`

PR #510, opened 2017-12-05
Moderate cleanup of authorship_api_spec

atz:
All the same tests, 119 fewer lines.

- Suppress copy-pasta.
- Use common `let` statements.
- Use `include` w/ multiple params
- Remove unneeded single-use or redundant `let` statements
- Remove unneeded intermediate variables
- replace `.where(...).first` with `find_by(...)`
- Use `a_hash_including` matcher to replace multiple hash interrogations
  - Don't need to check `.not_to be_nil` first before checking `include`
    The latter implies the former.

### DIFF
--- a/spec/api/sul_bib/authorship_api_spec.rb
+++ b/spec/api/sul_bib/authorship_api_spec.rb
@@ -1,9 +1,10 @@
-
+# Various factories called below (via "create")
+# @see spec/factories/publication.rb
+# @see spec/factories/author.rb
+# @see spec/factories/contribution.rb
 describe SulBib::API, :vcr do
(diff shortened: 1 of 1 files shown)


spec/api/sul_bib/authorship_api_spec.rb
```diff
@@ -565,53 +479,34 @@ def check_response_error(code)
             visibility: 'private'
           )
         end
+        let(:result) { JSON.parse(response.body) }
+        let(:id) { '0' }
```
darrenleeweber: `sul_pub_id` is more informative than `id`

Issue #489, opened 2017-12-04
Manually review citations for WOS Publications


darrenleeweber:
The `Csl::Citation` mapping for WOS Publications might need to be updated to create well formed citations.  This includes detection of the `wos` provenance and using the correct author/editor, doc-types and other mappings.

connected to #512

darrenleeweber:
Evaluating some data using

PR #512, opened 2017-12-05
[WIP] BibtexMapper

darrenleeweber:
[ This PR was too large and has been split out into bite size PRs; the remaining work is now focused on just the Bibtex code ]

Some code-climate issues are old tech-debt that are not intended to be fixed in this PR.  Also, most of the specs are not changed and most of these refactors are simple extract/move refactors that rest on green, there should be no significant change in code coverage.

Connected to #489


lib/csl/role_mapper.rb
```diff
@@ -0,0 +1,110 @@
+module Csl
+
+  class RoleMapper
+
+    class << self
+
+      # Parse authors
+      def authors(pub_hash)
+        return [] unless pub_hash[:author].present?
```
atz: Does it make sense to do:
```ruby
return [] unless pub_hash && pub_hash[:author].present?
```
in case a `nil` value gets passed in here?  Seems preferable to raising an exception.

atz: I know that isn't the point of this PR, just something I noticed.

lib/csl/role_mapper.rb
```diff
@@ -0,0 +1,110 @@
...
+          # SciencewireSourceRecord.convert_sw_publication_doc_to_hash
+          Csl::SciencewireMapper.authors_to_csl(authors)
+        else
+          parse_authors(authors)
+        end
+      end
+
+      # Parse editors for various provenance data:
+      def editors(pub_hash)
+        return [] unless pub_hash[:author].present?
```
atz: Same here.

lib/csl/bibtex_mapper.rb
```diff
@@ -0,0 +1,19 @@
+module Csl
+
+  # Convert BibTexIngester content into a CSL format
+  class BibtexMapper
+
+    # Convert BibTexIngester authors into CSL authors
+    # @param [Array<Hash>] authors array of hash data
+    # @return [Array<Hash>] CSL authors array of hash data
+    def self.authors_to_csl(authors)
```
atz: To clarify, you are relying on the previously established test coverage on `Csl::Mapper` to exercise these newly public methods?

lib/csl/role_mapper.rb
darrenleeweber: Well, that would avoid an exception, but we're in big trouble here if the pub_hash is nil.  So maybe that _should_ raise an exception.

lib/csl/bibtex_mapper.rb
darrenleeweber: Yes; and they do.  But now that they are public, that seems to entail they should have their own specs.  I'll see if I can quickly add specs (I just don't want to waste too much time on it, as they are sorta tested already).

darrenleeweber:
This bibtex refactor is not a high priority, it can be dropped.

PR #514, opened 2017-12-06
Authorship api coerce params

### LINKED ISSUE #509: Authorship API blows up when trying to add a publication to a user's profile
The "contrib_attr" method in app/api/sul_bib/authorship_api.rb is setting the visibility/status/featured attributes which is then later used in the "create_or_update_and_return_pub_hash" method to assign to a new contribution, but the incoming visibility attribute from the Profiles API call is upcased, and this fails when validating the contribution.

We may want to do a before_validation callback on the contributions model and call the existing 'init' method there.

atz:
Fix #509 

Continuation of spec cleanup, then added Grape param coercion w/ tests.

### DIFF
--- a/app/api/sul_bib/authorship_api.rb
+++ b/app/api/sul_bib/authorship_api.rb
@@ -5,11 +5,9 @@ module SulBib
 
     helpers do
       # Extract a hash of optional contribution parameters from the request.
-      # When this method is called for a PATCH request, it's important that it
-      # does not set any defaults.
-      # @return [Hash] May contain any of :features, :status and :visibility
+      # When this method is called for a PATCH request, it's important that it does not set any defaults.
+      # @return [Hash<Symbol => [String, Boolean]>] May contain any of :features, :status and :visibility
       def contrib_attr
-        # Gather optional contribution fields.
         contrib_attr = {}
         [:featured, :status, :visibility].each do |field|
           # check params[field].nil? not .blank? because featured can be `false`.
@@ -26,8 +24,8 @@ module SulBib
         contrib = pub.contributions.find_or_initialize_by(author_id: author.id)
         contrib.assign_attributes(authorship.merge(cap_profile_id: author.cap_profile_id, author_id: author.id))
         pub.pubhash_needs_update! if contrib.persisted?
-        contrib.save
         begin
+          contrib.save!
           # Save the publication so it will sync the contribution into the pub.pub_hash[:authorship] array.
           pub.save!
         rescue ActiveRecord::ActiveRecordError => e
@@ -199,8 +197,8 @@ module SulBib
       # the other attributes of an existing contribution.  Because this POST
       # conflates creation with update, these attributes must be explicit.
       requires :featured, type: Boolean, desc: 'The JSON body must indicate if the contribution is featured'
-      requires :status, type: String, desc: 'The JSON body must indicate if the contribution is approved'
-      requires :visibility, type: String, desc: 'The JSON body must indicate if the contribution is visible'
+      requires :status, type: String, desc: 'The JSON body must indicate if the contribution is approved', coerce_with: ->(val) { val.downcase }
+      requires :visibility, type: String, desc: 'The JSON body must indicate if the contribution is visible', coerce_with: ->(val) { val.downcase }
     end
     post do
       request_body_unparsed = env['api.request.input']
@@ -259,8 +257,8 @@ module SulBib
       # An update request can omit any of the contribution attributes, so long
       # as no defaults are specified here.
       optional :featured, type: Boolean, desc: 'The JSON body should indicate if the contribution is featured'
-      optional :status, type: String, desc: 'The JSON body should indicate if the contribution is approved'
-      optional :visibility, type: String, desc: 'The JSON body should indicate if the contribution is visible'
+      optional :status, type: String, desc: 'The JSON body should indicate if the contribution is approved', coerce_with: ->(val) { val.downcase }
+      optional :visibility, type: String, desc: 'The JSON body should indicate if the contribution is visible', coerce_with: ->(val) { val.downcase }
     end
     patch do
       request_body_unparsed = env['api.request.input']
--- a/spec/api/sul_bib/authorship_api_spec.rb
+++ b/spec/api/sul_bib/authorship_api_spec.rb
@@ -364,6 +364,15 @@ describe SulBib::API, :vcr do
         it_behaves_like 'it creates new contributions and publications' # TODO: modifies existing contributions.
       end
 
+      context 'with allcaps or mixed case strings' do
+        let(:request_data) { sul_author_hash.merge(sul_pub_id: publication_with_contributions.id, visibility: 'PRIVATE', status: 'New', featured: true) }
+        it 'downcases appropriately' do
+          expect { http_request }.not_to change { existing_contrib }
+          contrib = publication_with_contributions.contributions.reload.last
+          expect(contrib.status).to eq 'new'
+          expect(contrib.visibility).to eq 'private'
+        end
+      end
     end # context 'success'
 
     context 'failure' do
@@ -463,6 +472,16 @@ describe SulBib::API, :vcr do
           http_request # defined in context, uses request_data
           expect { existing_contrib.reload }.not_to change { [existing_contrib.status, existing_contrib.featured] }
           expect(existing_contrib.visibility).to eq 'private'
+        end
+      end
+
+      context 'with allcaps or mixed case strings' do
+        let(:request_data) { existing_contrib_ids.merge(visibility: 'PUBLIC', status: 'New') }
+        it 'downcases appropriately' do
+          http_request # defined in context, uses request_data
+          expect { existing_contrib.reload }.not_to change { existing_contrib.featured }
+          expect(existing_contrib.status).to eq 'new'
+          expect(existing_contrib.visibility).to eq 'public'
         end
       end
     end # context 'success'
--- a/spec/api/sul_bib/authorship_api_spec.rb
+++ b/spec/api/sul_bib/authorship_api_spec.rb
@@ -6,13 +6,7 @@ describe SulBib::API, :vcr do
   let(:headers) { { 'HTTP_CAPKEY' => Settings.API_KEY, 'CONTENT_TYPE' => 'application/json' } }
   let(:publication) { create :publication }
   let(:author) { create :author }
-  # These authorships can be merged with any contribution hash
-  let(:cap_author_hash) do
-    { cap_profile_id: author.cap_profile_id }
-  end
-  let(:sul_author_hash) do
-    { sul_author_id: author.id }
-  end
+  let(:sul_author_hash) { Hash[sul_author_id: author.id] }
   let(:contribution_count) { 2 }
   # let is lazy-evaluated: it is evaluated the first time it's method is invoked.
   # Use let! to force the method's invocation before each example.
@@ -24,43 +18,15 @@ describe SulBib::API, :vcr do
     pub.save # to update the pub.pub_hash
     pub
   end
-  let(:valid_data_for_post) do
+  let(:base_data) do
     {
-      sul_pub_id: publication.id,
       featured: false,
       status: 'denied',
       visibility: 'private',
     }
   end
-  let(:new_authorship_for_pub_with_contributions) do
-    # status 'approved', visibility 'public', featured true
-    # In this submission, use different values.
-    {
-      sul_pub_id: publication_with_contributions.id,
-      featured: false,
-      status: 'denied',
-      visibility: 'private'
-    }
-  end
-  let(:authorship_for_sw_id) do
-    {
-      sw_id: '10379039',
-      featured: false,
-      status: 'denied',
-      visibility: 'private'
-    }
-  end
-  let(:authorship_for_pmid) do
-    {
-      pmid: '23684686',
-      featured: false,
-      status: 'denied',
-      visibility: 'private'
-    }
-  end
-  let(:existing_contrib) do
-    publication_with_contributions.contributions.first
-  end
+  let(:valid_data_for_post) { base_data.merge(sul_pub_id: publication.id) }
+  let(:existing_contrib) { publication_with_contributions.contributions.first }
   let(:existing_contrib_ids) do
     {
       # author_hash is NOT merged here, because this data is concerned with an existing contribution.
@@ -68,13 +34,7 @@ describe SulBib::API, :vcr do
       sul_pub_id: existing_contrib.publication.id
     }
   end
-  let(:update_authorship_for_pub_with_contributions) do
-    existing_contrib_ids.merge(
-      featured: false,
-      status: 'denied',
-      visibility: 'private'
-    )
-  end
+  let(:update_authorship_for_pub_with_contributions) { existing_contrib_ids.merge(base_data) }
 
   # For PATCH, the attribute params are optional, so only include those to be updated.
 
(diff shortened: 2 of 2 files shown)


spec/api/sul_bib/authorship_api_spec.rb
```diff
@@ -24,57 +18,24 @@
...
     pub
   end
-  let(:valid_data_for_post) do
+  let(:base_data) do
     {
-      sul_pub_id: publication.id,
       featured: false,
       status: 'denied',
       visibility: 'private',
     }
```
darrenleeweber: Looks like `base_data` is contribution attributes now; maybe naming as such is more meaningful?

spec/api/sul_bib/authorship_api_spec.rb
```diff
@@ -24,57 +18,24 @@
...
-      status: 'denied',
-      visibility: 'private'
-    }
-  end
-  let(:existing_contrib) do
-    publication_with_contributions.contributions.first
-  end
+  let(:valid_data_for_post) { base_data.merge(sul_pub_id: publication.id) }
+  let(:new_authorship_for_pub_with_contributions) { base_data.merge(sul_pub_id: publication_with_contributions.id) }
+  let(:existing_contrib) { publication_with_contributions.contributions.first }
```
darrenleeweber: The reason all these data are grouped up here at the top of the spec is so it's easy to see what is different between them.  Looks like `authorship_for_sw_id` and `authorship_for_pmid`  moved somewhere else.  It's not a blocker, just noting that when they are all together it's easier to see what's different.

spec/api/sul_bib/authorship_api_spec.rb
```diff
@@ -6,13 +6,7 @@
...
   let(:publication) { create :publication }
   let(:author) { create :author }
-  # These authorships can be merged with any contribution hash
-  let(:cap_author_hash) do
-    { cap_profile_id: author.cap_profile_id }
-  end
-  let(:sul_author_hash) do
-    { sul_author_id: author.id }
-  end
+  let(:sul_author_hash) { Hash[sul_author_id: author.id] }
```
darrenleeweber: I guess `cap_author_hash` moved somewhere else.  The idea of grouping it all here is to see how it is based on the author.  Apart from one-liners, don't care much for this change, but it doesn't have to be a blocker, unless cap_author_hash disappeared (prob not).

spec/api/sul_bib/authorship_api_spec.rb
```diff
@@ -263,9 +217,7 @@
     it_behaves_like 'it is successful'
 
     it 'does not create a new contribution' do
-      expect do
-        http_request # defined in context, uses request_data
-      end.not_to change(Contribution, :count)
+      expect { http_request }.not_to change(Contribution, :count)
     end
```
darrenleeweber: The explanatory comment was there to dispel mysteries, esp. for junior devs or anyone not familiar with shared examples or let precedence.  It can also help with finding any lines related to `request_data`.  Could we find a way to accomplish this (if it's not done elsewhere)?

spec/api/sul_bib/authorship_api_spec.rb
```diff
@@ -437,7 +385,7 @@ def check_response_error(code)
       end
 
       context 'with cap_profile_id' do
-        let(:author_hash) { cap_author_hash }
+        let(:author_hash) { Hash[cap_profile_id: author.cap_profile_id] }
         it_behaves_like 'it creates new contributions and publications'
```
darrenleeweber: Ah, here it is.  So `cap_author_hash` has disappeared.  OK, if this is the only place it was used.

darrenleeweber:
I'm sorta OK with the first commit - fabe01afc57f3192a4ec86d080a6100981d6f379
I'll look over the 2nd commit, when I'm done looking at it I might ask for a squash on those spec cleanup commits.

spec/api/sul_bib/authorship_api_spec.rb
```diff
@@ -44,44 +43,24 @@
...
   shared_examples 'it creates new contributions and publications' do
     context 'with no other contributions' do
       let(:request_data) { valid_data_for_post.merge(author_hash) }
-      it_behaves_like 'it is successful'
-      it_behaves_like 'it creates a new contribution'
+      it 'successfully creates one new contribution' do
+        expect { http_request }.to change(Contribution, :count).by(1)
+        expect(response.status).to eq(201)
+      end
     end
```
darrenleeweber: OK, obviously we differ on grouping more than one expect in a spec, but that's not too important here.  The prior success check was obviously generic (any 200+ response is a success).  I suppose it could also use a `response.success?` or something similiar in an rspec dialect.  So long as _everything_ using this shared example must be returning a 201, this change is OK.  I actually don't know if that's reasonable or not for this API endpoint or the specs using this shared example.  It does make the shared example a little less share-able.

atz: Yeah, I made sure all the statuses lined up and preferred being exact where possible.  This keeps some of the insane nesting of `shared_examples` down.  

In integration tests in particular, I find one almost always needs a separate `expect` for checking status.  It doesn't make sense to me to have a separate request/response cycle for a complex operation and then just check if `status == 200` and throw it away.

spec/api/sul_bib/authorship_api_spec.rb
```diff
@@ -44,44 +43,24 @@
...
-      it 'increases number of contribution records for specified publication by one' do
-        expect do
-          http_request
-        end.to change(publication_with_contributions.contributions, :count).by(1)
+      let(:request_data) { base_data.merge(sul_pub_id: publication_with_contributions.id).merge(author_hash) }
+      it 'successfully increases the publication\'s contribution records by one' do
+        expect { http_request }.to change(publication_with_contributions.contributions, :count).by(1)
+        expect(response).to have_http_status(:success)
+        expect(publication_with_contributions.contributions.count).to eq(contribution_count + 1)
       end
```
darrenleeweber: I don't get how this clarifies stuff any - looks like it could be calling the new shared example and be done with it?  But the shared example now expects a 201 whereas this one uses a different rspec dialect to check for success.  So, if this one is not a 201, then change the shared example to use this rspec dialect and have this use the shared example?  I suppose this is an example of what it means to change some shared-example nesting, but this looks more complicated and less shared.  The whole point of shared examples is to consolidate code, not duplicate it.  Like, lines 61 and 63 in the new branch seem to do the same thing, don't they?

Agreed that using the `expect {}.to change` is a good pattern when we don't care about details.  (I might have coded some of this stuff in my early adjustments to using rspec, it looks like some things I would do differently now.)

spec/api/sul_bib/authorship_api_spec.rb
atz: Sorry, what are you referring to?  You're commenting on an unchanged line.   The comment above (L186/R163) was changed only inasmuch as the keys were corrected to be strings (not symbols).  Are you describing something else?

spec/api/sul_bib/authorship_api_spec.rb
atz: Not *exactly* the same thing.  The former checks that the increment is 1.  The latter checks the total value.  I agree it might be overkill, but the tests were both there previously.  Want to get rid of one?

spec/api/sul_bib/authorship_api_spec.rb
```diff
@@ -213,21 +190,19 @@
...
       http_request # defined in context, uses request_data
       existing_contrib.reload
-      expect(request_data[:featured]).not_to be_nil
+      expect(request_data).to include(featured: !be_nil, status: be_present, visibility: be_present)
       expect(existing_contrib.featured).to be request_data[:featured]
-      expect(request_data[:status]).not_to be_nil
       expect(existing_contrib.status).to eq request_data[:status]
-      expect(request_data[:visibility]).not_to be_nil
       expect(existing_contrib.visibility).to eq request_data[:visibility]
     end
```
darrenleeweber: `!be_nil` is different from `be_present`?  If this is in shared_examples, it might be shared between various HTTP methods, hence the check on the request payload(s); i.e. a PATCH could omit some.

spec/api/sul_bib/authorship_api_spec.rb
```diff
@@ -372,36 +347,26 @@ def check_response_error(code)
...
         let(:author_hash) { Hash[cap_profile_id: author.cap_profile_id] }
-        it_behaves_like 'it creates new contributions and publications'
-        # TODO: modifies existing contributions.
+        it_behaves_like 'it creates new contributions and publications' # TODO: modifies existing contributions.
       end
 
-      context 'to update a contribution' do
-        let(:request_data) { update_authorship_for_pub_with_contributions }
-        it_behaves_like 'it updates an existing contribution'
-      end
```
darrenleeweber: Seems like this one changed the request_data before but now it's using all the same request data as the others.

spec/api/sul_bib/authorship_api_spec.rb
```diff
@@ -372,36 +347,26 @@ def check_response_error(code)
...
 
   context 'POST /authorship' do
+    let(:http_request) { post '/authorship', request_data.to_json, headers }
+
     context 'success' do
-      let(:http_request) do
-        post '/authorship', request_data.to_json, headers
-        expect(response.status).to eq(201)
-      end
+      after(:example) { expect(response.status).to eq(201) }
```
darrenleeweber: The shared example(s) now checks for success (right?), so does this `after` block now check it again?

spec/api/sul_bib/authorship_api_spec.rb
```diff
@@ -458,71 +423,51 @@ def check_response_error(code)
...
       # requests that update a contribution (where the author already exists).
       # For all the PATCH specs, the contribution already exists, so testing
       # different author params is not important.
+      after(:example) { expect(response.status).to eq(200) }
 
-      context 'to update all contribution attributes' do
-        let(:request_data) { update_authorship_for_pub_with_contributions }
-        it_behaves_like 'it updates an existing contribution'
-      end
+      it_behaves_like 'it updates an existing contribution'
```
darrenleeweber: I seem to be missing what's happened to the request data in the "new" vs. "update" contexts.

spec/api/sul_bib/authorship_api_spec.rb
```diff
@@ -458,71 +423,51 @@ def check_response_error(code)
...
-          expect(request_data[:visibility]).to be_nil
+          expect(request_data).not_to include(:status, :visibility)
           http_request # defined in context, uses request_data
-          existing_contrib.reload
-          expect(existing_contrib.featured).to be request_data[:featured]
-          expect(existing_contrib.status).to eq 'approved'
-          expect(existing_contrib.visibility).to eq 'public'
+          expect { existing_contrib.reload }.not_to change { [existing_contrib.visibility, existing_contrib.status] }
+          expect(existing_contrib.featured).to be false
         end
```
darrenleeweber: :+1:

spec/api/sul_bib/authorship_api_spec.rb
```diff
@@ -458,71 +423,51 @@ def check_response_error(code)
...
-          expect(request_data[:visibility]).to be_nil
+          expect(request_data).not_to include(:featured, :visibility)
           http_request # defined in context, uses request_data
-          existing_contrib.reload
-          expect(existing_contrib.featured).to be true
-          expect(existing_contrib.status).to eq request_data[:status]
-          expect(existing_contrib.visibility).to eq 'public'
+          expect { existing_contrib.reload }.not_to change { [existing_contrib.visibility, existing_contrib.featured] }
+          expect(existing_contrib.status).to eq 'denied'
         end
```
darrenleeweber: :+1:

spec/api/sul_bib/authorship_api_spec.rb
```diff
@@ -458,71 +423,51 @@ def check_response_error(code)
   # PATCH
 
   context 'PATCH /authorship' do
-    context 'success' do
-      let(:http_request) do
-        patch '/authorship', request_data.to_json, headers
-        expect(response.status).to eq(200)
-      end
+    let(:http_request) { patch '/authorship', request_data.to_json, headers }
```
darrenleeweber: :+1: - I can see the extraction here for both success and failure, with the addition of the after block on success.  Just wondering whether the failure context could have a generic after block to test for failure (it's not there before, so this is asking for an addition that may not be necessary, it's not a blocker).

spec/api/sul_bib/authorship_api_spec.rb
```diff
@@ -364,6 +364,15 @@ def check_response_error(code)
         it_behaves_like 'it creates new contributions and publications' # TODO: modifies existing contributions.
       end
 
+      context 'with allcaps or mixed case strings' do
+        let(:request_data) { sul_author_hash.merge(sul_pub_id: publication_with_contributions.id, visibility: 'PRIVATE', status: 'New', featured: true) }
+        it 'downcases appropriately' do
+          expect { http_request }.not_to change { existing_contrib }
+          contrib = publication_with_contributions.contributions.reload.last
+          expect(contrib.status).to eq 'new'
+          expect(contrib.visibility).to eq 'private'
```
darrenleeweber: The previous commit introduced some new patterns for checking the changes to attributes, use them here?  It was something like: `expect {}.not_to change { [attr1, ... , attrN ] }`.  Maybe this spec is also wanting to check there are no changes to the `Contribution.count`?  It seems to be checking that the last contribution is the same before/after, like it's not sure if the last contrib _is_ the `existing_contrib`?

spec/api/sul_bib/authorship_api_spec.rb
```diff
@@ -465,6 +474,16 @@ def check_response_error(code)
...
+
+      context 'with allcaps or mixed case strings' do
+        let(:request_data) { existing_contrib_ids.merge(visibility: 'PUBLIC', status: 'New') }
+        it 'downcases appropriately' do
+          http_request # defined in context, uses request_data
+          expect { existing_contrib.reload }.not_to change { existing_contrib.featured }
+          expect(existing_contrib.status).to eq 'new'
+          expect(existing_contrib.visibility).to eq 'public'
+        end
+      end
```
darrenleeweber: - Maybe the POST and PATCH could use a shared_example for these specs?
- seems like testing for no changes to the `Contribution.count` is part of this case-sensitivity context?

app/api/sul_bib/authorship_api.rb
```diff
@@ -26,8 +24,8 @@ def create_or_update_and_return_pub_hash(pub, author, authorship)
         contrib = pub.contributions.find_or_initialize_by(author_id: author.id)
         contrib.assign_attributes(authorship.merge(cap_profile_id: author.cap_profile_id, author_id: author.id))
         pub.pubhash_needs_update! if contrib.persisted?
-        contrib.save
         begin
+          contrib.save!
```
darrenleeweber: Does this change entail adding a spec for the exception handing or logging?  Maybe not, as it just appears like a response failure anyway, I dunno.

spec/api/sul_bib/authorship_api_spec.rb
darrenleeweber: Can the JSON body have a string value for the boolean params?  If it can or usually does, should the specs also test it?

spec/api/sul_bib/authorship_api_spec.rb
atz: Yeah, this is more specific (201 is one flavor of success).

spec/api/sul_bib/authorship_api_spec.rb
atz: It moved above to R355 and didn't require any special `request_data` context.

spec/api/sul_bib/authorship_api_spec.rb
atz: I think the failures have various behaviors (404 vs. 401).

spec/api/sul_bib/authorship_api_spec.rb
atz: PATCH requests are by definition only updates.

app/api/sul_bib/authorship_api.rb
atz: The point was to move the contrib.save inside the existing exception trapping block and make it required (add `!`).  I didn't want `contrib.save` to be able to silently fail.

spec/api/sul_bib/authorship_api_spec.rb
atz: This test is only concerned with the downcasing via Grape.  So it is not particularly concerned with response (tested elsewhere).  The POST definition requires values for `:featured` and `status` (also tested elsewhere).

spec/api/sul_bib/authorship_api_spec.rb
atz: I don't think `Contribution.count` is important here.  That is established elsewhere.  The overlap is 2 lines (simple value checks, on different objects), so no real benefit from adding shared_examples for them.

spec/api/sul_bib/authorship_api_spec.rb
atz: Yeah, `!be_nil` is required for the boolean, because false is not "present" but it is not nil.

spec/api/sul_bib/authorship_api_spec.rb
darrenleeweber: OK, consolidate where possible.  Maybe there was excess paranoia there before.

spec/api/sul_bib/authorship_api_spec.rb
darrenleeweber: Ah, right, yea.

spec/api/sul_bib/authorship_api_spec.rb
darrenleeweber: Ah, yes.  OK, good.

PR #516, opened 2017-12-07
Similar refactoring comes to publications_api_spec

atz:
- common `let` statements (that are about their return values, not about establishing a context)
- removed unused `let` statement
- converted another to a `before` block, since the return values weren't being used, it was really about establishing context
- AVOID SERIALIZING/DESERIALING JSON... SERIALLY
- Manipulate hashes, serialize to JSON later
- Corrected some description statements that were lying about the numbers inolved
- Removed a single-use factory that was materially no different than the base factory, but had a name that suggested it did something special.
- harness the power of RSpec `include`

These changes are not as comprehensive as the previous ones on authorship_api_spec, but
they are a great improvement in performance and readability.

### DIFF
--- a/spec/factories/author.rb
+++ b/spec/factories/author.rb
@@ -18,10 +18,6 @@ FactoryBot.define do
     emails_for_harvest 'alice.edler@stanford.edu'
   end
 
-  factory :author_with_sw_pubs, parent: :author do
-    id 33
-  end
-
   factory :inactive_author, parent: :author do
     active_in_cap false
   end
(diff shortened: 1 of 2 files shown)


darrenleeweber:
Resolve conflicts

atz:
Conflict resolved

spec/factories/author.rb
```diff
@@ -18,10 +18,6 @@
     emails_for_harvest 'alice.edler@stanford.edu'
   end
 
-  factory :author_with_sw_pubs, parent: :author do
-    id 33
-  end
-
```
darrenleeweber: :ok_hand:

darrenleeweber:
I don't have a lot of time to review this in detail, with some deadlines looming.  Did some cursory look over it and the patterns are very similar to other refactors on specs.  Should be OK.

PR #517, opened 2017-12-07
Migration adding `wos_uid` field to publications table

### LINKED ISSUE #504: Create a migration to add WosID column to publications table
Could be done as a cached value retrieved from PublicationIdentifiers or denormalized as a column on publications table

connected to #467

atz:
fixes #504


atz:
Code climate should already be configured to ignore `db/` so I don't know how it is useful here.

PR #454, opened 2017-11-21
Cap::Client - use Faraday and VCR specs

darrenleeweber:
Fix #458
Fix #496
Fix #497
Fix #498
Fix #499 

- move CAP-API code into a `Cap` module
- use Faraday for the `Cap::Client`
- should fix some code-smells identified by code-climate
  - fixed issues in the `Cap::Client`

### Dependencies
Connected to #490 

### Notes
- there is no attempt here to fix anything for the existing cap client that will be replaced by this new client; the original PR did all those fixes and replaced the old client, that will now be done in #537

PR #523, opened 2017-12-09
Add unique constraint to Author cap_profile_id

atz:
Belt and suspenders approach here:
- model validation
- database constraint

I verified that there are no duplicate `cap_profile_id` values on
production server, so no data repair/salvage will be necessary:
```ruby
Author.select(:id, :cap_profile_id).group(:cap_profile_id).having("count(*) > 1")
=> []
```

This will help simplify API logic that runs in circles worrying that maybe `cap_profile_id` is not unique enough.

### DIFF
--- a/spec/lib/cap_authors_poller_spec.rb
+++ b/spec/lib/cap_authors_poller_spec.rb
@@ -1,4 +1,3 @@
-
 describe CapAuthorsPoller, :vcr do
   # The author is defined in /spec/factories/author.rb
   let(:author) { create :author }
@@ -82,14 +81,12 @@ describe CapAuthorsPoller, :vcr do
     context 'with an existing author' do
       before do
         expect(Author).to receive(:find_by_cap_profile_id).with(author.cap_profile_id).and_return(author)
-        expect(author).to receive(:'save!').and_call_original
+        expect(author).to receive(:'save!')
       end
 
       it 'updates an existing author' do
         expect(author).to receive(:harvestable?).and_return(true)
-        count = subject.instance_variable_get('@authors_updated_count')
-        subject.process_record(author_record)
-        expect(subject.instance_variable_get('@authors_updated_count')).to eq(count + 1)
+        expect { subject.process_record(author_record) }.to change { subject.instance_variable_get('@authors_updated_count') }.by(1)
       end
 
       it 'adds author.id to the harvest queue when harvesting is enabled' do
@@ -103,9 +100,7 @@ describe CapAuthorsPoller, :vcr do
       it 'skips harvesting for an existing author if not marked harvestable' do
         allow(author).to receive(:changed?).and_return(true)
         expect(author).to receive(:harvestable?).and_return(false)
-        count = subject.instance_variable_get('@no_sw_harvest_count')
-        subject.process_record(author_record)
-        expect(subject.instance_variable_get('@no_sw_harvest_count')).to eq(count + 1)
+        expect { subject.process_record(author_record) }.to change { subject.instance_variable_get('@no_sw_harvest_count') }.by(1)
       end
 
       context 'with an authorship record' do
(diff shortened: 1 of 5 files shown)


darrenleeweber:
This will conflict with the larger work in #454 and I don't want to rebase that work on this smaller PR; it should be easier to rework this after #454 is done (the latter has been open for a while now).

atz:
I don't foresee any significant conflict here.  #454 is getting broken up, so I don't think we need to defer this one for anything.

spec/lib/cap_authors_poller_spec.rb
```diff
@@ -82,14 +81,12 @@
...
+        expect(author).to receive(:'save!')
       end
 
       it 'updates an existing author' do
         expect(author).to receive(:harvestable?).and_return(true)
-        count = subject.instance_variable_get('@authors_updated_count')
-        subject.process_record(author_record)
-        expect(subject.instance_variable_get('@authors_updated_count')).to eq(count + 1)
+        expect { subject.process_record(author_record) }.to change { subject.instance_variable_get('@authors_updated_count') }.by(1)
       end
```
darrenleeweber: :+1: for the revised syntax.

Long lines are difficult to review on github; pls consider breaking them for easier reading/review, e.g. any variant where it can break the line will do and this is just one option:
```ruby
expect { subject.process_record(author_record) }
  .to change { subject.instance_variable_get('@authors_updated_count') }.by(1)
```

spec/lib/cap_authors_poller_spec.rb
```diff
@@ -82,14 +81,12 @@
     context 'with an existing author' do
       before do
         expect(Author).to receive(:find_by_cap_profile_id).with(author.cap_profile_id).and_return(author)
-        expect(author).to receive(:'save!').and_call_original
+        expect(author).to receive(:'save!')
       end
```
darrenleeweber: Looks like this needs to be mocked because the factory has already saved the author.  Perhaps another option is to have the factory `build` the author instead of `create` it?  Then the specs will allow the code to save authors, as before.

spec/lib/cap_authors_poller_spec.rb
```diff
@@ -1,4 +1,3 @@
-
 describe CapAuthorsPoller, :vcr do
   # The author is defined in /spec/factories/author.rb
   let(:author) { create :author }
```
darrenleeweber: Can this be changed to use `build` instead?  The `build` option does not save it.
- https://devhints.io/factory_bot

atz: Maybe!  I added both database and model constraints on uniqueness though, so I think testing "through" the database is fine for the scope of this PR.  I don't think that refactoring of the spec is tied to cap_profile_id uniqueness.

spec/lib/cap_authors_poller_spec.rb
atz: We never tested anything about the author *after* the `save!`, and this is not the test of the `Author` model, so there was no need to hit the DB here.  I made a similar change below at L123.

darrenleeweber: Sorry, perhaps I'm confused about the intent here.  This is the cap authors poller and it seems like these specs are checking that it saves or does not save an author for various scenarios.  Some of the scenarios assume the author already exists and others do not.  AFAIK, when the `save!` is mocked, without calling the original, it does nothing any more.  Is it fair to assume that the new constraints made this fail for some reason (like the cap_profile_id is not unique)?

spec/lib/cap_authors_poller_spec.rb
darrenleeweber: Wondering whether using a `build` would allow the other mocks to `save` and call original.

spec/lib/cap_authors_poller_spec.rb
atz: Yeah, the specs were more loose with faked Author objects than our actual code was.

darrenleeweber:
At standup today we agreed to pursue the unique constraint in this PR
- the cost of this is breaking the cap-profile-id rewriter
  - we resolved to raise an exception there and fix it later, if necessary

There has also been some consideration of requiring the cap-profile-id on the Author
- that could be addressed in a separate PR
- essentially that would a Non-NULL constraint on the field
- the uniq constraint might only allow one NULL anyway, if any (not sure how rails handles that)

PR #524, opened 2017-12-09
Convert bespoke "validations" to ActiveRecord basis

atz:
This extends #523.  Now rebased after that PR was merged.

### DIFF
--- a/app/api/sul_bib/authorship_api.rb
+++ b/app/api/sul_bib/authorship_api.rb
@@ -289,17 +289,13 @@ module SulBib
         authorship_hash[:status].present? ||
         authorship_hash[:visibility].present? ||
         error!("At least one authorship attribute is required: 'featured', 'status', 'visibility'.", 406)
-      unless authorship_hash[:featured].nil?
-        Contribution.featured_valid?(authorship_hash) ||
-          error!("The 'featured' param is invalid: #{authorship_hash[:featured]}.", 406)
-      end
-      if authorship_hash[:status].present?
-        Contribution.status_valid?(authorship_hash) ||
-          error!("The 'status' param is invalid: #{authorship_hash[:status]}.", 406)
-      end
-      if authorship_hash[:visibility].present?
-        Contribution.visibility_valid?(authorship_hash) ||
-          error!("The 'visibility' param is invalid: #{authorship_hash[:visibility]}.", 406)
+
+      prototype = Contribution.new(authorship_hash.slice(:featured, :status, :visibility))
+      prototype.validate # we KNOW it won't validate (w/o author and publication), but we check for the other fields
+      errors = prototype.errors.messages.slice(:featured, :status, :visibility)
+      if errors.present?
+        msg = errors.map { |k, v| "The '#{k}' param is invalid: #{authorship_hash[k]} -- #{v}" }.join("\n")
+        error!(msg, 406)
       end
       create_or_update_and_return_pub_hash(pub, author, authorship_hash)
     end # patch end
--- a/app/models/contribution.rb
+++ b/app/models/contribution.rb
@@ -45,29 +45,15 @@ class Contribution < ActiveRecord::Base
     end
   end
 
+  # checks featured, status and visibility
+  # @return [Boolean]
   def self.valid_fields?(contrib)
-    featured_valid?(contrib) &&
-      status_valid?(contrib) &&
-      visibility_valid?(contrib)
-  end
-
-  # @return [Boolean]
-  def self.visibility_valid?(contrib)
-    contrib = contrib.with_indifferent_access
-    VISIBILITY_VALUES.include? contrib[:visibility].to_s.downcase
-  end
-
-  # @return [Boolean]
-  def self.status_valid?(contrib)
-    contrib = contrib.with_indifferent_access
-    STATUS_VALUES.include? contrib[:status].to_s.downcase
-  end
-
-  # Allowed values for featured are true and false
-  # @return [Boolean]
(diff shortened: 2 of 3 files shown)


darrenleeweber:
Commit ab0471bbb8f4b8286899ff29bc129809a8781516 could be obsolete if we remove that method, now that `Publication.wos_uid` is available.

darrenleeweber:
Commit 2b05198b6e3caffdd49c4ca34fc428ec07dace38 could be a new PR, it looks like an atomic change on the Contribution validation (not clear that it depends on the author validation too).

atz:
`wos_uid` and `for_uid` are different things.  The former in an attribute.  The latter queries `publication_identifiers`, similar to the other methods that do the same for other identifier types.

atz:
Split off a PR for `for_uid`: #557

atz:
rebased

app/api/sul_bib/authorship_api.rb
```diff
@@ -289,17 +289,13 @@ def get_publication(sul_pub_id, pmid, sw_id)
...
-        Contribution.visibility_valid?(authorship_hash) ||
-          error!("The 'visibility' param is invalid: #{authorship_hash[:visibility]}.", 406)
+
+      prototype = Contribution.new(authorship_hash.slice(:featured, :status, :visibility))
+      prototype.validate # we KNOW it won't validate (w/o author and publication), but we check for the other fields
+      errors = prototype.errors.messages.slice(:featured, :status, :visibility)
+      if errors.present?
+        msg = errors.map { |k, v| "The '#{k}' param is invalid: #{authorship_hash[k]} -- #{v}" }.join("\n")
+        error!(msg, 406)
       end
```
darrenleeweber: 👌 - just wondering if the `msg` needs to be in JSON or any old String will do?  If JSON is required, it's probably still valid JSON as a String that contains the new line separators etc., probably OK either way.  In any case, seems consistent with what is there already.

app/models/contribution.rb
```diff
@@ -45,29 +45,15 @@ def self.author_valid?(contrib)
...
-    contrib = contrib.with_indifferent_access
-    contrib[:featured].to_s =~ /true|false/i ? true : false
+  def self.valid_fields?(contrib)
+    segment = contrib.with_indifferent_access.slice(:featured, :status, :visibility)
+    return false unless segment.size == 3
+    return false if segment.values.any?(&:nil?)
+    prototype = Contribution.new(segment)
+    prototype.validate # we KNOW it won't validate (w/o author and publication), but we check for the other fields
+    prototype.errors.messages.slice(:featured, :status, :visibility).empty?
   end
```
darrenleeweber: 👌 - only wondering why all the prelude checks instead of just throwing `contrib` to `Contribution.new(contrib)`.  As our own code is likely the only consumer of this method, it's probably going to give it the right stuff.  The only other thought is what happens to an API PATCH that only needs to validate one field?

spec/models/contribution_spec.rb
```diff
@@ -95,71 +95,23 @@
...
-      expect(described_class.valid_fields?(authorship)).to be false
-    end
-    it 'returns false for an authorship hash without "status"' do
-      authorship['status'] = nil
-      expect(described_class.valid_fields?(authorship)).to be false
-    end
-    it 'returns false for an authorship hash without "visibility"' do
-      authorship['visibility'] = nil
-      expect(described_class.valid_fields?(authorship)).to be false
-    end
```
darrenleeweber: Why do these specs get removed?  It looks like they test the desired functionality, although the implementation details have changed.

darrenleeweber: NM, may delete that comment ^^ when I see more details below.

spec/models/contribution_spec.rb
```diff
@@ -95,71 +95,23 @@
...
+      end
+      it 'returns false without "status"' do
+        authorship['status'] = nil
+        expect(described_class.valid_fields?(authorship)).to be false
+      end
+      it 'returns false without "visibility"' do
+        authorship['visibility'] = nil
+        expect(described_class.valid_fields?(authorship)).to be false
+      end
     end
```
darrenleeweber: What if `visibility` etc. are an empty String?  IIRC, all these attributes are validated against enums now, so is it safe to assume that an empty String is false?  Why is `featured` ommitted from the first `bad values` block?

app/api/sul_bib/authorship_api.rb
atz: Yeah, it hasn't changed in this PR.  I don't think it matters for a `406` status anyway.

app/models/contribution.rb
atz: I agree that is the goal.  This moves us in that direction.  I implement more in that direction in other PRs.

spec/models/contribution_spec.rb
atz: Right, empty string is not an enumerated value, so it won't validate via AR. 

`featured` defaults to `false` when nil (reasonably enough), so at the model level, it is not actually required.  Some parts of the API require it, but not the model.

app/api/sul_bib/authorship_api.rb
darrenleeweber: :ok_hand:

app/models/contribution.rb
darrenleeweber: :ok_hand:

spec/models/contribution_spec.rb
darrenleeweber: :ok_hand:

PR #539, opened 2017-12-12
Remove unused method, soon to become even more unused

atz:
`git grep update_from_sciencewire` hits only the lines removed.

### DIFF
--- a/app/models/publication.rb
+++ b/app/models/publication.rb
@@ -148,14 +148,6 @@ class Publication < ActiveRecord::Base
     rebuild_pub_hash
   end
 
-  def update_from_sciencewire
-    return false if sciencewire_id.blank?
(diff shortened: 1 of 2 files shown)


darrenleeweber:
```
$ git grep 'update_from_sciencewire'
app/models/publication.rb:151:  def update_from_sciencewire
spec/models/publication_spec.rb:301:  describe 'update_from_sciencewire' do
spec/models/publication_spec.rb:304:      expect(publication.update_from_sciencewire).to be false
spec/models/publication_spec.rb:322:      expect(publication.update_from_sciencewire).to be true
```

peetucket:
Was in the wiki under useful commands but I’ve not run enough to warrant keeping. Ok to kill

PR #541, opened 2017-12-12
SimpleCov - maximum coverage drop set to 1%

darrenleeweber:
Code climate relies on SimpleCov for code coverage reports.  Adjusting this setting could avoid some breaking builds when there are minor changes in coverage (often due to refactoring legacy code).


darrenleeweber:
Well, this didn't change how code-climate is checking the total coverage.  Filed an issue with code climate:
- https://github.com/codeclimate/test-reporter/issues/271

PR #544, opened 2017-12-12
WebOfScience::ProcessRecords - revise specs to use Publication.wos_uid

darrenleeweber:


spec/lib/web_of_science/process_records_spec.rb
```diff
@@ -74,10 +74,10 @@
       expect { processor.execute }.to change { Contribution.count }
     end
     it 'creates new contribution in the pub_hash[:authorship]' do
-      uids = processor.execute
-      uids.each do |uid|
-        pub = Publication.for_uid(uid)
-        expect(pub.pub_hash).to include(:authorship)
+      processor.execute
```
atz: why not just do `processor.execute.each` ?

atz: Ah, I see, you are just using the `uid`s from the fixture files.

darrenleeweber: This uses the original set of records, whereas the return from the processing could be a filtered set of results.  In this spec, all of the records should create new pubs, so it's got to check that they all exist etc.

PR #550, opened 2017-12-13
WebOfScienceSourceRecord#source_data converted to mediumtext length

### LINKED ISSUE #547: WOS Harvest bug - source record too long
e.g. for `WOS:000285683500018`
```
E, [2017-12-12T13:26:06.127226 #7394] ERROR -- : #<ActiveRecord::StatementInvalid: Mysql2::Error: Data too long for column 'source_data' at row 1: INSERT INTO `web_of_science_source_records` (
`source_data`, `active`, `uid`, `database`, `source_fingerprint`, `created_at`, `updated_at`) VALUES ('<REC r_id_disclaimer=\"ResearcherID data provided by Clarivate Analytics\"><UID>WOS:00028
5683500018</UID>
```

This likely requires a db-migration to increase the length of the field.  It's not easy to determine a data-dependent record length, the solution may need to use the maximum possible length.  This SO post is helpful
- https://stackoverflow.com/questions/12559762/is-there-a-limit-for-number-of-characters-in-active-record-mysql-db-text-field
- the `:mediumtext` or the `:longtext` options might work

Additional examples:
- `WOS:000302860700005`
- `WOS:000181695400003`
- `WOS:000373133100001`
  - this last one seems to be the largest one to fail

To fetch any individual publication, e.g.:
```
bundle exec rake wos:publication['WOS:000373133100001'] > rec.xml
# edit the XML file to remove cruft and count the lines and chars, e.g.
wc -lc rec.xml 
# =>  51688 2446522 rec.xml
```
If  `WOS:000373133100001` is the largest one to fail, the record length needs to acommodate that one.  It is about 2,446,522 chars, so rounding up a bit gets to about 2.5 million chars.  The `:mediumtext` option is 16,777,215 chars (bytes); that should be big enough.

### source record schema
- https://github.com/sul-dlss/sul_pub/blob/master/db/schema.rb#L212
```ruby
create_table "web_of_science_source_records", force: :cascade do |t|
    t.boolean  "active"
    t.string   "database",           limit: 255
    t.text     "source_data",        limit: 65535
    t.string   "source_fingerprint", limit: 255
    t.string   "uid",                limit: 255
    t.datetime "created_at",                       null: false
    t.datetime "updated_at",                       null: false
    t.string   "doi",                limit: 255
    t.string   "pmid",               limit: 255
  end
```

darrenleeweber:
Fix #547

Note that the source data is also assigned to `Publication#xml`, which is already a `:mediumtext` field
- https://github.com/sul-dlss/sul_pub/blob/master/db/schema.rb#L138
- to fix #547, only the WOS source record needs to be extended

The WOS harvest failed to save some long records.  This PR expands the source record capacity to handle large WOS records.  A review of the harvest logs identified some large records that failed to save; the largest of those records should now fit in this expanded field.  (The logs had some db ROLLBACK errors that were all related to this.)

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20171208233331) do
+ActiveRecord::Schema.define(version: 20171213201429) do
 
   create_table "author_identities", force: :cascade do |t|
     t.integer  "author_id",     limit: 4,               null: false
@@ -212,11 +212,11 @@ ActiveRecord::Schema.define(version: 20171208233331) do
   create_table "web_of_science_source_records", force: :cascade do |t|
     t.boolean  "active"
     t.string   "database",           limit: 255
-    t.text     "source_data",        limit: 65535
+    t.text     "source_data",        limit: 16777215
     t.string   "source_fingerprint", limit: 255
     t.string   "uid",                limit: 255
-    t.datetime "created_at",                       null: false
-    t.datetime "updated_at",                       null: false
+    t.datetime "created_at",                          null: false
+    t.datetime "updated_at",                          null: false
     t.string   "doi",                limit: 255
     t.string   "pmid",               limit: 255
   end
--- a/db/migrate/20171213201429_wos_src_record_medium_text.rb
+++ b/db/migrate/20171213201429_wos_src_record_medium_text.rb
@@ -1,0 +1,9 @@
+class WosSrcRecordMediumText < ActiveRecord::Migration
+  def up
+    change_column :web_of_science_source_records, :source_data, :mediumtext
+  end
+
+  def down
+    change_column :web_of_science_source_records, :source_data, :text
+  end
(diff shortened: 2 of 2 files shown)


db/schema.rb
```diff
@@ -212,11 +212,11 @@
   create_table "web_of_science_source_records", force: :cascade do |t|
     t.boolean  "active"
     t.string   "database",           limit: 255
-    t.text     "source_data",        limit: 65535
+    t.text     "source_data",        limit: 16777215
     t.string   "source_fingerprint", limit: 255
     t.string   "uid",                limit: 255
-    t.datetime "created_at",                       null: false
```
atz: These 2 lines shouldn't have been altered, right?  Or am I missing something?

darrenleeweber: I also looked carefully at that.  It's the auto-format.  It's padding out the `null` column now because now the `limit` column is wider.  It's a no-op, just a format change.  If we don't accept it here, it will just keep happening every time rails auto-formats this file anyway.

PR #556, opened 2017-12-14
Validate cap_profile_id presence, too

atz:
All author records on prod indeed have `cap_profile_id`.

I don't know if we are going to care about this in this work cycle, but I already had this commit on a different branch, so I cherry-picked it here.

With the rewriter locked down, I think we could merge this without difficulty.

### DIFF
--- a/app/models/author.rb
+++ b/app/models/author.rb
@@ -1,6 +1,6 @@
 class Author < ActiveRecord::Base
   has_paper_trail on: [:destroy]
-  validates :cap_profile_id, uniqueness: true
+  validates :cap_profile_id, uniqueness: true, presence: true
 
   has_many :author_identities, dependent: :destroy
   #
--- a/spec/api/sul_bib/authorship_api_spec.rb
+++ b/spec/api/sul_bib/authorship_api_spec.rb
@@ -315,24 +315,6 @@ describe SulBib::API, :vcr do
     end
   end # shared_examples 'it issues errors for cap_profile_id'
 
-  # When the database constraints for cap_profile_id are setup
-  # correctly and the AuthorshipAPI uses cap_profile_id before sul_author_id
-  # to find an author, it's virtually impossible to hit the errors tested
-  # by these specs.  However, the database constraints do not, yet,
-  # enforce NOT NULL and UNIQUE on the cap_profile_id.
-  shared_examples 'it checks author for a cap_profile_id' do
-    let(:request_data) do
-      author.cap_profile_id = ''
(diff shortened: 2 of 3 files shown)


spec/api/sul_bib/authorship_api_spec.rb
```diff
@@ -491,7 +472,6 @@ def check_response_error(code)
       it_behaves_like 'it issues errors when sul_author_id does not exist'
       it_behaves_like 'it issues errors for cap_profile_id'
       it_behaves_like 'it checks author for the correct cap_profile_id'
-      it_behaves_like 'it checks author for a cap_profile_id'
       it_behaves_like 'it handles invalid authorship attributes'
```
darrenleeweber: It's possible these specs are still testing useful behavior for the API, although the implementation has changed.  Something like 'it validates cap_profile_id' and other such behavior is still relevant at the API level?  Did the new validation break these specs?  Could they be adjusted to continue testing the API level?  Am I missing something here - is this behavior still tested somewhere?  Is there any new behavior introduced - does it now trigger any new error responses (either deliberately or through rails/grape magic)?

darrenleeweber:
Generally 👍 after the standup discussion today.  Hesitating on the changes to the API specs.

app/models/author.rb
```diff
@@ -1,6 +1,6 @@
 class Author < ActiveRecord::Base
   has_paper_trail on: [:destroy]
-  validates :cap_profile_id, uniqueness: true
+  validates :cap_profile_id, uniqueness: true, presence: true
```
darrenleeweber: - curious about a corresponding db-migration for `NOT NULL` on that field.

atz: I don't care about that right now.  It doesn't have the same value as DB uniqueness constraint, because presence can be validated by the model without touching the DB.  (There is also no race condition for presence.)

darrenleeweber: :ok_hand:

spec/api/sul_bib/authorship_api_spec.rb
atz: Please remember that I am following up with larger changes to the API.  These tests are excluded because the **setup** is a scenario that is now impossible:
```
author.cap_profile_id = ''		
author.save!
```
The whole idea here is that the model guarantees a `cap_profile_id`.

darrenleeweber: 👌 - I see, and other specs check for errors related to cap-profile-id.

PR #557, opened 2017-12-14
for_uid: use one DB query instead of two

atz:
Obvious.

### DIFF
--- a/app/models/publication.rb
+++ b/app/models/publication.rb
@@ -70,14 +70,12 @@ class Publication < ActiveRecord::Base
                .find_by("publication_identifiers.identifier_type": 'pmid', "publication_identifiers.identifier_value": pmid)
   end
 
-  # Publication for a WOS-UID
-  # @param [String] a WOS-UID
+  # Queries publication_identifiers (like the poorly-named find_by_xxx methods above)
+  # @param [String] a WOS-UID, including database prefix
   # @return [Publication, nil]
   def self.for_uid(uid)
-    # Until the Publication model supports a `wos_uid` field, get the pub via the identifiers
-    pub_ids = PublicationIdentifier.where(identifier_type: 'WosUID', identifier_value: uid)
-    return if pub_ids.empty?
-    pub_ids.first.publication
+    Publication.includes(:publication_identifiers)
+               .find_by("publication_identifiers.identifier_type": 'WosUID', "publication_identifiers.identifier_value": uid)
   end
 
   # @return [Publication] new object, unsaved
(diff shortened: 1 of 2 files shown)


app/models/publication.rb
```diff
@@ -70,14 +70,12 @@ def self.find_by_pmid_pub_id(pmid)
...
+  # @param [String] a WOS-UID, including database prefix
   # @return [Publication, nil]
   def self.for_uid(uid)
-    # Until the Publication model supports a `wos_uid` field, get the pub via the identifiers
-    pub_ids = PublicationIdentifier.where(identifier_type: 'WosUID', identifier_value: uid)
-    return if pub_ids.empty?
-    pub_ids.first.publication
+    Publication.includes(:publication_identifiers)
+               .find_by("publication_identifiers.identifier_type": 'WosUID', "publication_identifiers.identifier_value": uid)
   end
```
darrenleeweber: - For some reason, there is no spec coverage on this now?  I thought this was covered already, weird.
  - or did some hiccup occur in the code-coverage reporting for the travis build(s)?
  - would be good to be confident that the return results are all the same `[Publication, nil]`
- Although I agree with the new comment, but without the smirky bit in parens, i.e. consider
  - `Queries publication_identifiers by "WosUID"` looks good
  - I'm on the fence about using `publication_identifiers` or `PublicationIdentifier`, same diff I guess

atz: Codecliemate is being generally useless, blocking on a 0% diff.  This is a [`find_by`](https://apidock.com/rails/v4.2.7/ActiveRecord/FinderMethods/find_by) so its return signature should be easily understood.

darrenleeweber: I checked out this branch and ran the specs locally to check the simple-cov report. This method is not covered any more and it should be covered.  I ran
```
bundle exec rspec spec/models/publication_spec.rb
# then open coverage/index.html and look at the app/models/publication coverage
```

darrenleeweber: Now I recall why it was tested before - the specs on WOS process records was using that method and now they do not.  Maybe that's why "coverage" disappeared?  I thought I added a spec for that method when it was first created - checking history....

darrenleeweber: This could be a good time to add a spec for it, something like
- https://github.com/sul-dlss/sul_pub/blob/master/spec/models/publication_spec.rb#L368-L380

atz: Yeah, this was a timing issue.  It was covered when I wrote it.

atz: Added a spec for it.

PR #551, opened 2017-12-13
WebOfScience::ProcessRecords - separate pub creation from pubmed additions

darrenleeweber:
Fix #549 

### Rationale
- first create publication and contribution records
- then enhance select records with PubMed data
- isolate the Publication creation from PubMed failures
  - provide better exception log messages using NotificationManager
- isolate WOS links update failures to individual records
  - add record-specific log messages (could help identify problematic records)

### Extras
- fixed a typo in a spec for "creates new contribution in the pub_hash[:authorship]"

PR #558, opened 2017-12-15
WebOfScienceSourceRecord - set identifiers, if not already set

darrenleeweber:
Extracted from #551 to focus on this change alone.

Update: rebased on #559 (converted PMID to Integer) and fixed spec

### DIFF
--- a/spec/models/web_of_science_source_record_spec.rb
+++ b/spec/models/web_of_science_source_record_spec.rb
@@ -26,12 +26,32 @@ RSpec.describe WebOfScienceSourceRecord, type: :model do
     end
   end
 
+  context 'sets identifiers' do
+    before do
+      identifiers = WebOfScience::Identifiers.new(wos_record)
+      allow(identifiers).to receive(:doi).and_return('doi')
+      allow(identifiers).to receive(:pmid).and_return('123')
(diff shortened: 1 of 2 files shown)


spec/models/web_of_science_source_record_spec.rb
```diff
@@ -26,6 +26,26 @@
...
+      allow(WebOfScience::Identifiers).to receive(:new).and_return(identifiers)
+    end
+    it 'sets doi' do
+      expect(wos_src_rec.doi).to eq 'doi'
+    end
+    it 'sets pmid' do
+      expect(wos_src_rec.pmid).to eq 123
+    end
+    it 'allows select' do
+      expect(wos_src_rec.save!).to be true
```
atz: Not going to block over it, but since we aren't intending to test the return of `.save!` here, the following lines are equivalent:
```ruby
expect(wos_src_rec.save!).to be true
wos_src_rec.save!
```
In either case the exclamation point is the, err... point.  The test is invalid without the successful save.

darrenleeweber: Thanks, sorry I missed that.  I want to fix all the places where that happens.

PR #565, opened 2017-12-16
Xtend authorship api for wos

### LINKED ISSUE #408: Implement manual adding of publications by WoS UID
Users are able to search for publications by PMID, Title and DOI via the API and add these to their profile (in cases where these results were not returned via the name query).  The new WoS harvester class has a method for `process_pmids` that needs to be implemented.

calling class is the publications_controller API in the sourcelookup method

- [ ] DoiSearch class

atz:
The only open question I might have is about relative priority the various backends when multiple pub ID types are received.  I atomized `get_publication` into parts because it was getting unwieldy, and the logic for WoS required an additional argument (`author`).  The priority order should also be more obvious, anticipating we might want to tweak it.

fixes #408

### DIFF
--- a/app/api/sul_bib/authorship_api.rb
+++ b/app/api/sul_bib/authorship_api.rb
@@ -23,7 +23,8 @@ module SulBib
         # Publication.pub_hash field, breaking the elegant design of a RDBMS.
         contrib = pub.contributions.find_or_initialize_by(author_id: author.id)
         contrib.assign_attributes(authorship.merge(cap_profile_id: author.cap_profile_id, author_id: author.id))
-        pub.pubhash_needs_update! if contrib.persisted?
+        # error!('You have not supplied a valid authorship record.', 406)
+        pub.pubhash_needs_update! if contrib.persisted? && contrib.changed?
         begin
           contrib.save!
           # Save the publication so it will sync the contribution into the pub.pub_hash[:authorship] array.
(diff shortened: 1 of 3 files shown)


app/api/sul_bib/authorship_api.rb
```diff
@@ -201,34 +158,38 @@ def get_publication(sul_pub_id, pmid, sw_id)
...
-      pub = get_publication(
-        params[:sul_pub_id],
-        params[:pmid],
-        params[:sw_id]
-      )
+      # may be fetched from PubMed (pmid), WebOfScience (wos_uid) or ScienceWire (sw_id).
+      pub = get_local_publication!(ids[:sul_pub_id]) if ids[:sul_pub_id]
+      pub ||= get_publication_via_pubmed!(ids[:pmid]) if ids[:pmid]
+      pub ||= get_publication_via_sciencewire!(ids[:sw_id]) if ids[:sw_id]
+      pub ||= get_publication_via_wos!(author, ids[:wos_uid]) if ids[:wos_uid]
```
darrenleeweber: Maybe each of the `get_publication_*` could use some exception handling that would allow this block to move on to the next option?  Even if it's not our preferred source, if we can get it - good.  It looks like one source throwing an exception could break the chain here.

darrenleeweber:
Looked over the API code changes - seems like some good refactoring there
Need to review the specs - when I'm not so brain dead as now

app/api/sul_bib/authorship_api.rb
atz: It's not exception handling, we intentionally respond w/ `404` on the first id that doesn't work out.  I think this behavior is correct, given that the API is expecting a **known item** to be referenced, something that was *already* retrieved previously.  Also given that performance will degrade if we are serially searching N remote backends instead one.

The possibility of receiving multiple IDs of different types is an artifact of each param being optional, not something that had intentional use.  So it was barely unexplored by the existing implementation (only via the priority order of the backends).  No rectification was performed (e.g., if we receive 2 ids, does the retrieved object match both of them?), and I wouldn't really want to start that now.

It probably would have been better to use a single required `uid` param w/ namespace.  The cost is that it would have required some param validation and interpretation, but this question would have been effectively excluded.  Also would have been easier to extend for new DBs, since we wouldn't have to add a new param.

darrenleeweber:
Looking at the VCR cassettes recorded for this API, using `tree fixtures/vcr_cassettes/SulBib_API/` (if the `tree` utility is not available, maybe `brew install tree`?).
- It looks like some cassettes exist for the WOS data, maybe they came from a prior commit before the mocks were done to use a fixture file?
- there are still some SW retrievals/cassettes in place, so that's consistent with the scope of this PR, I think; that is, this PR is not expected to disable any SW-API calls.  Also, this PR is not expected to place the WOS-API calls in a higher precedence than SW-API calls.

atz:
Squashed your commit into the previous.

app/api/sul_bib/authorship_api.rb
darrenleeweber: Hmmm, yea, not sure about scope here.  On the one hand, this API should be mostly used, if not exclusively, for associating or updating authorship for an existing pub.  BUT, that's not entirely true.  It _does_ allow harvests too.  It's the kitchen sink of "hey, show me the pub that I own" (given one or more pub ids).  I'm not entirely clear about the assumption that this API is only used with a known pub-id that was retrieved from us before it came back to us.

darrenleeweber: I think this amounts to exactly where the `log_and_error` is called and whether it should wrap the entire block of options to find/harvest a pub or if it should be in each one of them.  As it is now, it's in each one of them and any one of them will break the chain.

darrenleeweber: See https://github.com/sul-dlss/sul_pub/commit/9decc0fc96903f95b76ba38c49fa873750bebcd0

atz: You misinterpret me.  The item **known**, that doesn't mean it is local.  Title search is not accomplished via authorship API.  

Your commit obscures a range of possible different errors with the same message.

PR #572, opened 2017-12-19
Xtend authorship api for wos - with mocks

darrenleeweber:
This mocks the record request, but it still needs to mock the WOS-API authentication.


atz:
Cherry-picked the commit b6dc128

PR #578, opened 2017-12-20
authorship_api - revised exception handling

darrenleeweber:
Follow-up changes to #565 
- allow the authorship to find or retrieve a publication from anywhere possible, given any of the params provided
- don't crash the chain of retrieval if any one of the steps fails

### DIFF
--- a/app/api/sul_bib/authorship_api.rb
+++ b/app/api/sul_bib/authorship_api.rb
@@ -87,36 +87,31 @@ module SulBib
 
       # Find an existing SUL publication or, if it doesn't exist, it may be fetched based on ID provided
       # @param [String] sul_pub_id internal ID
-      # @return [Publication]
+      # @return [Publication, nil]
       def get_local_publication!(sul_pub_id)
         Publication.find(sul_pub_id)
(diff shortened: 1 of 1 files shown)


atz:
I'm not following your motivation and I think you don't understand what is happening here.  If there is an important use case that needs to be preserved, then a test would be the right way to express that.  

`error!` is a method from Grape::API.  It responds to the client. That is "working", not "failing".  

We should NOT attempt to exhaustively query by all possible params, as I documented in other PR conversations and discussed in person.  That existing behavior of the old API does not do that. 
 It is not a supported use case.   

Why?  Because performance degrades and it is non-deterministic what you get back as a result.  The same request causing a PubMed record to be fetched and associated vs. a WoS record if PubMed happens to hiccup is **worse** than just returning an error message.  This is not **searching**.  The client is expected to be identifying **exactly one** record, no matter how many identifiers are received. 

For example, we don't want to go fetching a PubMed record by `pmid` as a failover if they gave us a bad `sul_pub_id` and `Publication.find_by` failed.  Because there is obviously something wrong with the request that they think they have an identifier (in this case, for one of our records), but we don't know what they are talking about.

PR #583, opened 2017-12-23
[WIP] WebOfScience::Harvester - extract WebOfScience::ProcessAuthor

### LINKED ISSUE #582: WOS query so it only returns record-UID values, not full records
How can we constrain a WOS-API query so that it only returns matching WOS-UID values, not the entire full record, in the results?  I can’t find anywhere in the documentation that describes such option.  Maybe it doesn’t exist and we always get back the entire full record for all the matching results.  I’m looking for modifier that turns off the full records but leaves the record-UIDs on in this section:
```
<retrieveParameters>
  <firstRecord>1</firstRecord>
  <count>2</count>
  <option>
    <key>RecordIDs</key>
    <value>On</value>
  </option>
</retrieveParameters>
```
The rationale for excluding the full-records in the results is that we can filter the results based only on the record-UID to exclude further processing for Publication records that already exist.  This could help with memory management for the harvesting process, by excluding extraneous content (that has been harvested before).

darrenleeweber:
Fix #582 - for more efficient WOS harvests by author name-institution.

- find all the WOS-UIDs for an author by the name-institution search
  - only retrieve the UIDs, not the entire source record
- then re-route the harvest through the process_uids method
  - find or create contributions to any existing publications
  - retrieve and process only the new WOS-UID without contributions
  - use the retrieve_by_uid query on the WOS-API to get full records

Ran this branch on my laptop using a prod-db dump and it seemed to work OK, using a limited sample of 100 authors.  Not sure what the requirements need to be for an integration test to pass.  These counts indicate that it created new records.  We should assume that most of these WOS-records have already been harvested, so it's not surprising that this Publication count is _much_ lower than the source record count.
```
PRODUCTION [1] pry(main)> WebOfScienceSourceRecord.count
=> 4443
PRODUCTION [2] pry(main)> Publication.pluck(:wos_uid).compact.count
=> 771
```

### DIFF
--- a/lib/web_of_science/harvester.rb
+++ b/lib/web_of_science/harvester.rb
@@ -29,7 +29,7 @@ module WebOfScience
       # TODO: iterate on author identities also, or leave that to the consumer of this class?
       raise(ArgumentError, 'author must be an Author') unless author.is_a? Author
       logger.info "#{self.class} - processing author: #{author.id}"
-      uids = author_record_uids(author)
+      uids = WebOfScience::ProcessAuthor.new(author).uids
       uids = process_uids(author, uids)
       logger.info "#{self.class} - processed author: #{author.id}"
       uids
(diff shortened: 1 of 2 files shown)


lib/web_of_science/harvester.rb
```diff
@@ -52,75 +58,72 @@ def process_pmids(author, pmids)
...
       # @param author [Author]
-      # @return [Agent::AuthorName]
-      def author_name(author)
-        Agent::AuthorName.new(
-          author.last_name,
-          author.first_name,
-          Settings.HARVESTER.USE_MIDDLE_NAME ? author.middle_name : ''
+      # @param contrib [Contribution]
+      # @return [Array<String>]
+      def author_contribution_saved?(author, contrib)
```
atz: This is a hazardous conflation of a boolean-like interrogative method and a an actual edit/`.save!`.  The documented return value appears to be incorrect.

darrenleeweber: OK, changed.

darrenleeweber:
Everything has been split out of this PR, this can be closed now.

PR #594, opened 2018-01-08
WebOfScience::Queries - surface query messages in the public API

darrenleeweber:
Extracting this from #583 to focus only on the query changes in this PR.

### Rationale
- push the query message details into the public API
  - allow clients to inspect and modify the query message details
  - this becomes important in the harvester - to retrieve only WOS-UIDs for queries, not full records
- provide new options to query for record fields (or no fields at all)
  - prior to this change, it was only possible to retrieve full records
  - retrieving full records is too expensive when they have already been harvested
  - the `viewField` is described in the WOS API doc on p. 37 and it can be used to retrieve only the UID

### Changes
- changes the `search` method signature, which might break existing clients
  - need to check specs and all calls to any methods that change
- adds utility method to return empty records
  - all the public APIs should use it when there's nothing to return

Checking for use of `WebOfScience::Queries#search`, using:
- `git grep 'WebOfScience.queries'`
- `git grep 'wos_queries'`

Most of the calls on `WebOfScience::Queries` do not use the generic `search` method, they use more specific methods.  Only the `WebOfScience::Harvester` needs to be changed (done).

### DIFF
--- a/spec/lib/web_of_science/queries_spec.rb
+++ b/spec/lib/web_of_science/queries_spec.rb
@@ -76,7 +76,8 @@ describe WebOfScience::Queries do
   describe '#search with valid query' do
     let(:title) { 'A research roadmap for next-generation sequencing informatics' }
     let(:query) { "TI=#{title}" }
-    let(:records) { wos_queries.search(query) }
+    let(:params) { wos_queries.params_for_search(query) }
+    let(:records) { wos_queries.search(params) }
 
     before do
       savon.expects(:authenticate).returns(wos_auth_response)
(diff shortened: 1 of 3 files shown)


spec/lib/web_of_science/queries_spec.rb
```diff
@@ -140,6 +142,37 @@
     end
   end
 
+  shared_examples 'search_params' do
+    let(:query) { params[:queryParameters] }
```
atz: no need for these `let` statements here.  In the examples below, just do:
```ruby
expect(params[:queryParameters]).to include ...
expect(params[: retrieveParameters]).to include ...
```
Or if you want to be paranoid/fancy, you can combine all 3 of these into one expectation:
```ruby
expect(params).to match a_hash_including(
  query_parameters: a_hash_including(databaseId: String, userQuery: String, timeSpan: Hash, queryLanguage: String),
  retrieveParameters: a_hash_including(firstRecord: Integer, count: Integer, option: Array)
)
```
I understand if you think that the rspec output might be more intimidating, but the point is that the let statements aren't giving us anything.

darrenleeweber: This habit of extracting a field helps to keep line lengths shorter and that keeps rubocop happy.  As this PR is already approved and this is a style issue, I'm going to take this advice for future work but skip it here in this PR (I want to move forward with more work on harvesting and want to avoid additional git conflicts or anything that might come up from changing this).

PR #595, opened 2018-01-08
raise from rebuild_pub_hash if the pub is a WoS record

atz:


darrenleeweber:
This was discussed as an interim change to protect WOS publications from any destructive changes that could be created by this method.  (Would be good if we had the issue linked to this PR too.)

PR #596, opened 2018-01-08
WebOfScience::Harvester - use WebOfScience.queries singleton

darrenleeweber:
Split out of #594


atz:
This is an argument to get rid of `wos_queries` as a method altogether, but that can be deferred for a followup.

PR #598, opened 2018-01-08
WebOfScience::Client - log to a different log file for routine client logs

darrenleeweber:
Split out of #583 

This is motivated by debugging the WOS harvest.  The WOS savon client logs contain a lot of repetitive noise.  Splitting out that logging to a separate file makes it easier to see actual WOS harvest log content without all the noise.

### DIFF
--- a/lib/web_of_science/client.rb
+++ b/lib/web_of_science/client.rb
@@ -13,8 +13,6 @@ module WebOfScience
     AUTH_WSDL = 'http://search.webofknowledge.com/esti/wokmws/ws/WOKMWSAuthenticate?wsdl'.freeze
     SEARCH_WSDL = 'http://search.webofknowledge.com/esti/wokmws/ws/WokSearch?wsdl'.freeze
 
-    delegate :logger, to: :WebOfScience
-
     def initialize(auth_code, log_level = Settings.WOS.LOG_LEVEL.to_sym)
       @auth_code = auth_code
       @log_level = log_level
@@ -77,6 +75,10 @@ module WebOfScience
       @auth = nil
       @search = nil
       @session_id = nil
+    end
+
+    def logger
+      @logger ||= Logger.new('log/web_of_science_client.log')
     end
   end
 end
--- a/spec/lib/web_of_science/client_spec.rb
+++ b/spec/lib/web_of_science/client_spec.rb
@@ -16,7 +16,7 @@ describe WebOfScience::Client do
 
   before do
     null_logger = Logger.new('/dev/null')
-    allow(WebOfScience).to receive(:logger).and_return(null_logger)
+    allow(wos_client).to receive(:logger).and_return(null_logger)
   end
 
   describe '#new' do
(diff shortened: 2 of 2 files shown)


lib/web_of_science/client.rb
```diff
@@ -13,9 +13,7 @@ class Client
     AUTH_WSDL = 'http://search.webofknowledge.com/esti/wokmws/ws/WOKMWSAuthenticate?wsdl'.freeze
     SEARCH_WSDL = 'http://search.webofknowledge.com/esti/wokmws/ws/WokSearch?wsdl'.freeze
 
-    delegate :logger, to: :WebOfScience
-
-    def initialize(auth_code, log_level = Settings.WOS.LOG_LEVEL.to_sym)
```
atz: For the changes you want, I don't see how it make sense to cut out the `Settings`.  Seems like you should just be overriding the setting to accomplish your debugging, not making it irrelevant, like:
```ruby
$ SETTINGS__WOS__LOG_LEVEL=error bundle exec rails c
Loading development environment (Rails 4.2.10)
[1] pry(main)> Settings.WOS.LOG_LEVEL
=> "error"
...
$ SETTINGS__WOS__LOG_LEVEL=warn bundle exec rails c
Loading development environment (Rails 4.2.10)
[1] pry(main)> Settings.WOS.LOG_LEVEL
=> "warn"
```

darrenleeweber: That's an interesting idea, thanks.  But, I think this change will be useful in general.  All the savon log info is really not informative in the main harvesting logs.  If you get a chance to look at the logs without this change, you can see why this change is generally useful.  At the level of the savon client, there's not much app-specific logging that we need an app-specific setting for it.  But, if we really should keep the log level, that could stay, so long as the client log file is separated out as in this PR.

darrenleeweber: OK, let's keep it.  Pushed.

atz:
If I understand the problem correctly, it is to reduce the noise from Savon logging.  That can already be accomplished directly by changing the log level in Settings or overriding via ENV.  You still get the same (default) log level from the delegated `WebOfScience.logger` as you get with your new one.  So if you set the Savon level to `:error`, I'm assuming there is no Savon noise left.

In general, Savon uses the `logging` gem, which can be finely tailored *per class* in the style of `log4j` if we need even greater control.

darrenleeweber:
@atz - huh?  I put back the setting already.  The use of a different file is important.  The savon logs are full of noise and it's useful to isolate those logs from _everything_ else that is using the WOS log file.  The client interactions with their API deserve their own log file for easily finding information about only those interactions.  Everything else in our WOS log file is under our control and should contain more app-specific information.

atz:
Right, you put it back but still seem unwilling to actually **use it**.  By raising the level of the Savon logger arbitrarily high, your debugging goal is already accomplished.  The only thing left in the logs would be what you want, which doesn't need a new file location.

darrenleeweber:
@atz - "So if you set the Savon level to :error, I'm assuming there is no Savon noise left."  Ah, I see what your getting at.  Sorry, I seem to recall that I already tried that at some point and still found too much noise in the logs.  Also, I then decided that I don't want to kill off all the info for normal use, but I do want a separate log file to capture what will be mostly noise but might contain some useful info that we can grep.  I'm particularly interested in using the log file with enough information to grep it when we run into throttling issues (and count the number of requests per period etc. from the log file).  I hope that clarifies the intent of using a separate log file and keeping around some of the "noise" but separating it from more important app-specific log content.

PR #599, opened 2018-01-08
WebOfScience::Harvester - log progress for author harvests

darrenleeweber:
Extracted from #583 

This provides more useful log information when debugging the WOS harvester for authors.

### DIFF
--- a/lib/web_of_science/harvester.rb
+++ b/lib/web_of_science/harvester.rb
@@ -10,13 +10,16 @@ module WebOfScience
     # @param [Enumerable<Author>] authors
     # @return [void]
     def harvest(authors)
+      logger.info "#{self.class} - started harvest(authors) - #{authors.count} authors"
+      author_success = 0
       authors.each do |author|
-        begin
-          process_author(author)
-        rescue StandardError => e
-          NotificationManager.error(e, "WebOfScience harvest failed for author: '#{author.id}'", self)
-        end
+        process_author(author)
+        author_success += 1
       end
+      logger.info "#{self.class} - completed harvest(authors) - #{author_success} processed"
+    rescue StandardError => err
+      message = "#{self.class} - harvest(authors) failed - #{author_success} processed"
+      NotificationManager.error(err, message, self)
     end
 
     # Harvest all publications for an author
@@ -24,7 +27,13 @@ module WebOfScience
     # @return [Array<String>] WosUIDs that create Publications
     def process_author(author)
       raise(ArgumentError, 'author must be an Author') unless author.is_a? Author
-      process_records author, records_for_author(author)
+      logger.info "#{self.class} - processing author: #{author.id}"
+      uids = process_records author, records_for_author(author)
+      logger.info "#{self.class} - processed author: #{author.id}"
+      uids
+    rescue StandardError => err
+      message = "#{self.class} - harvest failed for author"
+      NotificationManager.error(err, message, self)
     end
 
     # Harvest DOI publications for an author


lib/web_of_science/harvester.rb
```diff
@@ -10,13 +10,20 @@ class Harvester < ::Harvester::Base
...
         begin
           process_author(author)
-        rescue StandardError => e
-          NotificationManager.error(e, "WebOfScience harvest failed for author: '#{author.id}'", self)
+          logger.info "#{self.class} - processed author: #{author.id}"
+          author_success += 1
+        rescue StandardError => err
+          message = "#{self.class} - harvest failed for author: #{author.id}"
+          logger.error message
+          NotificationManager.error(err, message, self)
```
atz: We don't need two simultaneous loggers.

darrenleeweber: Pushed.

lib/web_of_science/harvester.rb
```diff
@@ -10,13 +10,19 @@ class Harvester < ::Harvester::Base
...
     # @return [void]
     def harvest(authors)
+      logger.info "#{self.class} - started harvest(authors) - #{authors.count} authors"
+      author_success = 0
       authors.each do |author|
         begin
           process_author(author)
-        rescue StandardError => e
-          NotificationManager.error(e, "WebOfScience harvest failed for author: '#{author.id}'", self)
+          logger.info "#{self.class} - processed author: #{author.id}"
```
atz: If the problem is identifying a given author that blows things up, you probably want to log *before* `process_author`.  Otherwise the log will only ever identify the author previous to the problematic one.

atz: (Because OOM kills the process and no `StandardError` is caught.)

darrenleeweber: Pushed up a revision, placed the author logging in the process_author method instead.

lib/web_of_science/harvester.rb
```diff
@@ -10,21 +10,30 @@ class Harvester < ::Harvester::Base
     # @param [Enumerable<Author>] authors
     # @return [void]
     def harvest(authors)
+      logger.info "#{self.class} - started harvest(authors) - #{authors.count} authors"
```
atz: To be clear, this `.count` will issue a new SQL query.

darrenleeweber: It's a one-off and quick query, I assume.  For logs, it's a sanity check on how many authors are going to be harvested.  Since that number is dynamic, it's useful log data.  It's also a useful before/after sanity check on the success rate for the harvest.

atz: Yeah, not objecting, just making sure we were on the same page.

lib/web_of_science/harvester.rb
```diff
@@ -10,21 +10,30 @@ class Harvester < ::Harvester::Base
...
-          process_author(author)
-        rescue StandardError => e
-          NotificationManager.error(e, "WebOfScience harvest failed for author: '#{author.id}'", self)
-        end
+        process_author(author)
+        author_success += 1
       end
+      logger.info "#{self.class} - completed harvest(authors) - #{author_success} processed"
+    rescue StandardError => err
+      message = "#{self.class} - harvest(authors) failed"
```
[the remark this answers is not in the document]
darrenleeweber: Can the rescue block here see the `author_success` variable?  Not entirely confident of the ruby scope rules on this one.  If it's in scope, sure, it could be added.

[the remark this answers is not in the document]
darrenleeweber: Done and pushed.

Issue #467, opened 2017-11-27
Find and document all of the places where ScienceWire Source Record is called


peetucket:

darrenleeweber:
This `git grep` is a comprehensive list:
```
$ git grep 'SciencewireSourceRecord'
app/models/publication.rb:60:    find_by_pmid(pmid) || SciencewireSourceRecord.get_pub_by_pmid(pmid) || PubmedSourceRecord.get_pub_by_pmid(pmid)
app/models/publication.rb:64:    find_by_sciencewire_id(sw_id) || SciencewireSourceRecord.get_pub_by_sciencewire_id(sw_id)
app/models/publication.rb:143:    sw_source = SciencewireSourceRecord.find_by_sciencewire_id(sciencewire_id)
app/models/publication.rb:152:      sw_source_record = SciencewireSourceRecord.find_by_sciencewire_id(sciencewire_id)
app/models/sciencewire_source_record.rb:4:class SciencewireSourceRecord < ActiveRecord::Base
app/models/sciencewire_source_record.rb:16:    SciencewireSourceRecord.convert_sw_publication_doc_to_hash(publication_item)
app/models/sciencewire_source_record.rb:127:        source_records << SciencewireSourceRecord.new(
app/models/sciencewire_source_record.rb:134:        NotificationManager.error(e, "Cannot create SciencewireSourceRecord: sciencewire_id: #{sciencewire_id}, pmid: #{pmid}", self)
config/initializers/okcomputer.rb:54:      SciencewireSourceRecord,
db/migrate/20130513135555_create_sciencewire_source_records.rb:1:class CreateSciencewireSourceRecords < ActiveRecord::Migration
lib/csl/citation.rb:86:          # SciencewireSourceRecord.convert_sw_publication_doc_to_hash
lib/csl/citation.rb:277:      # SciencewireSourceRecord.convert_sw_publication_doc_to_hash
lib/notification_manager.rb:14:      when ScienceWireHarvester, ScienceWireClient, SciencewireSourceRecord
lib/pubmed_harvester.rb:13:        result << SciencewireSourceRecord.convert_sw_publication_doc_to_hash(sw_doc)
lib/science_wire_client.rb:293:        pub_hash = SciencewireSourceRecord.convert_sw_publication_doc_to_hash(sw_xml_doc)
[...]

PR #605, opened 2018-01-10
Fix #503: remove find_or_create_by_sciencewire_id

### LINKED ISSUE #503: Remove unneeded find_or_create_by_sciencewire_id
connected to #467

atz:
Fix #503

### DIFF
--- a/app/api/sul_bib/authorship_api.rb
+++ b/app/api/sul_bib/authorship_api.rb
@@ -104,7 +104,8 @@ module SulBib
       # @param [String] sw_id ScienceWire ID
       # @return [Publication]
       def get_publication_via_sciencewire!(sw_id)
-        Publication.find_or_create_by_sciencewire_id(sw_id) ||
+        Publication.find_by(sciencewire_id: sw_id) ||
+          SciencewireSourceRecord.get_pub_by_sciencewire_id(sw_id) ||
           log_and_error!("The ScienceWire:#{sw_id} publication was not found either locally or at ScienceWire.")
       end
 
--- a/app/models/publication.rb
+++ b/app/models/publication.rb
@@ -56,10 +56,6 @@ class Publication < ActiveRecord::Base
     find_by_pmid(pmid) || SciencewireSourceRecord.get_pub_by_pmid(pmid) || PubmedSourceRecord.get_pub_by_pmid(pmid)
   end
 
-  def self.find_or_create_by_sciencewire_id(sw_id)
-    find_by_sciencewire_id(sw_id) || SciencewireSourceRecord.get_pub_by_sciencewire_id(sw_id)
-  end
-
   def self.find_by_doi(doi)
     Publication.includes(:publication_identifiers)
                .find_by("publication_identifiers.identifier_type": 'doi', "publication_identifiers.identifier_value": doi)
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -13,6 +13,7 @@ AllCops:
     - 'bin/**/*'
     - 'db/**/*'
     - 'config/**/*'
+    - 'spec/factories/*.local.rb'
     - 'spec/fixtures/**/*'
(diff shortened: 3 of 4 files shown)


app/models/publication.rb
```diff
@@ -56,10 +56,6 @@ def self.find_or_create_by_pmid(pmid)
     find_by_pmid(pmid) || SciencewireSourceRecord.get_pub_by_pmid(pmid) || PubmedSourceRecord.get_pub_by_pmid(pmid)
   end
 
-  def self.find_or_create_by_sciencewire_id(sw_id)
-    find_by_sciencewire_id(sw_id) || SciencewireSourceRecord.get_pub_by_sciencewire_id(sw_id)
-  end
-
```
darrenleeweber: `Publication` has no method defined for `'find_by_sciencewire_id`, i.e.

```
$ git grep 'def find_by_sciencewire_id'  # returns nothing
$ git grep 'find_by_sciencewire_id' app/models/
app/models/publication.rb:60:    find_by_sciencewire_id(sw_id) || SciencewireSourceRecord.get_pub_by_sciencewire_id(sw_id)
app/models/publication.rb:147:      sw_source_record = SciencewireSourceRecord.find_by_sciencewire_id(sciencewire_id)
```

So if this method works, it must be an AR dynamic method; wondering what rubocop is complaining about that needs to be silenced?  There is a db-field for `Publication#sciencewire_id`, perhaps this rubocop warning is to avoid the dynamic method and use the named field, e.g. maybe it should be this instead:
```
Publication.find_by(sciencewire_id: sw_id)
```

darrenleeweber:
In the context of #467, we need to check that these calls only hit existing db-content without making any SW-API calls.  This requirement could be a new spec to test that there are no calls to a SW-client?

app/models/publication.rb
atz: Ah, that is a good point.  Might be able to satisfy rubocop after all.

atz:
Right, this is disentangling the Publication model from the stuff public method(s) that are sciencewire-specific, which makes them easier to remove altogether.  But you are correct that this just changes the location of the code and not the overall behavior. 

Certainly ActiveRecord's `find_by` is not hitting any remote backends, so it is `SciencewireSourceRecord.get_pub_by_sciencewire_id` that is doing that.

app/models/publication.rb
atz: Yeah, that worked after all.

darrenleeweber:
Are we going to tackle the SW-API call in this PR or spin off another issue for that?  I'm just concerned that we don't get too far away from the intent of #467 and related work that might use the new Settings that enable/disable the SW-API.  I'm not sure we have an issue or issues to address this and wonder if the current focus on this PR is a good time to look into it.  If we spin off an issue, we can move forward with this PR as is I guess.

atz:
This isn't #467, it is #503.  I don't intend to combine the two.

app/api/sul_bib/authorship_api.rb
```diff
@@ -104,7 +104,8 @@ def get_publication_via_pubmed!(pmid)
       # @param [String] sw_id ScienceWire ID
       # @return [Publication]
       def get_publication_via_sciencewire!(sw_id)
-        Publication.find_or_create_by_sciencewire_id(sw_id) ||
+        Publication.find_by(sciencewire_id: sw_id) ||
+          SciencewireSourceRecord.get_pub_by_sciencewire_id(sw_id) ||
           log_and_error!("The ScienceWire:#{sw_id} publication was not found either locally or at ScienceWire.")
       end
```
darrenleeweber: `Publication.find_by(sciencewire_id: sw_id)` 👍 

- wondering whether the `SciencewireSourceRecord.get_pub_by_sciencewire_id(sw_id)` needs to change in this PR to focus only on hitting the db without any SW-API calls or if that might happen in a follow on PR that will avoid hitting the SW-API.

PR #609, opened 2018-01-10
Bibtex identifiers

darrenleeweber:
Extracted from #512 
- this class encapsulates the extraction of publication identifiers from a BibTex record
- this is part of a larger refactoring for the citation mapping code
- this piece is self contained, it can be reviewed in isolation


atz:
@darrenleeweber can you give a little context where you might see this class being used?

darrenleeweber:
@atz - see #512 where it is used in other bibtex classes; I just want to squeeze out some content from that huge PR to make it easy for reviews and gradually arrive at the bigger picture in 512.  I know it looks like this stands alone without any use in this PR, but it will be used when follow-up PRs are crafted out of 512 to use this one.

PR #613, opened 2018-01-10
Csl::Mapper - extract class Csl::SciencewireMapper

darrenleeweber:
Extracted from #512

This is a straight extract-class refactor to clarify responsibilities for the Csl::Mapper class.


peetucket:
Csl is "citation style language" right?

PR #614, opened 2018-01-11
Refactors yard and cleanup

atz:


app/models/author.rb
```diff
@@ -121,44 +119,48 @@ def mirror_author_identities(import_settings)
...
+    seed_hash[:cap_first_name]  = pref_name['firstName']
+    seed_hash[:cap_middle_name] = pref_name['middleName']
+    seed_hash[:cap_last_name]   = pref_name['lastName']
+    # TODO: duplicative values
+    seed_hash[:emails_for_harvest]    = seed_hash[:email]
+    seed_hash[:preferred_first_name]  = seed_hash[:cap_first_name]
     seed_hash[:preferred_middle_name] = seed_hash[:cap_middle_name]
-    seed_hash[:preferred_last_name] = seed_hash[:cap_last_name]
+    seed_hash[:preferred_last_name]   = seed_hash[:cap_last_name]
     seed_hash
```
darrenleeweber: All these changes to `seed_hash` are puzzling and it's difficult to see exactly what is just style/whitespace and what is functional changes (if any).  The previous code seems to default to an empty string `''` but the new code will default to `nil` values.  That seems like a functional change and could have some impact.  What are the consequences of this change?

darrenleeweber: Also puzzled by the idea of the hash default, consider:
```ruby
[15] pry(main)> x = Hash.new('')
=> {}
[16] pry(main)> x[:abc]
=> ""
[17] pry(main)> x[:abc] = nil
=> nil
[18] pry(main)> x[:abc]
=> nil
```

atz: It's the repetition of a dozen of failovers to the same default value that calls for a Hash default, e.g.:

```ruby
seed_hash[:cap_first_name] = pref_name['firstName'] || ''
seed_hash[:cap_middle_name] = pref_name['middleName'] || ''
seed_hash[:cap_last_name] = pref_name['lastName'] || ''
...
```

I agree w/ doing more in a single block, where possible (I think I changed the initial lines to at least do the first 3 at once) and you're right that this gets us closer to that.  The exceptions are the `duplicative values` which rely on values already being set (or having to repeat dives down levels of incoming hash).

atz: You are right about the behavior where an explicit `nil` value is assigned.  The default is invoked when the hash has no matching key.  That's why I documented the expectations about the incoming parameter as:
```ruby
@param [Hash<String => [String, Hash]>]
```
I.E., `nil` values are not expected.  (Of course this is not comprehensive since I didn't fully specify the nested Hash characteristics, but you get the idea.)   If this concern is material, we should make a test case for it.  The old approach could be applied or filters on the instantiation of the smaller hashes.

app/api/sul_bib/publications_api.rb
```diff
@@ -10,13 +10,11 @@ class PublicationsAPI < Grape::API
...
-          pub = Publication.find(id)
-        rescue ActiveRecord::RecordNotFound
-          error!({ 'error' => 'No such publication', 'detail' => "You've requested a non-existant publication." }, 404)
-        end
+        pub = Publication.find(id)
         error!('Gone - old resource deleted.', 410) if pub.deleted? && !request.delete?
         pub
+      rescue ActiveRecord::RecordNotFound
+        error!({ 'error' => 'No such publication', 'detail' => "You've requested a non-existant publication." }, 404)
       end
```
darrenleeweber: Generally :+1: on this change
The extra check on the delete here adds a responsibility to this method that wasn't there before; code climate is complaining about the extra responsibility.  I get that the single responsibility principle might be a concern, but I'm not entirely clear on the whole context of how this method is used in this part of the API.  If it helps to DRY the code, then maybe this is best place for it.  I don't have any strong preferences about it, just commenting.
It's curious that the `error!` is accepting either a `String` or a `Hash` here.  I'm not entirely familiar with where this error method is defined (rails, grape, ?); just wanting to confirm that it does accept a `String` and also that when it is called by a `Hash` whether that has some impact on how the CAP-team client can parse the error response (if they do).

app/models/author.rb
darrenleeweber: Perhaps another syntax option is a fetch, e.g. `pref_name.fetch('firstName', '')`.  In general, it's a pity the key strings are all mapped to different strings.

app/models/author.rb
```diff
@@ -150,12 +150,14 @@ def self.build_attribute_hash_from_cap_profile(auth_hash)
...
+  # @param [Cap::Client] cap_client
+  # @return [Author] newly fetched, created and saved Author object
+  def self.fetch_from_cap_and_create(profile_id, cap_client = Cap::Client.new)
     profile_hash = cap_client.get_auth_profile(profile_id)
     a = Author.new
     a.update_from_cap_authorship_profile_hash(profile_hash)
-    a.save! # TODO: does this nullify `new_record?` later on, and why save now?
+    a.save!
     a
   end
```
darrenleeweber: For the yardoc - it could add the `@raise` info about AR exceptions that `save!` could generate.  (Assuming this method does not intend to handle any of those exceptions here.)

darrenleeweber: This addition to yardoc might be considered superfluous, because this is an AR model and clients should generally expect it to raise AR exceptions.  I think the complication is that this is doing some HTTP calls that might generate some non-AR exceptions (depending on where they get handled).  Like a lot of "utility" methods in this app, this one may help to DRY the code, but might add confusing responsibility to the AR model; the responsibility for this method really belongs somewhere in a CAP harvester or something, I guess.  Anyhow, this change is not intended to handle that, but adding the yardoc prompts such thoughts about it.

app/models/author.rb
darrenleeweber: Another option for nested hash, where intermediates might be nil, is `dig` (but it doesn't have a default option like fetch):
- https://docs.ruby-lang.org/en/2.4.0/Hash.html#method-i-dig

app/api/sul_bib/publications_api.rb
```diff
@@ -120,17 +100,15 @@ def original_source
...
 
     desc 'DELETE - MARK A RECORD AS OBSOLETE'
     params do
       requires :id
     end
     delete ':id' do
-      pub = publication_find(params[:id])
-      pub.delete!
+      publication_find(params[:id]).delete!
     end
```
darrenleeweber: :+1: on the chain change;  I understand this doesn't change the existing behavior, so it's fine.  Looking at it prompts a thought about whether HTTP response codes should be different depending on whether the record is already deleted (i.e. it's a no-op) or this call does in fact delete it.  If those HTTP response codes should be different, we might want a backlog issue to look into that so this API is better RESTful citizen.

app/api/sul_bib/publications_api.rb
```diff
@@ -17,37 +16,18 @@ def publication_find(id)
...
 
-      # Check for existing authors.
-      # @param authorship_list [Array<Hash>]
-      # @return [boolean] true if any of the authors exist.
-      def existing_authors?(authorship_list)
-        # At least one of the authors in the authorship array must exist.
-        return false if authorship_list.nil?
-        authorship_list.any? { |auth| Contribution.authorship_valid?(auth) }
-      end
-
```
darrenleeweber: Appears to be so, i.e.
```
$ git grep 'existing_authors?'  # on master
app/api/sul_bib/publications_api.rb:25:      def existing_authors?(authorship_list)
```
I suppose this prompts the question about what to do with it, the options are to remove it (as here) or find somewhere that it was supposed to be called.  Perhaps this was replaced by `Contribution.authorship_valid?` calls without removing the method (need to look at a git history or something to find the answer).  OK with this, I guess.

app/api/sul_bib/publications_api.rb
```diff
@@ -17,37 +16,18 @@ def publication_find(id)
...
-          end
-        end
-        status
+        return false if authorship_list.blank?
+        groups = authorship_list.group_by { |auth| Contribution.authorship_valid?(auth) }
+        unknowns = groups[false] || []
+        cap_ids = unknowns.map { |auth| auth[:cap_profile_id] }.compact
+        cap_ids.map { |id| Author.fetch_from_cap_and_create(id) }.any? ||
+          groups[true].any?
       end
```
darrenleeweber: Generally OK with this, if I understand it correctly, it behaves the same way as before.  It might introduce extra loops, but it's not important here because the authorship array is not large.

The only significant concern is that `Author.fetch_from_cap_and_create(id)` could throw an exception somewhere that could blow up before it gets to the boolean `.any?` calls.  (The prior code sorta had this problem too, it seems; so the code enforces that `cap_profile_id` params must belong to real records somewhere, but the API might return a 500 instead of a more graceful handling of it.)  Using the `Author.fetch_from_cap_and_create(id)` is an attempt to fulfill the request when we don't have it already and it can be retrieved; I'm not entirely sure what the behavior should be when it can't be retrieved and/or saved.  It does kinda significantly complicate the API because it introduces a "loop back" call to the API that the request is coming from in the first place; kinda prompts the thought about whether the response should throw the problem back to the client, which could then hit a different API endpoint to first add a new author resource (but that's not the responsibility of this PR, of course).

app/models/author.rb
atz: Yeah, I agree w/ your line of thought here.

app/api/sul_bib/publications_api.rb
atz: Yeah, the design of this original method seems off.  To me, it seems like it really only matters that the **first** valid author is found.  But the existing code goes on to try to fetch/create any other authors also, without any provision for that possibly failing.  Certainly the execution behavior would be more predictable if at most 1 author were created, but I didn't want to change that here.

PR #617, opened 2018-01-11
CSL citation spec fixtures

darrenleeweber:
Extracted from #512 for easier review

These spec fixture data are extracted from one spec file so they are accessible for specs on new mapper classes in the `lib/csl/` module.  This is a prelude to creating more specs for that module, placed into this PR as these changes are easier to review like this.


atz:
Aren't these basically trying to be factories?

PR #618, opened 2018-01-11
WOS queries singleton for harvester

darrenleeweber:
This is a follow-up to #596 

- the delegated methods can be private in the harvester
- this also modifies a query mock for the authorship-api to bring it into line with this change
  - this mock might have been broken by #596 and surfaces in #583 as new VCR recordings

### DIFF
--- a/lib/web_of_science/harvester.rb
+++ b/lib/web_of_science/harvester.rb
@@ -5,7 +5,6 @@ module WebOfScience
   # This class is responsible for processing WebOfScience API response data
   # to integrate it into the application data models.
   class Harvester < ::Harvester::Base
-    delegate :logger, to: :WebOfScience
 
     # @param [Enumerable<Author>] authors
     # @return [void]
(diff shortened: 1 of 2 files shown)


lib/web_of_science/harvester.rb
```diff
@@ -73,11 +72,15 @@ def process_uids(author, uids)
...
       return [] if uids.empty?
-      process_records author, wos_queries.retrieve_by_id(uids)
+      process_records author, queries.retrieve_by_id(uids)
     end
 
     private
 
+      delegate :logger, to: :WebOfScience
+
+      delegate :queries, to: :WebOfScience
```
atz: Could combine these 2 lines, if you like:  
```ruby
delegate :logger, :queries, to: :WebOfScience
```

darrenleeweber: Pushed.

PR #620, opened 2018-01-12
CSL role mapper

darrenleeweber:
Extracted from #512 

An extract class refactor to clarify responsibility for CSL mapper classes.
- adds some rubocop auto fixes and updates the TODOs
- fixes some broken citation specs for first author name spacing

### Tech Debts

Work on this PR identified and ticketed some existing tech-debt, see #622 and #623 
- it's not the goal of this PR to resolve that existing tech-debt


lib/csl/role_mapper.rb
```diff
@@ -93,23 +97,29 @@ def parse_family_name(author)
...
         # Add period after single character (initials).
+        # @param [Hash] author various key:value pairs for author names
         # @return [String, nil]
         def parse_given_names(author)
-          given_names = ''
-          %i(firstname middlename).map { |k| author[k] }.reject(&:blank?).each do |name_part|
-            given_names << ' ' << name_part
-            given_names << '.' if name_part =~ /^[[:upper:]]$/
-          end
+          names = %i(firstname middlename).map { |k| author[k] }
```
atz: If you prefer, you could just use: 
```ruby
names = author.slice(%i(firstname middlename))
```

darrenleeweber: It's plucking values from a Hash by those keys (I need to double check the indifferent access too or symbolize keys).  I wish it supported `.slice`
```ruby
[5] pry(main)> author = {firstname: 'mad', middlename: 'm', lastname: 'max' }
=> {:firstname=>"mad", :middlename=>"m", :lastname=>"max"}
[6] pry(main)> author.slice([:firstname, :lastname])
NoMethodError: undefined method `slice' for {:firstname=>"mad", :middlename=>"m", :lastname=>"max"}:Hash
from (pry):6:in `__pry__'
```

darrenleeweber: Also need to be confident the `slice` preserves the order of the key:value pairs, as the order of the first and last name are important, which the current construct ensures.

darrenleeweber: I'm not entirely confident that `.slice` will preserve the order, plus it returns a `Hash` whereas the map extracts an array of values (in the correct order).

darrenleeweber: I push up a change to use symbols, as we generally assume that the `Publication.pub_hash` data is all symbolized keys.

PR #624, opened 2018-01-12
WebOfScience::ProcessRecords - convert records to a mutable array

### LINKED ISSUE #600: Fix "method_complexity" issue in lib/web_of_science/process_records.rb
Method `process_links` has a Cognitive Complexity of 9 (exceeds 5 allowed). Consider refactoring.

https://codeclimate.com/github/sul-dlss/sul_pub/lib/web_of_science/process_records.rb#issue_5a54039aad19160001000115

darrenleeweber:
Extracted from #583 to focus on this class in this PR.

- aims to reduce memory consumption by using a mutable array of records
- progressive filters now select or reject records in-place
- initialize no longer holds onto the original set of records
  - they are immediately filtered for accepted-dbs

### Tech-Debt
Fix #600

### DIFF
--- a/lib/web_of_science/process_records.rb
+++ b/lib/web_of_science/process_records.rb
@@ -30,6 +30,8 @@ module WebOfScience
       attr_reader :author
       attr_reader :records
 
+      delegate :links_client, to: :WebOfScience
+
       # ----
       # Record filters and data flow steps
 
@@ -127,7 +129,18 @@ module WebOfScience
       # ----
       # WOS Links API methods
 
-      # Retrieve a batch of publication identifiers from the Links-API
+      # Retrieve a batch of publication identifiers for WOS records from the Links-API
+      # @example {"WOS:000288663100014"=>{"pmid"=>"21253920", "doi"=>"10.1007/s12630-011-9462-1"}}
+      # @return [Hash<String => Hash<String => String>>]
+      def retrieve_links
+        uids = records.map { |rec| rec.uid if rec.database == 'WOS' }.compact
+        links_client.links uids
+      rescue StandardError => err
+        message = "Author: #{author.id}, retrieve_links failed"
+        NotificationManager.error(err, message, self)
+      end
+
+      # Integrate a batch of publication identifiers from the Links-API
       #
       # IMPORTANT: add nothing to PublicationIdentifiers here, or new_records will reject them
       # Note: the WebOfScienceSourceRecord is already saved, it could be updated with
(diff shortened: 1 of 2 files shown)


lib/web_of_science/process_records.rb
```diff
@@ -11,12 +11,12 @@ def initialize(author, records)
...
       raise(ArgumentError, 'records must be an WebOfScience::Records') unless records.is_a? WebOfScience::Records
       @author = author
-      @records = records
+      @records = records.to_a
     end
 
     # @return [Array<String>] WosUIDs that create a new Publication
     def execute
-      return [] if records.count.zero?
+      return [] if records.empty?
```
atz: 👍

atz:
I am missing some context here.  `@records` was getting assigned a `WebOfScience::Records` object, changed here to get be `.to_a` off that object.  That comes from `Enumerable` and hits `WebOfScience::Records#each`, which is:
```ruby
rec_nodes.each { |rec| yield WebOfScience::Record.new(record: rec.to_xml) }
```
Right so far?  

Then, with the mutable array, the `select!` and `reject!` calls should logically reduce the size of the `@records` variable.  I think that is what you are going for.  I'm sorta OK throwing out the kafka/functional-influenced design to keep the memory footprint down.  BUT we're also giving up processing as we go via `Enumerable`, so that is a downside, even for our intended purpose of smaller footprint.  Do you have any measurements to make sure this isn't *worse*?

IMO, the ideal places to look for memory savings are where we duplicate XML documents/objects.  I.E. the places `WebOfScience.Records` calls `.dup` and the places we added serialization/deserialization.  `WebOfScience.Record`

Also, why do `filter_databases` at this level instead of just calling `WebOfScience::Records#by_database()`?  If we're giving up on immutability, we can undo  7fb193642ea58b038061b1b6b8a22705609f4645 and use it correctly this time.

lib/web_of_science/process_records.rb
```diff
@@ -34,42 +34,37 @@ def execute
...
       end
 
       ## 1
       # Filter and select new WebOfScienceSourceRecords
-      # @return [Array<WebOfScience::Record>]
       def filter_databases
-        return [] if records.count.zero?
-        return records if Settings.WOS.ACCEPTED_DBS.empty?
-        records.select { |rec| Settings.WOS.ACCEPTED_DBS.include? rec.database }
+        return if records.empty? || Settings.WOS.ACCEPTED_DBS.empty?
```
atz: This `Settings.WOS.ACCEPTED_DBS.empty?` check should be moved up into `#execute`.  No point even connecting if we aren't going to accept any records returned!

darrenleeweber: Ah, yea; this is probably a cause to raise an exception actually, smells of some kind of oversight in creating or changing the settings file.

darrenleeweber: Pushed up an exception the `Settings.WOS.ACCEPTED_DBS.empty?`

PR #626, opened 2018-01-13
WOS harvest check contributions

darrenleeweber:
Extracted out of #583 to focus on checking existing publications for author contributions.

The aim of this PR is to replace simple checks for any `PublicationIdentifier` with logic that finds or creates a contribution for an author on an existing `Publication`, using only unique identifiers.  It's not enough to just check for existing `PublicationIdentifier`, because there must also be a contribution to the associated `Publication` for the current harvest-author.  There are cases where the harvest will be finding multiple authors for a single `Publication`.


lib/notification_manager.rb
```diff
@@ -17,7 +17,8 @@ def error(e, message, callee = nil)
         log_exception(pubmed_logger, log_message, e)
       when Cap::AuthorsPoller, Cap::Client
         log_exception(cap_logger, log_message, e)
-      when WebOfScience::Client, WebOfScience::Harvester, WebOfScience::ProcessRecords, WebOfScience::Record
+      when WebOfScience::Client, WebOfScience::Contributions,
```
atz: Still literally no reason to keep using this bogus pattern.

spec/lib/web_of_science/contributions_spec.rb
```diff
@@ -0,0 +1,67 @@
+describe WebOfScience::Contributions do
+  class WosContributions
```
atz: Please use an anonymous class to prevent namespace pollution in other tests.  This has been a problem in our other projects where the same test "setup" gets replicated and changed in different files, but the class definitions persist/interact/conflict.

E.g. 
```ruby
let(:test_class) { Class.new { include WebOfScience::Contributions } } # or described_class
subject(:wos_contributions) { test_class.new }
```

atz:
Not sure I like the module mix-in approach here since it seems like maybe there is more in common between the 2 places the module is used (Harvester and ProcessRecords) that could align in another OO pattern.  Not a huge deal, but maybe you can comment on it?  

Otherwise I appreciate you trying to address a rather tricky portion of logic.

lib/notification_manager.rb
darrenleeweber: I can't tell if this is a general gripe that needs more discussion outside this PR or if this is asking for a change to this PR.  I'm assuming for the former for now.

spec/lib/web_of_science/contributions_spec.rb
darrenleeweber: Pushed up this change, thanks.

darrenleeweber:
I agree that the `WebOfScience::Contributions` module as a mix-in is a slight departure from the more common OOP class instances used generally in the `WebOfScience` module.  It's essentially a collection of "class methods" that are needed in two places (for this PR).  I took a quick look around for similar use-cases in the app, but didn't immediately identify any; so I understand why your asking about it.  For the purposes of this PR, I'm punting on this for now, but I would like to have some opportunity for additional research/discussion about this de-dup problem and how this approach works here in this WOS harvesting and whether it needs to be applied elsewhere.  If we can identify additional use cases, that would be good reason to refactor this and surface it at a higher level.

Issue #625, opened 2018-01-13
Fix existing nil provenance values


atz:
We have a lot of runaround logic pertaining to `Publication` provenance, even basic things like:
```ruby
  def pubmed_pub?
    pub_hash[:provenance].to_s.downcase.include?('pubmed')
  end
```

This is weird for many reasons.  This should be an exact `==` comparison, not `include?`.  All existing values  are already lowercase.  But some records have `nil` provenance!  Using:

```ruby
counts = Hash.new(0)
Publication.find_in_batches(batch_size: 5000).each{ |batch| 
  batch.map{ |x| x.pub_hash[:provenance] }.each_with_object(counts){ |x, h| h[x] += 1 }
} 
```
We get `counts` of:
```ruby
# cap-qa
{"pubmed"=>3806, "sciencewire"=>342042, nil=>10, "cap"=>29493, "batch"=>27146, "wos"=>337}

# prod
{"pubmed"=>5174, "sciencewire"=>379403, nil=>10, "cap"=>35197, "batch"=>26181}
```

`nil` values should be impossible.  We should correct those records.

Note that all the values are just regular single-valued strings (the `include?` made me think there might be Arrays).  

IMO, we should make provenance a regular attribute/field on `Publication`.  It can appear in `pub_hash` in exactly the same way, but we shouldn't have to fetch, instantiate and drill into each object just to filter by provenance.  By making it required we can guarantee that `nil` values are impossible.

peetucket:
ok with making provenance a top level attribute on publication --- did you happen to spit out IDs for those records that are nils before I go and run a (likely) expensive query like above again?  maybe obvious to fix once we look at the records

atz:
The records on prod are:
```ruby
[...]

PR #631, opened 2018-01-18
Pub controller conversion

atz:
Looks bigger than it really is, but I can break off pieces if necessary.

Conflict in the `rubocop_todo.yml` counts are 100% unimportant.

### DIFF
--- a/app/controllers/publications_controller.rb
+++ b/app/controllers/publications_controller.rb
@@ -63,7 +63,9 @@ class PublicationsController < ApplicationController
       msg << "title '#{params[:title]}'"
       logger.info(msg)
       if Settings.WOS.enabled
-        wos_matches = [] # TODO
+        query = "TI=#{params[:title]}"
+        query += " AND PY=#{params[:year]}" if params[:year]
+        wos_matches = WebOfScience.queries.user_query(query).to_a # TODO: limit
         all_matching_records += wos_matches
         logger.debug(" -- WOS (#{wos_matches.length})")
       end
@@ -72,7 +74,7 @@ class PublicationsController < ApplicationController
         all_matching_records += sw_matches
         logger.debug(" -- sciencewire (#{sw_matches.length})")
       end
-      # lastly, check for manual
+      # lastly, always check for manual
       results = Publication.joins(:user_submitted_source_records)
                            .where(UserSubmittedSourceRecord.arel_table[:title].matches("%#{params[:title]}%"))
       results = results.where(year: params[:year]) if params[:year]
--- a/app/controllers/publications_controller.rb
+++ b/app/controllers/publications_controller.rb
@@ -53,51 +53,44 @@ class PublicationsController < ApplicationController
     end
   end
 
-  # desc "Look up existing records by title, and optionally by author, year and source"
+  # Look up existing records by title, and optionally by author, year and source
   def sourcelookup
     all_matching_records = []
+    msg = 'Sourcelookup of '
     if params[:doi]
-      logger.info("Sourcelookup of doi #{params[:doi]}")
-      sources = [Settings.sciencewire_source]
-      all_matching_records += DoiSearch.search params[:doi]
+      msg << "doi #{params[:doi]}"
+      logger.info(msg)
+      all_matching_records += DoiSearch.search(params[:doi])
     elsif params[:pmid]
-      logger.info("Sourcelookup of pmid #{params[:pmid]}")
-      sources = [Settings.sciencewire_source, Settings.pubmed_source]
-      all_matching_records += PubmedHarvester.search_all_sources_by_pmid params[:pmid]
+      msg << "pmid #{params[:pmid]}"
+      logger.info(msg)
+      all_matching_records += PubmedHarvester.search_all_sources_by_pmid(params[:pmid])
     else
       raise ActionController::ParameterMissing, :title unless params[:title].presence
-
-      source = params.fetch(:source, Settings.manual_source + '+' + Settings.sciencewire_source)
-      logger.info("Executing source lookup for title #{params[:title]} with sources #{source}")
-
-      sources = source.split('+')
-
-      if sources.include?(Settings.sciencewire_source)
-        all_matching_records += ScienceWireClient.new.query_sciencewire_for_publication(params[:firstname], params[:lastname], params[:middlename], params[:title], params[:year], params.fetch(:max_rows, 20).to_i)
-        logger.debug(" -- sciencewire (#{all_matching_records.length})")
+      msg << "title '#{params[:title]}'"
+      logger.info(msg)
+      if Settings.WOS.enabled
+        wos_matches = [] # TODO
+        all_matching_records += wos_matches
+        logger.debug(" -- WOS (#{wos_matches.length})")
       end
-
-      if sources.include?(Settings.manual_source)
-        user_submitted_source_records = UserSubmittedSourceRecord.arel_table
-
-        results = UserSubmittedSourceRecord.where(user_submitted_source_records[:title].matches("%#{params[:title]}%"))
-
-        results = results.where(user_submitted_source_records[:year].eq(params[:year])) if params[:year]
-        logger.debug(" -- manual source (#{results.length})")
-
-        all_matching_records += results.map(&:publication)
+      if Settings.SCIENCEWIRE.enabled
+        sw_matches = ScienceWireClient.new.query_sciencewire_for_publication(params[:firstname], params[:lastname], params[:middlename], params[:title], params[:year], params.fetch(:max_rows, 20).to_i)
+        all_matching_records += sw_matches
+        logger.debug(" -- sciencewire (#{sw_matches.length})")
       end
+      # lastly, check for manual
+      results = Publication.joins(:user_submitted_source_records)
+                           .where(UserSubmittedSourceRecord.arel_table[:title].matches("%#{params[:title]}%"))
+      results = results.where(year: params[:year]) if params[:year]
+      logger.debug(" -- manual source (#{results.length})")
+      all_matching_records += results
     end
-    # When params[:maxrows] is nil, rows is -1 and returns everything
-    rows = params[:maxrows].to_i - 1
-    matching_records = all_matching_records[0..rows]
-
-    description = "Search results from requested sources: #{sources.join(', ')}"
-
+    # When params[:maxrows] is nil/zero, -1 returns everything
+    matching_records = all_matching_records[0..params[:maxrows].to_i - 1]
     respond_to do |format|
       format.json do
-        bibjson = wrap_as_bibjson_collection(description, matching_records)
-        self.response_body = JSON.dump(bibjson)
+        self.response_body = JSON.dump(wrap_as_bibjson_collection(msg, matching_records))
       end
     end
   end
--- a/app/controllers/publications_controller.rb
+++ b/app/controllers/publications_controller.rb
@@ -70,7 +70,7 @@ class PublicationsController < ApplicationController
         logger.debug(" -- WOS (#{wos_matches.length})")
       end
       if Settings.SCIENCEWIRE.enabled
-        sw_matches = ScienceWireClient.new.query_sciencewire_for_publication(params[:firstname], params[:lastname], params[:middlename], params[:title], params[:year], params.fetch(:max_rows, 20).to_i)
+        sw_matches = ScienceWireClient.new.query_sciencewire_for_publication(nil, nil, nil, params[:title], params[:year], params.fetch(:max_rows, 20).to_i)
         all_matching_records += sw_matches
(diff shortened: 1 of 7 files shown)


spec/models/publication_spec.rb
```diff
@@ -336,31 +336,28 @@
 
     it 'should refuse to add a publication with the same source record' do
       save_new_publication
-      expect do
-        Publication.build_new_manual_publication(pub_hash, 'some string', 'some where')
-      end.to raise_error(ActiveRecord::RecordNotUnique)
+      expect { Publication.build_new_manual_publication(pub_hash, 'some string') }.to raise_error(ActiveRecord::RecordNotUnique)
     end
```
darrenleeweber: The line was probably broken to avoid rubocop long line issue, plus it's possible to see everything in a github PR; not it's not all visible in this PR as one long line.

spec/models/publication_spec.rb
```diff
@@ -323,7 +323,7 @@
```
darrenleeweber: I don't really get the whole point of this commit.  What's the point of remove a param value?  It seems like adding value to the app would be to add a test for it, rather than remove it completely.  (I may be missing some context here or just don't get the point of these changes.)

spec/models/publication_spec.rb
```diff
@@ -346,19 +346,25 @@
   end
 
   describe 'update_manual_pub_from_pub_hash' do
+    let(:pub) { Publication.build_new_manual_publication({ a: :b }, 'some string') }
```
darrenleeweber: :+1:  -  although ? about removing the last param that was there before the previous commit.

spec/models/publication_spec.rb
```diff
@@ -346,19 +346,25 @@
...
+      pub.save!
+      other = Publication.build_new_manual_publication({ b: :c }, 'some other string')
+      other.update_manual_pub_from_pub_hash({ b: :c }, 'some string')
+      expect { other.save! }.to raise_error(ActiveRecord::RecordNotUnique)
+    end
+
+    it 'does not alter the hash provided' do
+      h = { pub: 'hash' }
+      expect { pub.update_manual_pub_from_pub_hash(h, 'whatev') }.not_to change { h }
     end
```
darrenleeweber: :+1: - this adds value by adding a spec that looks useful.

spec/models/publication_spec.rb
```diff
@@ -346,19 +346,25 @@
...
     it 'should raise an exception if you try to update a record to match an existing source record' do
-      Publication.build_new_manual_publication({ a: :b }, 'some string').save!
-      pub = Publication.build_new_manual_publication({ b: :c }, 'some other string')
-      pub.update_manual_pub_from_pub_hash({ b: :c }, 'some string')
-      expect { pub.save! }.to raise_error(ActiveRecord::RecordNotUnique)
+      pub.save!
+      other = Publication.build_new_manual_publication({ b: :c }, 'some other string')
+      other.update_manual_pub_from_pub_hash({ b: :c }, 'some string')
+      expect { other.save! }.to raise_error(ActiveRecord::RecordNotUnique)
+    end
```
darrenleeweber: :+1: - the use of `other` makes it clear, although what's not entirely clear is why the `other` now matches `pub` after the update.  If `some string` is a particular field on the model, it seems fair to create a `let` for it that uses the field label; e.g., if the field is title, a `let(:title) { 'some title' }` would help to clarify what this is about.

spec/models/publication_spec.rb
```diff
@@ -346,19 +346,25 @@
   end
```
darrenleeweber: This commit and the previous commit "Finer test targeting" are dependent commits, might be good to squash them.

spec/models/publication_spec.rb
darrenleeweber: I suppose I'd prefer to see `some where` become a `let(:provenance) { 'some provenance' }` or have it actually explicitly use and create unit tests around the default value or changes to the default value.

app/models/publication.rb
```diff
@@ -86,8 +86,7 @@ def self.build_new_manual_publication(pub_hash, original_source_string, provenan
 
   # @return [self]
   def update_manual_pub_from_pub_hash(incoming_pub_hash, original_source_string, provenance = Settings.cap_provenance)
-    incoming_pub_hash[:provenance] = provenance
-    self.pub_hash = incoming_pub_hash.dup
+    self.pub_hash = incoming_pub_hash.merge(provenance: provenance)
     match = UserSubmittedSourceRecord.find_by_source_data(original_source_string)
```
darrenleeweber: :+1:

spec/models/publication_spec.rb
atz: It's a setup for a later commit that removes the 3rd param because it isn't actually used.  These tests aren't testing with that param, so it shouldn't be contentious to tailor them to what they *are* testing.

app/api/sul_bib/publications_api.rb
```diff
@@ -78,10 +78,9 @@ def original_source
...
       new_pub = params[:pub_hash].to_hash.deep_symbolize_keys # Avoid Hashie::Mash!
       old_pub = publication_find(params[:id])
-      case
-      when !old_pub.sciencewire_id.blank? || !old_pub.pmid.blank?
+      if old_pub.sciencewire_id.present? || old_pub.pmid.present? || old_pub.wos_uid.present?
         error!({ 'error' => 'This record may not be modified.  If you had originally entered details for the record, it has been superceded by a central record.', 'detail' => 'missing widget' }, 403)
-      when !validate_or_create_authors(new_pub[:authorship])
+      elsif !validate_or_create_authors(new_pub[:authorship])
         error!('You have not supplied a valid authorship record.', 406)
       end
```
darrenleeweber: :ok:

spec/api/sul_bib/publications_api_spec.rb
```diff
@@ -269,49 +269,87 @@ def validate_authorship(pub_hash, submission)
...
+      it 'deletes an identifier from the db if it is not in the incoming json' do
+        post '/publications', with_isbn_hash.to_json, headers
+        id = Publication.last.id
+        put "/publications/#{id}", with_isbn_deleted_doi.to_json, headers
+        expect(result['identifier'].size).to eq(2)
+        expect(result['identifier']).to include(
+          a_hash_including('type' => 'isbn', 'id' => '1177188188181'),
+          a_hash_including('type' => 'SULPubId', 'url' => "#{Settings.SULPUB_ID.PUB_URI}/#{id}", 'id' => id.to_s)
+        )
+      end
```
darrenleeweber: Does it really behave this way?  Or does something, somewhere in this PR make it behave this way?  Is it OK for a PUT on a publication to delete existing identifiers?  I wonder if this depends on the provenance of the publication?  AFAIK, we do not allow any changes to publication or publication-identifiers unless the provenance is a manual submission ('cap' provenance).  This spec indicates otherwise.

spec/api/sul_bib/publications_api_spec.rb
```diff
@@ -269,49 +269,87 @@ def validate_authorship(pub_hash, submission)
...
+          expect(response.status).to eq(403)
+        end
+        it 'wos_uid' do
+          allow(publication).to receive(:wos_uid).and_return(4)
+          put "/publications/#{id}", with_isbn_hash.to_json, headers
+          expect(response.status).to eq(403)
+        end
+      end
     end
   end
```
darrenleeweber: :+1: to adding these specs; agree with those HTTP response codes.  It looks like this rounds out my earlier comments above about whether the non-cap provenance data is protected from PUTs.

darrenleeweber: So, as a whole, it looks like I approve of this commit - 7d3708cbc91132c4895dba95302b75b3542862db

app/api/sul_bib/publications_api.rb
```diff
@@ -46,7 +46,7 @@ def original_source
       logger.info(original_source)
       pub_hash = params[:pub_hash].to_hash.deep_symbolize_keys # Avoid Hashie::Mash!
       fingerprint = Digest::SHA2.hexdigest(original_source)
-      existing_record = UserSubmittedSourceRecord.where(source_fingerprint: fingerprint).first
+      existing_record = UserSubmittedSourceRecord.find_by(source_fingerprint: fingerprint)
       if existing_record
         logger.info("Found existing record for #{fingerprint}: #{existing_record.inspect}; redirecting.")
         # the GRAPE redirect method issues a 303 (when original request is not a get, and a 302 for a get)
```
darrenleeweber: 6ef1b40bd35c623c168ebbbd36dfa82b644be074 approved

app/models/publication.rb
darrenleeweber: 299add1be9a4fc84a100103fcad14e547468fc13 approved - requesting it is squashed with prior commits that change specs by removing this parameter there also - noting that this PR review would have been easier if they were merged and consolidated into one PR (less guess work during the review)

darrenleeweber: Noting that this change means our API cannot have additional manual consumers outside the 'cap' consumer after this change is made.  In other words, this change ensures there can only ever be one source of manual publications.  That it, it entails, for example, that there could be other Stanford (or other) consumers of this API that we would want to identify as a source (provenance) for manual submissions.  As we have no use case for that now, this might be OK; but if we do discover such a use case, this change might need to be reverted.  Hence, consider crafting the commit/PR for just removing this method parameter so it could be easily reverted.

darrenleeweber: I'm trying to think of whether we have any additional use cases where we need to keep the method param - I suppose we consider all bibtex ingests as belonging to a cap provenance.  So maybe not and the git grep in the commit message seems to suggest not.

app/models/publication.rb
```diff
@@ -225,21 +225,32 @@ def year
   end
```
darrenleeweber: A change from `String#include?` to an equality checker has some functional significance.  That might require analysis of the prod-db to be sure this change will be OK.  Other than that, I'm :+1: for it.  But it's a reserved/qualified :+1: until I can be sure that the prod-db is OK with it.  There's a lot of legacy data in there and all the code analysis in the world is not going to help me now whether or not this works with our prod-db.

lib/science_wire_client.rb
```diff
@@ -287,6 +287,7 @@ def validate_query_id_response_xml(xml_doc)
         !xml_doc.xpath('//queryResultRows').empty?
     end
 
+    # @return [Array<Hash>]
     def send_query_and_return_pub_hashes(xml_query)
       xml_results = query_sciencewire(xml_query)
       xml_results.xpath('//PublicationItem').map do |sw_xml_doc|
```
darrenleeweber: :ok: - c9ab3da8153d1fca239f210f04ae4e1d0414557a approved (without checking that that's what it actually returns, assuming that it's correct)

app/models/publication.rb
atz: See https://github.com/sul-dlss/sul_pub/issues/625 for the analysis.  No problem with inexact matches, but there are 10 `nil` values (a separate issue).

app/models/publication.rb
atz: Can't get much tighter than 4 lines touched.

app/controllers/publications_controller.rb
```diff
@@ -28,82 +28,71 @@ def index
...
-
-      if sources.include?(Settings.sciencewire_source)
-        all_matching_records += ScienceWireClient.new.query_sciencewire_for_publication(params[:firstname], params[:lastname], params[:middlename], params[:title], params[:year], params.fetch(:max_rows, 20).to_i)
-        logger.debug(" -- sciencewire (#{all_matching_records.length})")
+      msg << "title '#{params[:title]}'"
+      logger.info(msg)
+      if Settings.WOS.enabled
+        query = "TI=#{params[:title]}"
+        query += " AND PY=#{params[:year]}" if params[:year]
+        wos_matches = WebOfScience.queries.user_query(query).to_a # TODO: limit
```
peetucket: limit at the query level is tricky, right? since the WoS client doesn't support passing this as a parameter?  But if I read the code correctly, it is already defaulting to 100 ?  or will it just keep retrieving in chunks of 100 (this could be a problem if you typed in a really vague title)

atz: Right, WoS client doesn't expose this functionality.  I have a separate branch where I started to add that.  The ScienceWire integration default is **20**.  WoS default is ALL.  Given enough hits for common strings (and serial trips back to WoS), the CAP client will HTTP timeout before we make a response.

peetucket: ok, so the plan is to address in this workcycle which seems important to keep those timeouts from happening

atz: But not in this PR.

PR #637, opened 2018-01-19
Tweak CSV generation

atz:
- Remove intermediate variable
- Make `journ` the string value needed, like the others, not a hash

### DIFF
--- a/app/controllers/publications_controller.rb
+++ b/app/controllers/publications_controller.rb
@@ -133,10 +133,10 @@ class PublicationsController < ApplicationController
     # @param [Author] author
     # @return [String] contains csv report of an author's publications
     def generate_csv_report(author)
-      csv_str = CSV.generate do |csv|
+      CSV.generate do |csv|
         csv << %w(sul_pub_id sciencewire_id pubmed_id doi wos_id title journal year pages issn status_for_this_author created_at updated_at contributor_cap_profile_ids)
         author.publications.find_each do |pub|
-          journ = pub.pub_hash[:journal] ? pub.pub_hash[:journal] : { name: '' }
+          journ = pub.pub_hash[:journal] ? pub.pub_hash[:journal][:name] : ''
           contrib_prof_ids = pub.authors.pluck(:cap_profile_id).join(';')
           wos_id = pub.publication_identifiers.where(identifier_type: 'WoSItemID').pluck(:identifier_value).first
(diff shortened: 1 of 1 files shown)


atz:
Obviously CodeClimate prefers the documented methods to be as bloated as possible.

app/controllers/publications_controller.rb
```diff
@@ -133,20 +133,19 @@ def wrap_as_bibjson_collection(description, records, page = 1, per_page = 'all')
     # @param [Author] author
     # @return [String] contains csv report of an author's publications
     def generate_csv_report(author)
-      csv_str = CSV.generate do |csv|
+      CSV.generate do |csv|
         csv << %w(sul_pub_id sciencewire_id pubmed_id doi wos_id title journal year pages issn status_for_this_author created_at updated_at contributor_cap_profile_ids)
         author.publications.find_each do |pub|
-          journ = pub.pub_hash[:journal] ? pub.pub_hash[:journal] : { name: '' }
+          journ = pub.pub_hash[:journal] ? pub.pub_hash[:journal][:name] : ''
           contrib_prof_ids = pub.authors.pluck(:cap_profile_id).join(';')
```
darrenleeweber: Maybe `pub.pub_hash.dig(:journal, :name) || ''` works here?
```ruby
[1] pry(main)> x = { journal: {} }
=> {:journal=>{}}
[2] pry(main)> x.dig(:journal, :name)
=> nil
[3] pry(main)> x = { journal: { name: 'name' } }
=> {:journal=>{:name=>"name"}}
[4] pry(main)> x.dig(:journal, :name)
=> "name"
```

darrenleeweber:
It's just the diff-coverage that code-climate is complaining about.  Looks like there are no specs for this and touching it in this PR is uncovering this lack of specs.

app/controllers/publications_controller.rb
atz: Yes, but that's not the point.  We're not worried about traversing the depths of the hash.  The point was to make `journ` a string instead of a structured object that only ever gets used to extract the string.

darrenleeweber: Yes, but dig is better - consider when/if `pub.pub_hash[:journal] == {}` - the current code will assign `nil` to `journ` and that's not what you want here (and looks like it was a bug before too).

PR #644, opened 2018-01-21
WebOfScience::Retriever - fetch records from WOS

### LINKED ISSUE #641: WOS Harvest stops at author 21
This seems to be a recurring problem with data from the cap-qa DB.
```
...
I, [2018-01-19T18:20:02.829885 #11131]  INFO -- : WebOfScience::Harvester - processed author: 19
I, [2018-01-19T18:20:02.830042 #11131]  INFO -- : WebOfScience::Harvester - processing author: 20
I, [2018-01-19T18:20:04.448556 #11131]  INFO -- : WebOfScience::Harvester - processed author: 20
I, [2018-01-19T18:20:04.448665 #11131]  INFO -- : WebOfScience::Harvester - processing author: 21
```

darrenleeweber:
fixes #641 

An extract class refactor to pull out the record fetching from the WebOfScience::Queries class and try to expose an API that allows a consumer to iterate on batches of records.  A strict Enumerable implementation is not possible, because we don't want to store all of the records.  When the total number of records found is large, retrieving them all can exceed memory limits.  The goal of this retriever is to fetch the first batch and then allow a consumer to fetch more batches, as required.  Note that the first batch of records is a special query and then retrieving more records uses the `query_id` from the first query response.

- `WebOfScience::Queries` no longer needs a `WebOfScience::Client`
- `WebOfScience::Queries` no longer returns `WebOfScience::Records`
- `WebOfScience::Queries` now returns `WebOfScience::Retriever`
- `WebOfScience::Retriever` exposes an API to fetch batches of records
  - there are options to fetch and merge all records (but this is not a good idea)
  - the API is simple:  `next_batch?` and `next_batch` returns `WebOfScience::Records` or `nil`
  - there are additional attr_reader for metadata about the retrieval, like `records_found`

Consumer of the Retriever can check the status of records retrieved after each batch.  A total limit should be defined by the query, when using the WOS-Queries.  However, if a total limit is not set in the query, the consumer can choose to retrieve the first N records.  After the first batch is retrieved, the Retriever exposes:
```
    attr_reader :records_found
    attr_reader :records_retrieved
```
Say `records_found == 1000`, then `retriever.next_batch while retriever.next_batch?` will get them all, whereas say `retriever.next_batch while retriever.records_retrieved < 500` would only get 500.


darrenleeweber:
A quick test of this branch indicates that it's working, e.g. this test shows that it harvested records for an author that was never harvested before and, in the second pass, it doesn't re-create the same publications again.
```
PRODUCTION [6] pry(main)> author
=> #<Author:0x00005645b9c1aa98
 id: 25,
 cap_profile_id: 3808>
PRODUCTION [7] pry(main)> WebOfScience.harvester.process_author(author)
=> ["MEDLINE:28036479",
 "WOS:A1985ATX1500001",
 "WOS:000396841300013",
 "WOS:A1996VK07402901",
 "WOS:000374242500022",
 "WOS:000342331403079",
 "WOS:A1997XY10300605",
 "WOS:000411186500729",
 "WOS:000344384900320",
 "WOS:A1987F759700001",
 "WOS:000415229200016",
 "WOS:000189498600380",
 "WOS:A1996UW63400004",
 "WOS:000240005300014",
 "WOS:000417864700001",
 "WOS:000351214000001",
 "WOS:000369270700506",
 "WOS:000076906702317",
 "WOS:A1983PX37700061",
 "WOS:000403888200004",
 "WOS:000321561205025",
 "WOS:000386939500025",
 "WOS:000229950000011",
 "WOS:000324621804061",
 "WOS:000295333300011",
 "WOS:000305505500102",
 "WOS:000408065700006",
 "WOS:000400230700002",
 "WOS:A1977DZ16900068",
 "WOS:000370302500506",
 "WOS:000394467300517",
 "WOS:000349950900161"]
PRODUCTION [8] pry(main)> WebOfScience.harvester.process_author(author)
=> []
```

darrenleeweber:
Running this branch for the problematic author worked OK.  The process ran with a max memory footprint of about 4-6 Gb of RAM.

```
I, [2018-01-22T18:19:07.750304 #27375]  INFO -- : WebOfScience::Harvester - processing author: 21
I, [2018-01-22T18:22:34.683550 #27375]  INFO -- : WebOfScience::Harvester - 1389 records found by author query
I, [2018-01-22T19:06:24.559283 #27375]  INFO -- : WebOfScience::Harvester - 70 records created new publications
I, [2018-01-22T19:06:24.559442 #27375]  INFO -- : WebOfScience::Harvester - processed author: 21
I, [2018-01-22T19:10:40.858416 #27375]  INFO -- : WebOfScience::Harvester - processing author: 21
I, [2018-01-22T19:14:01.319570 #27375]  INFO -- : WebOfScience::Harvester - 1389 records found by author query
I, [2018-01-22T19:25:47.527959 #27375]  INFO -- : WebOfScience::Harvester - 0 records created new publications
I, [2018-01-22T19:25:47.528127 #27375]  INFO -- : WebOfScience::Harvester - processed author: 21
```

lib/web_of_science/queries.rb
```diff
@@ -12,89 +12,80 @@ class Queries
...
+    def initialize(database = 'WOK')
       @database = database
     end
 
     # @param uid [String] a WOS UID
-    # @return [WebOfScience::Records]
+    # @return [WebOfScience::Retriever]
     def cited_references(uid)
-      return empty_records if uid.blank?
+      raise(ArgumentError, 'uid must be a WOS-UID String') if uid.blank?
```
atz: 👍

lib/web_of_science/harvester.rb
```diff
@@ -15,6 +15,7 @@ def harvest(authors)
       authors.each do |author|
         process_author(author)
         author_success += 1
+        GC.start # suggest that now is a good time to clean up some memory
```
atz: I'm skeptical of adding explicit `GC` calls without concrete measurement that it helps.  As frequently as not, it can hurt.  `GC.start` is a process-blocking FULL SWEEP, not a "suggestion".  It is **guaranteed** to add latency every time, so the questions is "to what benefit?"

During debugging, you can interrogate the state w/ `GC.stat` to examine what the allocation stats are before/after running to see if you are really saving a significant amount.  More powerfully, you can try [`GC::Profiler`](http://ruby-doc.org/core-2.2.0/GC/Profiler.html).

I don't see anything we're doing here that makes us smarter than Ruby itself about the timing of collection.  Processing some authors will use a lot of memory and others comparably little.  So it seems weird to force Ruby's hand here.  AFAIK, our problem was that our process was **actually using** too much memory.  We have not determined that it was failing to recover assigned memory.  Please let me know if I'm missing something.

darrenleeweber: Understood, expected this to be a point of discussion.  I've read a little about `GC`, but I have never used it and do not understand it.  I added it in a couple places because it's also used in the SW harvester and didn't want to miss an opportunity to clean up memory (if possible).

peetucket: concur with Joe - I think we should hold off on doing an explicit GC for now unless we have evidence that it is helping (even though the current harvest is running fine with this code, the other changes are more likely responsible for this)

peetucket: Also of note, I have tried using the GC explicitly in the past with pre-assembly and after some experimentation so no difference in memory usage.

atz:
This plays havoc with the `DoiSearch` code that *just* merged and breaks the Pubmed code I have parallel to it in #698.

lib/doi_search.rb
```diff
@@ -12,17 +14,30 @@ def self.search(doi)
...
+  # Web of Science results are based on partial string matching for DOI queries, unfortunately.
+  # @see WebOfScience::Queries#search_by_doi
+  # @param [String] doi
+  # @return [Array<Hash>] matching hashes
+  def self.web_of_science(doi)
+    return [] unless Settings.WOS.enabled
+    doi = doi_name(doi)
+    return [] if doi.blank?
+    retriever = WebOfScience.queries.search_by_doi(doi)
+    retriever.next_batch.map { |record| add_citation(record.pub_hash) }
```
atz: All this, and we **still** don't get an Enumerator over all the records regardless of what `batch` they are in?  Why is this considered an improvement?

spec/lib/web_of_science/queries_spec.rb
```diff
@@ -27,99 +24,101 @@
...
+    wos_client = WebOfScience::Client.new('secret')
+    allow(WebOfScience).to receive(:client).and_return(wos_client)
+  end
+
   describe '#new' do
     it 'works' do
-      result = described_class.new(wos_client)
-      expect(result).to be_an described_class
+      instance = described_class.new
+      expect(instance).to be_an described_class
```
atz: This is literally testing "Does Ruby Work?".

PR #656, opened 2018-01-22
[WIP] PENDING INTEGRATION TESTING Send records count (via JSON) as proper integer, not a String

atz:
Also consolidates an unnecessary duplication of VCR cassette.


atz:
It *is* possible that this change could be disruptive, but there can't be much argument that the proper way to model (and express) a count is an Integer.  If there is a problem on their end, the fix would be pretty trivial.  Visibility to their codebase would help here.

peetucket:
Pending confirmation/integration testing with CAP/Profiles team

PR #670, opened 2018-01-24
WebOfScience::Contributions - tweaks

darrenleeweber:
Extracted from #644 
- minor changes to `WebOfScience::Contributions` methods
- let `Publication.where(wos_uid: uids).find_each` manage the work load
- refactor `contribution_by_identifier?` into two methods

### DIFF
--- a/lib/web_of_science/contributions.rb
+++ b/lib/web_of_science/contributions.rb
@@ -11,11 +11,9 @@ module WebOfScience
     # @return [Array<String>] uids that already have a contribution
     def author_contributions(author, uids)
       contrib_uids = []
-      uids.each_slice(50) do |batch_uids|
-        Publication.where(wos_uid: batch_uids).find_each do |pub|
-          contrib = find_or_create_contribution(author, pub)
-          contrib_uids << pub.wos_uid if contrib.persisted?
-        end
+      Publication.where(wos_uid: uids).find_each do |pub|
+        contrib = find_or_create_contribution(author, pub)
+        contrib_uids << pub.wos_uid if contrib.persisted?
       end
       contrib_uids
     end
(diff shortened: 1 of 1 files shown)


lib/web_of_science/contributions.rb
```diff
@@ -66,10 +64,19 @@ def create_contribution(author, contrib)
     # @param value [String]
     # @return [Boolean] contribution exists
     def contribution_by_identifier?(author, type, value)
+      contrib = contribution_by_identifier(author, type, value)
+      contrib.nil? ? false : contrib.persisted?
```
atz: `persisted?` when called here will always be `true`, since the object by definition came from `find_or_create...`.  

That means this method can shrink down to 1 line:
```
contribution_by_identifier(author, type, value) ? true : false
```
IMO, that suggests the method is unnecessary.  A quick look at all the callers:
```
lib/web_of_science/harvester.rb:51:      dois.reject! { |doi| contribution_by_identifier?(author, 'doi', doi) }
lib/web_of_science/harvester.rb:66:      pmids.reject! { |pmid| contribution_by_identifier?(author, 'pmid', pmid) }
lib/web_of_science/process_records.rb:75:          contribution_by_identifier?(author, 'WosUID', rec.uid) ||
lib/web_of_science/process_records.rb:76:            contribution_by_identifier?(author, 'WosItemID', rec.wos_item_id) ||
lib/web_of_science/process_records.rb:77:            contribution_by_identifier?(author, 'doi', rec.doi) ||
lib/web_of_science/process_records.rb:78:            contribution_by_identifier?(author, 'pmid', rec.pmid)
```
Literally nothing changes if they just start using `contribution_by_identifier` instead of `contribution_by_identifier?`, right?

It is sorta anti-pattern to have an interrogative (*`?`*) method do `create` operations as a side-effect.

darrenleeweber: `contribution_by_identifier(author, type, value)` can return `nil`

peetucket: for ruby, is nil == false  for comparison operations?

darrenleeweber: No (that's why rails adds in `.blank?` and `.present?`)
```
> nil == false
=> false
```
It's also a question of the method signature for return values.  In the first phase of the harvest operations, filtering the UIDs just needs a boolean response, it doesn't care about doing anything with a contribution (it just finds it or triggers the creation of one, if that's possible).

peetucket: i would be in favor of shrinking the method down to a single line, as Joe suggested, that's a style change

atz: As I said, in the context where it is used (including the filters), `nil` is indifferentiable from `false`.  In parlance, "nil isn't false, but it *is* falsey."  
```ruby
> [1,2,3,4].reject{|x| x.even? ? true : nil }
=> [1, 3]
> [1,2,3,4].reject{|x| x.even? ? true : false }
=> [1, 3]
```

atz: I would say if we ever have a method that can be wholesale substituted for another method with no difference in functionality, that argues against retaining both of them.  That's more composition than style, imo.

PR #674, opened 2018-01-24
More versatile factory, not a dozen bespoke single-use factories

atz:
Also remove a bunch of unused `let` statements.

Factories by default extrapolate URI from value.  Cases that need empty URI's are now obvious: they pass `identifier_uri: nil`.


atz:
No, this is a factory that I use in my **very next** PR/commit, that I could not have used otherwise.

Issue #673, opened 2018-01-24
WOS harvest efficiency


darrenleeweber:
Suggestions for improving the efficiency of WOS harvests

darrenleeweber:
## Use background jobs (e.g. use REDIS) for Author harvests

### Serial, batch retrieval from WOS API

- Run an author query to get back WOS-UIDs
  - get all the WOS-Links identifiers for all the UIDs
  - remove any UIDs that can be matched to author-contributions via PublicationIdentifiers
  - when matching PublicationIdentifiers - use only identifiers: UID, DOI, PMID (others?)
- For all UIDs that do not match any contributions
  - retrieve the full records and save them to a key:value store (REDIS)
  - use the author-id + the UID as a compound key
  - store the UID full record XML as the value
  - include the Links-API identifiers in there too
  - consider various alternatives for the key:value details

### Concurrent integration into the sul_pub database
- Run concurrent processes that consume from the REDIS key:value store
  - these concurrent processes should have no HTTP requests
    - they should never hit the WOS-API or the Links-API
    - one exception - they might need to queue or spawn some PubMed integration
 - retrieve any author-UID:xml (key:value) data and process it
   - see `WebOfScience::ProcessRecords`
   - refactor  `WebOfScience::ProcessRecords` to create `WebOfScience::ProcessRecord`
- once an author:UID record is processed, remove it from the queue (key:value store)
  - or consider keeping some cached data available for a short period of time?
  - a cache might be useful, because not all of the retrieved records are stored in the sul_pub db
  - this depends on whether a WOS-src-record is saved even if it matches existing publication data

atz:
I'm proceeding under the assumption of zero architectural revisions.  

### `ProcessRecords` 
- does (up to) 4 queries on the same table that could be done in one query, [here](https://github.com/sul-dlss/sul_pub/blob/master/lib/web_of_science/process_records.rb#L75-L78).

### `WebOfScience::Contributions` 
is full of anti-patterns.  `#create_contribution` does:
```ruby
contrib.save!
...
contrib.publication.pubhash_needs_update!
contrib.publication.save!
```
The initial save is unnecessary since `Publication` has an autosave on its associated objects.  

`#publication_identifier` does:
```ruby
 pub_ids = PublicationIdentifier.where(identifier_type: type, identifier_value: value)
 pub_ids.empty? ? nil : pub_ids.first
```
[...]

Issue #639, opened 2018-01-20
Replace SW-API for WOS-API in author harvest job


darrenleeweber:
See https://github.com/sul-dlss/sul_pub/blob/master/app/jobs/author_harvest_job.rb

```
$ git grep 'AuthorHarvestJob'
app/controllers/authors_controller.rb:7:    if AuthorHarvestJob.perform_later(author_params[:cap_profile_id], harvest_alternate_names: alt_names)
app/jobs/author_harvest_job.rb:1:class AuthorHarvestJob < ActiveJob::Base
app/jobs/author_harvest_job.rb:19:    msg = "AuthorHarvestJob.perform(cap_profile_id=#{cap_profile_id}, harvest_alternate_names=#{harvest_alternate_names})"
spec/controllers/authors_controller_spec.rb:17:      it 'enqueues an AuthorHarvestJob with an Author' do
spec/controllers/authors_controller_spec.rb:22:        end.to have_enqueued_job(AuthorHarvestJob).with(author.cap_profile_id.to_s, harvest_alternate_names: false)
spec/controllers/authors_controller_spec.rb:24:      it 'enqueues an AuthorHarvestJob with an Author and altNames parameter' do
spec/controllers/authors_controller_spec.rb:29:        end.to have_enqueued_job(AuthorHarvestJob).with(author.cap_profile_id.to_s, harvest_alternate_names: true)
```

(Also, what backend jobs system is this using?)

peetucket:
This is invoked as an API endpoint by CAP.

darrenleeweber:
There are no specs on the existing sw-job, so looking into creating them, e.g.
- https://medium.com/@chuckjhardy/testing-rails-activejob-with-rspec-5c3de1a64b66

PR #677, opened 2018-01-25
WebOfScience::ProcessRecord

darrenleeweber:
Connected to #639, #673 

This class could be used by anything that needs to process only one WOS-record, e.g. an API request for one record or a background jobs process for async processing.

This opens up some follow-on possibilities.  This class could be used inside the WOS-process-records class (but, if so, it might forgo some enumerator optimizations there; not sure).  This class could be attached directly to the WOS-record class as in `WebOfScience::Record#process` or something.

### DIFF
--- a/lib/web_of_science/process_record.rb
+++ b/lib/web_of_science/process_record.rb
@@ -1,0 +1,110 @@
+module WebOfScience
+
+  # This class complements the WebOfScience::Harvester
+  class ProcessRecord
+    include WebOfScience::Contributions
+
+    # @param [Author] author
+    # @param [WebOfScience::Record] record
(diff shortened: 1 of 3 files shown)


peetucket:
so this new class isn't currently used by the harvester or other processes?

darrenleeweber:
As noted above, I expect to use this for single-record "harvests" and/or explore using this in an async background job.  The uses of this are various, but can't be explored until this PR is reviewed/merged/available.

peetucket:
Ok ,since the next step is actual production use, then I'm in -- I more reluctant to include experimental code in master, seems like keeping that in a branch until the experiment pays off makes more sense.

PR #678, opened 2018-01-25
DOI module

darrenleeweber:
fix #675

PR #680, opened 2018-01-26
DoiSearch against WOS

atz:
- honors configs
- for wos, ensures FULL doi string match returned if available

I don't care about "cognitive complexity" for this method at the moment.  The two things I anticipate doing to resolve that are:

1. Eventually ripping out the ScienceWire code, obviously
2. Moving the DOI full match stuff deeper into the stack, e.g. WOS `Queries`

Given the decent test coverage and temporary nature, I'm willing to punt on the CoCliCogCom, for now.

### DIFF
--- a/lib/doi_search.rb
+++ b/lib/doi_search.rb
@@ -1,15 +1,53 @@
 class DoiSearch
+  # The first authoritative result(s) are returned.
+  # If an authoritative local match is found, no remote services are hit.
+  # WoS hits need to be post-processed because the API does partial string
+  # matching for DOI queries, unfortunately.
+  # @param [String] doi
+  # @return [Array<Hash>] matching hashes
   def self.search(doi)
-    results = [Publication.find_by_doi(doi)].compact
-    results += ScienceWireClient.new.get_pub_by_doi(doi) if results.none?(&:authoritative_doi_source?)
-    results.reject! { |pub| non_sw_pub pub } if results.size > 1
+    pub = Publication.find_by_doi(doi)
+    return [pub.pub_hash] if pub && pub.authoritative_doi_source?
+    if Settings.SCIENCEWIRE.enabled
+      sw_hits = ScienceWireClient.new.get_pub_by_doi(doi)
+      return sw_hits unless sw_hits.empty?
+    end
+    results = []
+    if Settings.WOS.enabled
+      results += WebOfScience.queries.search_by_doi(doi).map { |rec| add_citation(rec.pub_hash) }
+      full_match = normalized_dois(results)[normalized_doi(doi)]
+      return [full_match] if full_match
+    end
+    # Everything else is non-authoritative
+    results.unshift(pub.pub_hash) if pub
     results
   end
 
-  # @return [Boolean] true if the pub is not from sciencewire
-  def self.non_sw_pub(pub)
-    return false if pub.is_a? Hash # Hashes are returned from the SW only query
-    !pub.authoritative_doi_source?
+  # @param [Hash] pub_hash modifies passed hash to include citation k/v pairs
+  # @return [Hash] the same modified hash
+  def self.add_citation(pub_hash)
+    cite = Csl::Citation.new(pub_hash)
+    pub_hash[:apa_citation] ||= cite.to_apa_citation
+    pub_hash[:mla_citation] ||= cite.to_mla_citation
+    pub_hash[:chicago_citation] ||= cite.to_chicago_citation
+    pub_hash
   end
-  private_class_method :non_sw_pub
+  private_class_method :add_citation
+
+  # @param [Array<Hash>] pub_hashes WOS pub_hashes
+  # @return [Hash<String => Hash>] normalized DOI strings mapped to pub_hashes
+  # pub_hash[:identifier].select{|h| h[:type] == 'doi'}}}
+  # [{:type=>"doi", :id=>"10.1172/jci.insight.87623", :url=>"https://dx.doi.org/10.1172/jci.insight.87623"}]
+  def self.normalized_dois(pub_hashes)
+    Hash[pub_hashes.map { |h| normalized_doi(h[:doi]) }.zip(pub_hashes)]
+      .select { |k| k } # exclude nil key(s) (normalization fail)
+  end
+  private_class_method :normalized_dois
+
+  # @param [String, nil] doi
+  # @return [String, nil] normalized DOI string
+  def self.normalized_doi(doi)
+    Identifiers::DOI.extract(doi).first
+  end
+  private_class_method :normalized_doi
 end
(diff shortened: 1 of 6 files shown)


darrenleeweber:
This PR is punting on https://github.com/sul-dlss/sul_pub/issues/681

lib/doi_search.rb
```diff
@@ -1,15 +1,41 @@
...
+    results = []
+    if Settings.WOS.enabled
+      results += WebOfScience.queries.search_by_doi(doi)
+      full_match = normalized_dois(results)[normalized_doi(doi)]
+      return [full_match] if full_match
+    end
+    # Everything else is non-authoritative
+    results.unshift(pub) if pub
     results
   end
```
darrenleeweber: :+1: for use of Settings
- the cognitive complexity has gone up :hankey: but punting on this for now
ln 9-10 - seems to behave the same as the old code
ln 10 - if pub is not authoritative (manual pub), the pub hangs around for the rest of the method (just noting behavior)
ln 11-14 makes sense although it's not obvious that `sw_hits` is an array (the old code ensured it and make it clearer)
ln 15-20 and called methods are complex, but OK;  :+1: for `normalized_doi`
  - the Hash construction is complex and hits all the results (it's difficult to read), maybe an `Array.find` would be easier to read? e.g. `results.find {|i| normalized_doi(i) == normalized_doi(doi) }` but do the `normalized_doi(doi)` first (not every time);  assuming the results are not very large, the trade-off for easier to read might be more important than performance of Hash constuct vs. the array find (?)

atz: I had Aaron spot-check the `Hash[]` constructor line asking if it was readable, and he was fine w/ it.

spec/lib/doi_search_spec.rb
```diff
@@ -1,36 +1,101 @@
...
   let(:doi_value) { '10.1016/j.mcn.2012.03.007' }
-  let(:doi_identifier) do
-    FactoryBot.create(:publication_identifier,
-                       identifier_type: 'doi',
-                       identifier_value: doi_value,
-                       identifier_uri: "http://dx.doi.org/#{doi_value}")
-  end
+  let(:doi_identifier) { create(:doi_pub_id, identifier_value: doi_value) }
+  let(:wos_html) { File.read('spec/fixtures/wos_client/medline_encoded_records.html') }
+  let(:wos_records) { WebOfScience::Records.new(encoded_records: wos_html) }
```
darrenleeweber: After the factory changes, it's no longer obvious what the `doi_identifier` contains; the `identifier_uri` is now an "under-the-table" construct (if I understand the changes correctly).

spec/lib/doi_search_spec.rb
```diff
@@ -1,36 +1,101 @@
...
-        doi_identifier.publication.save
-        expect(ScienceWireClient).to receive(:new).and_call_original
-        result = DoiSearch.search doi_value
-        expect(result.size).to eq 1
-        expect(result.first[:provenance]).to eq 'sciencewire'
+    context 'ScienceWire and WOS both disabled' do
+      it 'only searches locally, returning Array of one Publication' do
+        expect(ScienceWireClient).not_to receive(:new)
+        expect(WebOfScience).not_to receive(:queries)
+        expect(described_class.search(doi_value)).to eq([doi_identifier.publication])
```
darrenleeweber: `described_class.search(doi_value)` -> `result` ?

spec/lib/doi_search_spec.rb
atz: Yes, the default is closer to a "good" object, which is the more typical case for consumers of the factory.

spec/lib/doi_search_spec.rb
atz: `.search` returns an array.  (In this test, an array w/ 1 element)

lib/doi_search.rb
atz: @darrenleeweber Yes, you're right about the method signature, will update that to include `WebOfScience::Record`

atz: Line 22 is informed by the above comment.  If we reached the bottom of the method, we have no authoritative matches.  We might have `pub` and might have some non-exact WOS matches.  The response will include both when available, with the local one first.

spec/lib/doi_search_spec.rb
darrenleeweber: - ln 15:  `subject(:result) { described_class.search(doi_value) }`
- ln 27:  `described_class.search(doi_value)`

So, looks like ln 27 can be:
- `expect(result).to eq([doi_identifier.publication])`

spec/lib/doi_search_spec.rb
```diff
@@ -1,36 +1,101 @@
...
+          allow(publication).to receive(:wos_pub?).and_return(false)
+          expect(WebOfScience.queries).to receive(:search_by_doi).with(doi_value).and_return([])
+          described_class.search(doi_value)
+        end
+      end
+
+      it 'without a local hit, queries WOS' do
+        expect(WebOfScience.queries).to receive(:search_by_doi).with(other_doi).and_return([])
+        expect(described_class.search(other_doi)).to eq []
       end
```
darrenleeweber: Looks like ln 59, 64 & 70 might also use a `result` subject?

spec/lib/doi_search_spec.rb
```diff
@@ -1,36 +1,101 @@
...
+  describe '.normalized_dois' do
+    subject(:result) { described_class.send(:normalized_dois, wos_records) }
+
+    it 'returns a hash with only authoritative DOI keys mapped to records' do
+      expect(wos_records.count).to eq 5 # the fixture has many records
+      expect(result).to match a_hash_including(other_doi => WebOfScience::Record)
+      expect(result.count).to eq 1
+      expect(result[other_doi].doi).to eq other_doi
+    end
+  end
```
darrenleeweber: Not required for this PR, just thoughts.  Started to think that this method might be useful as `WebOfScience::Records#normalized_dois`  (i.e. move-method refactor), if the only param it can take is a `WebOfScience::Records` (or an Enumerator that contains only items that respond to `:doi`).  A related idea is `WebOfScience::Identifiers#normalized_doi`.  This is kinda motivation for #678 as a possible owner of such things.

atz:
I disagree.  The ScienceWire behavior would need to change to provide that level of consistency w/ what we need from WoS, and I am uninterested in making updates to code we intend to rip out rather soon.  SW code did not do normalization before, and not adding it now is **intentional**.  

The entire motivation of feature-flag configuration is to keep the old implementation as unchanged as possible while adding the new behavior beside it.  Your argument would make sense if we intended to preserve both backends in parallel, but that is not our strategy.

darrenleeweber:
We also need to be aware that work in https://github.com/sul-dlss/sul_pub/pull/644 might break this work even if it doesn't conflict with it in terms of any code overlap.

PR #696, opened 2018-01-30
WOS harvest progress

darrenleeweber:
Small PR to add logging progress of authors harvested.
- use a relation for the query
- progress is logged at the completion of each batch

### DIFF
--- a/lib/harvester/base.rb
+++ b/lib/harvester/base.rb
@@ -4,9 +4,13 @@ module Harvester
   class Base
     # @return [void]
     def harvest_all
-      Author.where(active_in_cap: true, cap_import_enabled: true)
-            .find_in_batches(batch_size: batch_size)
-            .each { |batch| harvest(batch) }
+      total = authors_query.count
+      count = 0
+      authors_query.find_in_batches(batch_size: batch_size).each do |batch|
+        harvest(batch)
+        count += batch_size
+        logger.info "completed #{count} of #{total} authors for harvest"
+      end
     end
 
     # @param [Enumerable<Author>] _authors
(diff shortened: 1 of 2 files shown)


atz:
What information are you trying to convey?  Why is 
```
completed 0% of all authors for harvest
```
more useful than
```
completed 15 authors of 32101
```
Shouldn't the count be based on `author_success` or `authors.count`?

The `authors` enumerator could encapsulate any logic.  E.G.
- `Author.where(official_last_name: 'Mitchell')`
- `Author.where(created_at > '2017-06-01')`
- `Author.limit(1000).offset(20000)`
- `Author.last`

In none of those cases does the log line have any relationship to what just executed.  The latter would tell us it completed 100% of all authors when that isn't remotely true.

darrenleeweber:
- the `Harvester::Base` controls what authors get harvested, so the metrics are in that base class
  - the base class is the only place that knows anything about the entire collection of authors for harvests
- a % complete is a reasonable summary metric
  - yes, others are possible but not the intention of this PR to be the swiss-army knife of harvest metrics
  - the % complete is not meant to represent % successful completions
- if some additional metrics should be added to this PR, let's define and agree on those requirements
  - I put this together quickly only so it doesn't need to be done manually every time at a console

peetucket:
seems logical to include the total running and where I am in there, as well as percentage (helpful to know if 80% done means I'll be done in 10 minutes or 10 hours, which knowing the total authors running would help with)

darrenleeweber:
The revision pushed relocates the overall progress logger to the base class, which is the only place that can track the overall progress.  It might not have access to a logger, so it has to check for one first.

atz:
You are 100% missing the substance of my objection.

atz:
Apparently `find_in_batches` applies its own order to accomplish chunking/pagination, so you do get rails internally ordained order, and specifically **cannot** apply any other: http://api.rubyonrails.org/classes/ActiveRecord/Batches.html#method-i-find_in_batches
```
NOTE: It's not possible to set the order. 
That is automatically set to ascending on the primary key (“id ASC”) 
to make the batch ordering consistent. 
```

darrenleeweber:
For this PR, it is convenient that `find_in_batches` imposes the order that is desired - asc on Author.id is what is required.

atz:
I can't ever support the idea that the base class is specializing itself and adding so much cruft just to apply logging logic that only satisfies one case.  That isn't how base classes are supposed to operate.  When I look at the file before and the file after, it is clear I would always prefer to inherit (both legacy and OO-wise) the former.

Issue #693, opened 2018-01-30
Duplicate values in Publication.pmid


atz:
```ruby
Publication.select(:id,:pmid).where.not(pmid: nil).group(:pmid).having("count(*)>1").to_a.count
=> 187
 Publication.select(:id,:pmid).where.not(pmid: nil).group(:pmid).having("count(*)>1")
=> [#<Publication:0x00000015974158 id: 122975, pmid: 22806268>,
 #<Publication:0x0000001596ff40 id: 124711, pmid: 22961735>,
 #<Publication:0x0000001596fd88 id: 285164, pmid: 23163767>,
 #<Publication:0x0000001596fc48 id: 128469, pmid: 23313911>,
 #<Publication:0x0000001596fae0 id: 128715, pmid: 23339536>,
 #<Publication:0x0000001596f978 id: 131244, pmid: 23588301>,
 #<Publication:0x0000001596f810 id: 132251, pmid: 23697924>,
 #<Publication:0x0000001596f680 id: 132337, pmid: 23708745>,
 #<Publication:0x0000001596f4f0 id: 266714, pmid: 23736837>,
 #<Publication:0x0000001596f388 id: 132632, pmid: 23745906>,
 #<Publication:0x0000001596f248 id: 266724, pmid: 23747482>,
 #<Publication:0x0000001596f090 id: 283571, pmid: 23764482>,
 #<Publication:0x0000001596eed8 id: 266908, pmid: 23764571>,
 #<Publication:0x0000001596ed48 id: 218758, pmid: 23768493>,
 #<Publication:0x0000001596ebb8 id: 241598, pmid: 23770727>,
 #<Publication:0x0000001596ea28 id: 260447, pmid: 23799945>,
 #<Publication:0x0000001596e8c0 id: 266907, pmid: 23801384>,
 #<Publication:0x0000001596e758 id: 242120, pmid: 23840927>,
 #<Publication:0x0000001596e5f0 id: 242156, pmid: 23845225>,
 #<Publication:0x0000001596e410 id: 260015, pmid: 23859315>,
 #<Publication:0x0000001596e280 id: 242295, pmid: 23867624>,
 #<Publication:0x0000001596e140 id: 242330, pmid: 23872309>,
 #<Publication:0x0000001596dfb0 id: 273972, pmid: 23872676>,
 #<Publication:0x0000001596ddf8 id: 242434, pmid: 23884931>,
[...]

PR #698, opened 2018-01-31
Pubmed from WOS

### LINKED ISSUE #690: Update publication API to use WoS client for searching by PMID

atz:
Fix #690 

- Broke out the remote searching to a private_class_method.
- Changed the expectations w/ regards to local hits based on analysis of actual duplicate values in #693. No reason to present the user with indifferentiable duplicate records.
- Honor configs

100% disinterested in Code Climate Cognitive Complexity for the same reasons as before, already anticipating removing SW code, which would make it pretty basic.  This PR reduces the flagged complexity from 13 to 8.

### DIFF
--- a/app/models/publication.rb
+++ b/app/models/publication.rb
@@ -240,7 +240,7 @@ class Publication < ActiveRecord::Base
   end
 
   def authoritative_pmid_source?
-    pubmed_pub? || sciencewire_pub?
+    pubmed_pub? || sciencewire_pub? || wos_pub?
   end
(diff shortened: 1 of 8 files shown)


darrenleeweber:
The large fixture diff on this PR might be reduced if some existing fixtures are used for requests etc.  See the specs in https://github.com/sul-dlss/sul_pub/blob/master/spec/lib/web_of_science/process_pubmed_spec.rb for examples.

PR #699, opened 2018-01-31
[WIP] WebOfScience::QueryAuthor - use settings to modify the query time span

### LINKED ISSUE #663: Investigate and implement use of "symbolicTimeSpan" parameter when harvesting to limit results
Page 20 of the Web of Science API Documentatoin (PDF) lists the "symbolicTimeSpan" parameter, with options of 1week, 2week, and 4week (also shown at http://ipscience-help.thomsonreuters.com/wosWebServicesExpanded/WebServiceOperationsGroup/WSPremiumOperations/wokSearchGroup/search.html). Rob at Clarivate confirms this restricts records returned based on database load times (not publication dates), which is what we'd want for an incremental update.  Can we use the 4week parameter to reduce the run times for future harvest runs?  How would this work (two rake tasks, one for an initial run, one for normally scheduled bi-weekly runs)

darrenleeweber:
### Update
- work on this branch is getting extracted in other PRs
- an alternative implementation is in progress in #711 

Fix #663

- this applies the spirit of #663, if not the letter.
- had to add specs for this class, which was otherwise exercised by the harvester specs
- added rake tasks so the suite of tasks includes:
  - harvest all authors, for the past 30 days
  - harvest all authors, for all time

## Implementation Notes
- this applies a specific setting to change the time span only for author search queries
  - the scope of the issue does not require surfacing a generic time-span option for all queries
  - for most queries, the encapsulated time span in WOS-queries is the appropriate one to use
- as the spec fixture indicates, when using this PR, the records matched for Russ Altman decrease from 400+ matches to only 2 for the period of Jan 2018; this will have the desired effect of substantially speeding up the scheduled harvests for authors

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -50,6 +50,7 @@ WOS:
     - WOS
     - MEDLINE
   AUTH_CODE: secret
+  AUTHOR_UPDATE: 30 # days ago
   LOG: log/web_of_science.log
   LOG_LEVEL: warn
 
--- a/lib/tasks/wos.rake
+++ b/lib/tasks/wos.rake
@@ -1,6 +1,13 @@
 namespace :wos do
-  desc 'Harvest from Web of Science, for all authors'
+  desc 'Harvest from Web of Science, for all authors, for all time'
   task harvest_authors: :environment do
+    Settings.WOS.AUTHOR_UPDATE = 0 # harvest publications from the beginning of time
+    WebOfScience.harvester.harvest_all
+  end
+
+  desc 'Harvest monthly updates from Web of Science, for all authors'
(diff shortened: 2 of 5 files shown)


darrenleeweber:
RE 2.  "Why not something at the level of the harvester that can take in a parameter at initialization (defaulting to all time)?"

- exposing a time-span parameter at this level requires changes to the `Harvester::Base` class, which is likely to be a source of conflict with other harvesters
- if the update param is specific to the WOS-harvester, then it might no longer inherit from the base class
- the use of WOS-specific settings allows application of the setting only in the specific place it is required

darrenleeweber:
This work has moved to #711

PR #702, opened 2018-01-31
skip SW harvest in cap poller when not enabled

### LINKED ISSUE #700: Skip SW harvesting as a result of CAP poller updating when disabled
Do not harvest for SW new authors unless SW is enabled

see https://github.com/sul-dlss/sul_pub/blob/master/lib/cap/authors_poller.rb#L58 to change  analogous to https://github.com/sul-dlss/sul_pub/blob/master/lib/cap/authors_poller.rb#L63

peetucket:
fixes #700

### DIFF
--- a/spec/lib/cap/authors_poller_spec.rb
+++ b/spec/lib/cap/authors_poller_spec.rb
@@ -223,4 +223,21 @@ describe Cap::AuthorsPoller, :vcr do
       expect { subject.process_next_batch_of_authorship_data(bogus: 'data') }.to raise_error(Net::HTTPBadResponse)
     end
   end
+
+  describe '.do_science_wire_harvest' do
+    it 'does nothing if sw client is disabled' do
+      allow(Settings.SCIENCEWIRE).to receive(:enabled).and_return(false)
+      expect(ScienceWireHarvester).not_to receive(:harvest_pubs_for_author_ids)
+      subject.do_science_wire_harvest
+    end
+  end
+
+  describe '.do_wos_harvest' do
+    it 'does nothing if WoS client is disabled' do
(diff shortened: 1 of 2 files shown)


spec/lib/cap/authors_poller_spec.rb
```diff
@@ -223,4 +223,25 @@
...
+  describe '.do_wos_harvest' do
+    it 'does nothing if WoS client is disabled' do
+      current_wos_setting = Settings.WOS.enabled
+      Settings.WOS.enabled = false
+      subject.do_wos_harvest
+      expect(Author).not_to receive(:where)
+      expect(WebOfScience.harvester).not_to receive(:harvest)
+      Settings.WOS.enabled = current_wos_setting
+    end
+  end
```
darrenleeweber: :+1: to adding specs for this :balloon: 

Consider using `allow(Settings.WOS).to receive(:enabled).and_return(true)` (or false) to mock setting these values.

darrenleeweber: It might simplify to `expect(WebOfScience).not_to receive(:harvester)`

darrenleeweber: I usually put an `expect` before calling something to ensure it's mocked, e.g.
```
allow(Settings.WOS).to receive(:enabled).and_return(false)
expect(WebOfScience).not_to receive(:harvester)
subject.do_wos_harvest
```

peetucket: thanks for the suggestions, the mocking of settings is a much better way of doing that... all made, rebased and pushed

PR #703, opened 2018-01-31
WOS harvest job

### LINKED ISSUE #639: Replace SW-API for WOS-API in author harvest job

darrenleeweber:
Fix #639 

- added active job specs
  - consolidated common author details into a factory
- extract method refactor for the sciencewire harvest
  - use settings to enable/disable it
- add harvest method for the WOS harvester
  - extract method refactor for logging publication info

### DIFF
--- a/app/jobs/author_harvest_job.rb
+++ b/app/jobs/author_harvest_job.rb
@@ -1,24 +1,31 @@
 class AuthorHarvestJob < ActiveJob::Base
   queue_as :default
 
-  ##
   # Performs an asynchronous harvest and save for a given Author
   # @param [String] cap_profile_id
   # @param [Boolean] harvest_alternate_names
   def perform(cap_profile_id, harvest_alternate_names: false)
-    harvester = ScienceWireHarvester.new
-    harvester.use_author_identities = harvest_alternate_names
-    author = Author.where(cap_profile_id: cap_profile_id).first
+    author = Author.find_by(cap_profile_id: cap_profile_id)
     author ||= Author.fetch_from_cap_and_create(cap_profile_id)
-    harvester.harvest_pubs_for_author_ids author.id
-    pubs = Contribution.where(author_id: author.id).map(&:publication).each do |p|
-      logger.info "publication #{p.id}: #{p.pub_hash[:apa_citation]}"
-    end
-    logger.info "Number of publications #{pubs.count}"
+    raise "Could not find or fetch author: #{cap_profile_id}" unless author.is_a?(Author)
+    sciencewire(author, harvest_alternate_names)
   rescue => e
     msg = "AuthorHarvestJob.perform(cap_profile_id=#{cap_profile_id}, harvest_alternate_names=#{harvest_alternate_names})"
     NotificationManager.log_exception(logger, msg, e)
     Honeybadger.notify(e, context: { message: msg })
     raise
   end
+
+  private
+
+    def sciencewire(author, harvest_alternate_names)
+      return unless Settings.SCIENCEWIRE.enabled
+      harvester = ScienceWireHarvester.new
+      harvester.use_author_identities = harvest_alternate_names
+      harvester.harvest_pubs_for_author_ids author.id
+      pubs = Contribution.where(author_id: author.id).map(&:publication).each do |p|
+        logger.info "publication #{p.id}: #{p.pub_hash[:apa_citation]}"
+      end
+      logger.info "Number of publications #{pubs.count}"
+    end
 end
--- a/app/jobs/author_harvest_job.rb
+++ b/app/jobs/author_harvest_job.rb
@@ -4,13 +4,15 @@ class AuthorHarvestJob < ActiveJob::Base
   # Performs an asynchronous harvest and save for a given Author
   # @param [String] cap_profile_id
   # @param [Boolean] harvest_alternate_names
+  # @return [void]
   def perform(cap_profile_id, harvest_alternate_names: false)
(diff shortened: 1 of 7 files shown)


app/jobs/author_harvest_job.rb
```diff
@@ -1,24 +1,52 @@
...
-    harvester.harvest_pubs_for_author_ids author.id
-    pubs = Contribution.where(author_id: author.id).map(&:publication).each do |p|
-      logger.info "publication #{p.id}: #{p.pub_hash[:apa_citation]}"
-    end
-    logger.info "Number of publications #{pubs.count}"
+    raise "Could not find or fetch author: #{cap_profile_id}" unless author.is_a?(Author)
+    sciencewire(author, harvest_alternate_names)
+    web_of_science(author, harvest_alternate_names)
   rescue => e
     msg = "AuthorHarvestJob.perform(cap_profile_id=#{cap_profile_id}, harvest_alternate_names=#{harvest_alternate_names})"
```
atz: I know you didn't touch this line, but I think the log message intends to have `harvest_alternate_names:`, not `harvest_alternate_names=`...

spec/factories/author.rb
```diff
@@ -1,12 +1,19 @@
 FactoryBot.define do
   sequence(:random_id) do |n|
-    @random_ids ||= (10_000..1_000_000).to_a.shuffle
+    @random_ids ||= Array.new(10_000) { SecureRandom.random_number(990_000) + 10_000 }
```
atz: This factory is ridiculous.  It generates 10,000 random numbers to give back one.  Previously, it initialized a `990,000` member `Array` and shuffled it, which is **worse**!  The most authors our tests ever use is `5`.  

Why not just generate the ids on demand (per `n`), record each generated ids in a Hash (`@random_ids`) and regenerate if there is ever collision?  The only downside of this approach would be when the namespace started to fill up and we had collisions, but clearly our usage will never approach that!

spec/factories/author.rb
```diff
@@ -32,4 +39,37 @@
       end
     end
   end
+
+  # Public data from
+  # - https://stanfordwho.stanford.edu
+  # - https://med.stanford.edu/profiles/russ-altman
+  factory :russ_altman, parent: :author do
```
atz: Does this mean `author.local.rb` can go away?

app/jobs/author_harvest_job.rb
darrenleeweber: Yea, it's a bit weird for a log message to appear to construct the call, as if you could cop-n-pasta to run it again.  If that's the intention, it _could_ be fixed up.

spec/factories/author.rb
darrenleeweber: It's all public information; if it's also in that local file, yea, will throw that out in this commit.

spec/factories/author.rb
darrenleeweber: The only question with adopting a Hash solution is whether that could interfere with the design of the `sequence` method.  Originally, this was a cop-n-pasta solution and it works, although I also don't love it for these reasons ^^.  I'll do a bit more homework on that `sequence` method and see if a Hash solution works out OK.

spec/factories/author.rb
darrenleeweber: Turns out the local version adds a ton of publications that are specific to SW.  Until we remove SW, punting on removing that data.  This PR allows the public information to be used generally in specs and CI.

spec/factories/author.rb
```diff
@@ -1,12 +1,33 @@
 FactoryBot.define do
   sequence(:random_id) do |n|
-    @random_ids ||= (10_000..1_000_000).to_a.shuffle
-    @random_ids[n]
+    @random_ids ||= {}
+    @random_ids[n] ||= begin
+      x = nil
+      while x.nil?
+        i = SecureRandom.random_number(990_000) + 10_000
+        x = i unless @random_ids.values.include?(i)
```
atz: This is a more reasonable approach, why not just go ahead and just put the value in as the hash as the key.  That's all we care about, not `n`.

spec/factories/author.rb
atz: But now, the only thing we are keeping "private" is a bunch of ID numbers?  I'm not saying they need to be the same factory, but I don't see any privacy concern over a few hundred numerals.

PR #708, opened 2018-02-01
Harvester::Base - allow options

darrenleeweber:
Open the harvest base class to harvesting options.  The base class is agnostic to any details of the options and uses none by default.


atz:
This is weird and unnecessary.  If the concrete class needs options, it can have them (in an initializer) without requiring the base class to pass around empty arguments it literally never cares about.

PR #709, opened 2018-02-01
update schedule for harvesting on -qa and -prod to further separate a…

peetucket:
…nd reduce likely of overlap

### DIFF
--- a/config/schedule.rb
+++ b/config/schedule.rb
@@ -6,8 +6,8 @@ end
 
 set :output, 'log/cron.log'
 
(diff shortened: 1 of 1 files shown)


darrenleeweber:
The WOS harvest on sul-pub-dev was about 25% done after about 4-5 days
- approx. extrapolation is completion in 20 days
- if we have a monthly harvest on two machines, they will overlap (if they are doing a full harvest process)
- we don’t yet know the speed of an update harvest (it's likely to be much faster)

peetucket:
Let's leave this open until we know what the update harvest will be - since the update is the only one that will actually be scheduled via cron (a full will only be run manually presumably)

PR #711, opened 2018-02-01
WOS harvest authors using symbolic time span

### LINKED ISSUE #663: Investigate and implement use of "symbolicTimeSpan" parameter when harvesting to limit results

darrenleeweber:
Fix #663

Using a symbolicTimeSpan of 4 weeks, the results for Russ Altman drop from 471 to 16.

This is an alternative implementation to #699 
- #699 is closed
- this applies the letter of #663
- added specs for this class, which was otherwise exercised by the harvester specs
- added a rake task so the suite of tasks will include:
  - harvest all authors, for the past 4 weeks
  - harvest all authors, for all time
  - harvest one author, for all time


lib/web_of_science/query_author.rb
```diff
@@ -34,10 +34,38 @@ def uids
...
+      #     If timeSpan and symbolicTimeSpan are both omitted, then the maximum publication date time span
+      #     will be inferred from the editions data.
+      #
+      #     The valid values are strings: '1week', '2week', '4week' (prior to today)
+      #
+      # @param [Hash] params from queries.params_for_fields
+      # @return [void]
+      def update_time_span(params)
+        return if options[:update].blank?
+        params[:queryParameters].delete(:timeSpan)
```
peetucket: may want to put an inline comment here indicating why this is done for the benefit of future developers (i.e. WoS documentation states you can't have both timeSpan and symbolicTimeSpan)

spec/lib/web_of_science/query_author_spec.rb
```diff
@@ -0,0 +1,121 @@
...
+                       first_name: 'Russ',
+                       middle_name: nil,
+                       last_name: 'Altman',
+                       email: nil,
+                       institution: nil)
+    author
+  end
+
+  it 'works' do
+    expect(query_author).to be_a described_class
```
atz: These kind of tests never make any sense.  In Ruby is impossible to call `AnyClass.new` and get back anything but an instance of `AnyClass`.  The only alternative is that initialization throws an exception.  If the intent is to test against that, it should be:
```
expect { described_class.new }.not_to raise_error
```

Otherwise it is just testing "Is Ruby Still Working?"

spec/lib/web_of_science/query_author_spec.rb
```diff
@@ -0,0 +1,121 @@
...
+    it 'contains author institution' do
+      expect(params[:userQuery]).to include(institution)
+    end
+
+    it 'uses a default timeSpan when update option is not set' do
+      expect(params[:timeSpan]).to include(begin: String, end: String)
+    end
+
+    it 'uses options to set a symbolicTimeSpan' do
+      query_author = described_class.new(author, update: true)
```
atz: It is bad form to reassign a variable over the already declared `subject` name.

spec/lib/web_of_science/query_author_spec.rb
```diff
@@ -0,0 +1,121 @@
...
+    let(:query) { query_author.send(:author_query) }
+    let(:params) { query[:queryParameters] }
+
+    it 'contains query parameters' do
+      expect(query).to include(queryParameters: Hash)
+      expect(params).to include(databaseId: String, userQuery: String, timeSpan: Hash, queryLanguage: String)
+    end
+
+    it 'contains author name' do
+      expect(params[:userQuery]).to include(names)
```
atz: These 3 tests can be combined into a single `includes` expectation.

lib/web_of_science/query_author.rb
darrenleeweber: That documentation is all there in the method docs already.

spec/lib/web_of_science/query_author_spec.rb
darrenleeweber: Sort of.  But if the initializer fails by throwing an exception (e.g. arg-exception), then the spec fails.  FYI, I use the rspec-kickstarter gem often to auto-populate specs on a new class and then rework the auto-generated specs a little.  These specs often come from rspec-kickstarter, so if they think it's a useful pattern, that's OK by me too.

spec/lib/web_of_science/query_author_spec.rb
darrenleeweber: OK, yea, I was wondering about whether to declare a new context for this.  I'll do that.

spec/lib/web_of_science/query_author_spec.rb
darrenleeweber: I don't care to make this change, it's fine as it is.  I prefer to get back more specific information about what is failing.  Since the userQuery is a String with various moving parts, separate specs are useful.

spec/lib/web_of_science/query_author_spec.rb
atz: Right, so if you **intend** to test whether the initializer fails, then I gave you the correct form to use to make that explicit.  Instead you are testing a ruby guarantee and indirectly seeing if it explodes.

lib/web_of_science/query_author.rb
peetucket: ok, missed it

lib/web_of_science/query_author.rb
```diff
@@ -34,10 +34,46 @@ def uids
...
+          userQuery: query[:userQuery],
+          symbolicTimeSpan: '4week',
+          queryLanguage: query[:queryLanguage]
+        }
+      end
+
+      # @param [Hash] params from queries.params_for_fields
+      # @return [void]
+      def update_user_query(params)
+        params[:queryParameters][:userQuery] = "AU=(#{names}) AND AD=(#{institution})"
```
atz: Breaking this one-liner out to a separate method doesn't seem to add anything.  To me, there is less clarity because the model of "mutate a deep Hash" is not a good way to define behavior.  Ready to merge this otherwise.

darrenleeweber: A possible refactor of all that is to encapsulate the query params into a class of it's own - most of it was coded as private details on the WOS-query, assuming all of it could remain private.  But, it has turned out that surfacing access to param details from the WOS-query has been useful.  A WOS-params class is a possibility, although punted so far.

darrenleeweber:
Fixed all substantial changes requested.

lib/web_of_science/query_author.rb
```diff
@@ -35,9 +35,39 @@ def uids
...
+        # The Savon gem and WOS SOAP API are _very_ sensitive to the order of the params in the Hash; it must be
+        # reconstructed in the correct order - do not try conventional ruby Hash updates here.  This makes no sense
+        # to me [DLW], but learning from experience indicates the following works OK:
+        query = params[:queryParameters].dup
+        params[:queryParameters] = {
+          databaseId: query[:databaseId],
+          userQuery: query[:userQuery],
+          symbolicTimeSpan: '4week',
+          queryLanguage: query[:queryLanguage]
+        }
```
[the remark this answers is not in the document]
darrenleeweber: Yes, absoluely, we're on the same page here, almost.  As discussed in standup, I share the same concerns and this method was originally as requested.  As noted in the comments and in standup, Savon queries behave strangely when adding a new field to a query hash and I do not understand why.  The SOAP API responded with errors when the requested code was used.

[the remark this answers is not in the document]
atz: I think my PR against this branch is now viable and a pretty simple fix.

PR #712, opened 2018-02-01
Consolidation and cleanup around `contribution_by_identifier?`

atz:
This reduces the duplication of logic.  I will follow up by converting the 4x queries to a single query.

### DIFF
--- a/app/models/publication.rb
+++ b/app/models/publication.rb
@@ -2,6 +2,8 @@ class Publication < ActiveRecord::Base
   has_paper_trail on: [:destroy]
   scope :with_active_author, -> { joins(:authors).where('authors.active_in_cap' => true).uniq }
 
+  attr_accessor :pubhash_needs_update
+
   after_create do
     set_sul_pub_id_in_hash
     save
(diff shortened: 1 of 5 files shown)


lib/web_of_science/process_records.rb
```diff
@@ -61,22 +60,11 @@ def select_new_wos_records
       def save_wos_records
         return if records.empty?
         process_links
-        records.select! do |rec|
```
atz: This `select!` never did any selecting, by definition, since `.save!` was never going to return false, it would just explode!

app/models/publication.rb
```diff
@@ -164,7 +166,7 @@ def pubhash_needs_update!(*_args)
   end
 
   def pubhash_needs_update?
-    @pubhash_needs_update || false
+    pubhash_needs_update || false
```
peetucket: aren't there other uses of @pubhash_needs_update in the publication model?

atz: Really just the `pubhash_needs_update!` setter, which I opted to let assign the variable directly.

PR #714, opened 2018-02-01
Drop empty `xml` field from publications table

### LINKED ISSUE #694: Empty Publication#xml field
The DB `publications` table has an `xml` field.  It is universally empty on production:
```
Publication.count
=> 446418
Publication.where(xml:nil).count
=> 446418
```
It seems useless given the abstraction of the various source records, and the fact that they aren't all XML.  We should drop the field.

atz:
Fixes #694

### DIFF
--- a/db/migrate/20180201233754_remove_xml_from_publication.rb
+++ b/db/migrate/20180201233754_remove_xml_from_publication.rb
@@ -1,0 +1,5 @@
+class RemoveXmlFromPublication < ActiveRecord::Migration
+  def change
+    remove_column :publications, :xml, :string
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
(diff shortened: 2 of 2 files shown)


peetucket:
Presumably xml is nil in all rows?

atz:
@darrenleeweber What do you imagine a test on a dumped DB would possibly show?  Either the field can be dropped or it cannot.  Loading tens or hundreds of GBs of data doesn't change that.

There is no code referencing `.xml` according to:
```
git grep -P "\.xml\b" lib/ app/ 
lib/clarivate/links_client.rb:          file: Rails.root.join('lib', 'clarivate', 'links_request.xml'),
lib/csl/cap_mapper.rb:      # http://aurimasv.github.io/z2csl/typeMap.xml#map-report
```

@peetucket Yes, all rows `nil` as demonstrated in #694.

atz:
Also, there is no point testing on yet another separate DB dump.  Just deploy to any non-prod system first and let that be a valid demonstration.

PR #715, opened 2018-02-02
WOS process records src handler

darrenleeweber:
Work from #712 prompted reconsideration of the fault-tolerance for processing records when the src record fails to save.  The changes here handle failures to save, log a message and move on.  Any src record that fails to save will be skipped in the process.

### DIFF
--- a/lib/web_of_science/process_record.rb
+++ b/lib/web_of_science/process_record.rb
@@ -48,7 +48,7 @@ module WebOfScience
       # whether it should create a new Publication.pub_hash, PublicationIdentifier(s) and Contribution(s).
       # @return [String, nil] WosUID that create a new Publication
       def process_record
-        save_record # as WebOfScienceSourceRecord
+        return unless save_record # as WebOfScienceSourceRecord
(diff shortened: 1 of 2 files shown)


atz:
How are you expecting this to fail?  I mean, I understand the *theory*, I just don't see any way for this to get tripped up given there are no significant validations on the model.  

You might wait to try to add this in until I'm done consolidating queries in processing.  I expect to replace the entire series of record writes with a single `create!` statement for all of them.

darrenleeweber:
- just want to be sure that it's not possible to save a publication/contribution without first saving the src data, because the src might never get saved if that happens (i.e. the UIDs are culled if they match a contribution)
- using `create!` with the pub-hash-update attr would be better
  - it could either get modified here in this PR or wait on another PR

darrenleeweber:
I've pushed up changes to be consistent with changes in #716 (to avoid conflicts when one or other is merged).

darrenleeweber:
Superceded by #721 and #727

Issue #671, opened 2018-01-24
WOS Harvester profiling


darrenleeweber:
Profile the WOS Harvester for authors, using authors who have a range of N publications from < 100 (one batch) up to 1000's of publications (many batches).

- try using https://github.com/ruby-prof/ruby-prof
- see related work on Argo in https://github.com/sul-dlss/argo/issues/250

### Results of profiling on the `wos_retriever` branch  (on Jan 24)
- using a cap-qa db-dump from Dec 31, 2017 (it has no WOS record content)
- [wos_harvest_profile.log](https://github.com/sul-dlss/sul_pub/files/1668826/wos_harvest_profile.log)
- [wos_harvest_profile.txt](https://github.com/sul-dlss/sul_pub/files/1668827/wos_harvest_profile.txt)

darrenleeweber:
The profiling script is no-frills in this first pass on a clean cap-qa-db-dump
```
# RAILS_ENV=production bundle exec rails runner ./script/wos_harvest_test.rb > log/wos_harvest_profile.txt
require 'ruby-prof'
RubyProf.start
authors = Author.where(active_in_cap: true, cap_import_enabled: true).limit(15)
WebOfScience.harvester.harvest(authors)
result = RubyProf.stop
# print a flat profile to text
printer = RubyProf::FlatPrinter.new(result)
printer.print(STDOUT)
```

darrenleeweber:
[wos_harvest_profile.txt](https://github.com/sul-dlss/sul_pub/files/1662474/wos_harvest_profile.txt)

darrenleeweber:
The profile data has a ton of stuff in it and all the WebOfScience data is way down the list in terms of performance costs.  To find some of it...

`grep 'WebOfScience' log/wos_harvest_profile.txt` using file ^^

```
Measure Mode: wall_time
Thread ID: 47107530575720
Fiber ID: 47107555250160
Total: 4223.323206
Sort by: self_time

 %self      total      self      wait     child     calls  name
  0.01      5.592     0.317     0.000     5.275    95228   WebOfScience::Record#identifiers
  0.00     11.985     0.200     0.000    11.785     4151   WebOfScienceSourceRecord#init_from_source
  0.00      0.191     0.159     0.000     0.032    40392   WebOfScience::Identifiers#database
  0.00      5.233     0.158     0.000     5.075     8328   WebOfScience::Identifiers#initialize
[...]

PR #716, opened 2018-02-03
Query efficiently

atz:
Most of the work here is in `WebOfScience::Contributions`.
- use `find_or_create_by!` with a block to collapse two methods.  
- several guard clause checks move further up the chain
- Avoid querying `publication_identifiers` up to 4x serially by using an single query covering all 4 types.  The volume of these queries were indicated as one bottleneck in #671.
- Return fewer Booleans.  Instead: for the `true` case, return an object the caller might use (to avoid another query), for `false`, return `nil`.  

The resulting SQL (from execution) looks like:
```
SELECT  `publications`.* FROM `publications` INNER JOIN `publication_identifiers` ON `publication_identifiers`.`publication_id` = `publications`.`id` 
WHERE (publication_identifiers.identifier_value IS NOT NULL AND (
     (publication_identifiers.identifier_type = 'WosUID' AND publication_identifiers.identifier_value = 1) OR
     (publication_identifiers.identifier_type = 'WosItemID' AND publication_identifiers.identifier_value = 2) OR
     (publication_identifiers.identifier_type = 'doi' AND publication_identifiers.identifier_value = 3) OR
     (publication_identifiers.identifier_type = 'pmid' AND publication_identifiers.identifier_value = 4)))  
ORDER BY CASE
          WHEN publication_identifiers.identifier_type = 'WosUID' THEN 0
          WHEN publication_identifiers.identifier_type = 'WosItemID' THEN 1
          WHEN publication_identifiers.identifier_type = 'doi' THEN 2
          WHEN publication_identifiers.identifier_type = 'pmid' THEN 3
END LIMIT 1
```

### DIFF
--- a/lib/web_of_science/contributions.rb
+++ b/lib/web_of_science/contributions.rb
@@ -34,28 +34,18 @@ module WebOfScience
     # @param [Publication]
     # @return [Contribution]
     def find_or_create_contribution(author, publication)
-      contrib = publication.contributions.find_or_initialize_by(author_id: author.id)
-      create_contribution(author, contrib) unless contrib.persisted?
-      contrib
-    end
-
-    # Save a new contribution to a WOS-UID publication for author.
-    # @param author [Author]
-    # @param contrib [Contribution]
-    # @return [Boolean]
-    def create_contribution(author, contrib)
-      contrib.assign_attributes(
-        cap_profile_id: author.cap_profile_id,
-        featured: false, status: 'new', visibility: 'private'
-      )
-      contrib.save!
-      # Add the pub_hash[:authorship] data to the publication
-      contrib.publication.pubhash_needs_update!
-      contrib.publication.save!
+      publication.contributions.find_or_create_by!(author_id: author.id) do |contrib|
+        contrib.assign_attributes(
+          cap_profile_id: author.cap_profile_id,
+          featured: false, status: 'new', visibility: 'private'
+        )
+        # Add the pub_hash[:authorship] data to the publication
+        publication.pubhash_needs_update!
+        publication.save! # contrib.save! not needed
+      end
     rescue ActiveRecord::ActiveRecordError => err
-      message = "Failed to create contribution for author: #{author.id}, pub: #{contrib.publication.id}"
+      message = "Failed to find/create contribution for author: #{author.id}, pub: #{publication.id}"
       NotificationManager.error(err, message, self)
-      false
     end
 
     # Does record have a contribution for this author? (based on matching PublicationIdentifiers)
@@ -77,10 +67,10 @@ module WebOfScience
     # @return [Boolean] contribution exists
     def contribution_by_identifier?(author, type, value)
       return false if type.blank? || value.blank?
-      pub_id = PublicationIdentifier.find_by(identifier_type: type, identifier_value: value)
-      return false if pub_id.nil?
-      contrib = find_or_create_contribution(author, pub_id.publication)
-      contrib.nil? ? false : contrib.persisted?
+      pub = Publication.includes(:publication_identifiers)
+                       .find_by("publication_identifiers.identifier_type": type, "publication_identifiers.identifier_value": value)
+      return false unless pub
+      find_or_create_contribution(author, pub) ? true : false
     end
   end
 end
--- a/lib/web_of_science/contributions.rb
+++ b/lib/web_of_science/contributions.rb
@@ -76,30 +76,11 @@ module WebOfScience
     # @param value [String]
     # @return [Boolean] contribution exists
     def contribution_by_identifier?(author, type, value)
-      contrib = contribution_by_identifier(author, type, value)
+      return false if type.blank? || value.blank?
+      pub_id = PublicationIdentifier.find_by(identifier_type: type, identifier_value: value)
+      return false if pub_id.nil?
+      contrib = find_or_create_contribution(author, pub_id.publication)
       contrib.nil? ? false : contrib.persisted?
     end
-
-    # Find any matching contribution by author and PublicationIdentifier
-    # @param author [Author]
-    # @param type [String]
-    # @param value [String]
-    # @return [Contribution, nil]
-    def contribution_by_identifier(author, type, value)
-      pub_id = publication_identifier(type, value)
-      return if pub_id.nil?
-      find_or_create_contribution(author, pub_id.publication)
-    end
-
-    # Is there a PublicationIdentifier matching the type and value?
-    # @param type [String]
-    # @param value [String]
-    # @return [PublicationIdentifier, nil]
-    def publication_identifier(type, value)
-      return if type.blank? || value.blank?
-      pub_ids = PublicationIdentifier.where(identifier_type: type, identifier_value: value)
-      pub_ids.empty? ? nil : pub_ids.first
-    end
-
   end
 end
--- a/lib/web_of_science/contributions.rb
+++ b/lib/web_of_science/contributions.rb
@@ -39,8 +39,7 @@ module WebOfScience
           cap_profile_id: author.cap_profile_id,
           featured: false, status: 'new', visibility: 'private'
         )
-        # Add the pub_hash[:authorship] data to the publication
-        publication.pubhash_needs_update!
+        publication.pubhash_needs_update! # Add to pub_hash[:authorship]
         publication.save! # contrib.save! not needed
       end
     rescue ActiveRecord::ActiveRecordError => err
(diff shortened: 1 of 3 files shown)


lib/web_of_science/contributions.rb
```diff
@@ -34,40 +34,43 @@ def author_contribution(author, uid)
...
-    # @return [Boolean]
-    def found_contribution?(author, record)
-      contribution_by_identifier?(author, 'WosUID', record.uid) ||
-        contribution_by_identifier?(author, 'WosItemID', record.wos_item_id) ||
-        contribution_by_identifier?(author, 'doi', record.doi) ||
-        contribution_by_identifier?(author, 'pmid', record.pmid)
+    # @return [::Contribution, nil] a matched or newly minted Contribution
+    def matching_contribution(author, record)
+      pub = Publication.joins(:publication_identifiers).where(
+        "publication_identifiers.identifier_value IS NOT NULL AND (
```
atz: This clause would not be necessary if we had the desired validations/constraints on `publication_identifiers`.

lib/web_of_science/process_records.rb
```diff
@@ -71,13 +71,12 @@ def save_wos_records
...
+        pub = Publication.create!(
           active: true,
           pub_hash: record.pub_hash,
-          wos_uid: record.uid
+          wos_uid: record.uid,
+          pubhash_needs_update: true
         )
-        pub.pubhash_needs_update!
-        pub.save!
         contrib = find_or_create_contribution(author, pub)
```
darrenleeweber: :+1:

lib/web_of_science/process_records.rb
```diff
@@ -98,33 +98,18 @@ def create_publication(record)
       # @return [void]
       def process_links
         links = retrieve_links
-        records.each { |rec| update_links(rec, links[rec.uid]) }
+        records.each { |rec| rec.identifiers.update(links[rec.uid]) if rec.database == 'WOS' }
       rescue StandardError => err
-        message = "Author: #{author.id}, process_links failed"
-        NotificationManager.error(err, message, self)
+        NotificationManager.error(err, "Author: #{author.id}, process_links failed", self)
       end
```
darrenleeweber: Not sure I can approve this change because if any exception arises when updating one record, it will kill the iteration for all the rest of the records.  That's why it's split out into two methods, so the method that handles the links update for one record can handle and log any exception for it without killing the iteration across records.

lib/web_of_science/contributions.rb
```diff
@@ -76,30 +76,11 @@ def found_contribution?(author, record)
...
-    # @param type [String]
-    # @param value [String]
-    # @return [PublicationIdentifier, nil]
-    def publication_identifier(type, value)
-      return if type.blank? || value.blank?
-      pub_ids = PublicationIdentifier.where(identifier_type: type, identifier_value: value)
-      pub_ids.empty? ? nil : pub_ids.first
-    end
-
   end
```
darrenleeweber: - At some point, the original version of this method was something like this.  To fix the code-climate complexity, it was broken out into component methods.  In doing so, it focuses each method on a specific purpose.
- I need to double check whether any of these methods are used elsewhere.
- Not entirely sure this buys significant performance improvement (apart from a bit of method call overhead)

lib/web_of_science/contributions.rb
```diff
@@ -34,28 +34,18 @@ def author_contribution(author, uid)
...
+        # Add the pub_hash[:authorship] data to the publication
+        publication.pubhash_needs_update!
+        publication.save! # contrib.save! not needed
+      end
     rescue ActiveRecord::ActiveRecordError => err
-      message = "Failed to create contribution for author: #{author.id}, pub: #{contrib.publication.id}"
+      message = "Failed to find/create contribution for author: #{author.id}, pub: #{publication.id}"
       NotificationManager.error(err, message, self)
-      false
     end
```
darrenleeweber: - guess this is OK, although not entirely clear about what happens with model callbacks that trigger changes to pub_hash data when using `create!` vs. `save!` - need to be sure that the callback behavior is the same.
- need to double check that `create_contribution` is not used anywhere

lib/web_of_science/contributions.rb
darrenleeweber: - Maybe the `Publication.includes(:publication_identifiers).find_by` is a performance gain?
- Perhaps performance improvements would be better gained by looking at indexes on the PublicationIdentifier values (and cleaning up the data in that table, along with validations).

darrenleeweber: (BTW, going through this commit-by-commit, not sure how github will display the resulting comments.  Seems like some commits on the Contributions might be squashed, if all this is consolidated and methods removed are not used elsewhere, and code-climate complexity is ignored.)

lib/web_of_science/contributions.rb
```diff
@@ -50,14 +49,28 @@ def find_or_create_contribution(author, publication)
...
+        "CASE
+          WHEN publication_identifiers.identifier_type = 'WosUID' THEN 0
+          WHEN publication_identifiers.identifier_type = 'WosItemID' THEN 1
+          WHEN publication_identifiers.identifier_type = 'doi' THEN 2
+          WHEN publication_identifiers.identifier_type = 'pmid' THEN 3
+         END"
+      ).first
+      return unless pub
+      find_or_create_contribution(author, pub)
     end
```
darrenleeweber: Does this version of the query work when any of `record.wos_item_id, record.doi, record.pmid` are `nil`?  AFAIK, the `record.uid` should be guaranteed by the WOS-API.

lib/web_of_science/contributions.rb
```diff
@@ -53,7 +53,7 @@ def find_or_create_contribution(author, publication)
     # @param [Author] author
```
darrenleeweber: Commit 3cfd37e07cb9efa8a168f6346c68fdf7a45b93a6 needs to be squashed with the commit that changed the return values.

lib/web_of_science/contributions.rb
atz: Re: `Publication.includes(:publication_identifiers).find_by`, yes.  We ask for the thing we want rather than searching for the pub ID and then asking for `pub_id.publication` (which causes a separate DB query).

lib/web_of_science/contributions.rb
atz: Any `nil` value won't matter because they will be excluded by the first clause: `publication_identifiers.identifier_value IS NOT NULL`.  Can't match a `NOT NULL` condition and a `= NULL` condition at the same time.

`.first` does `LIMIT 1`, as you can verify in your local rails dev console.  

I'm assuming we cared based on the priority of the previous implementation.  I reproduced that priority for the unified query.

lib/web_of_science/process_records.rb
atz: Right now there is basically no validation on pub IDs, so the circumstances under which `update` might fail would be extraordinary. Previously, the same code could *still* fail for all records in the more likely spot of the `retrieve_links` remote call, so I don't think the cost of separate method and `rescue` context actually delivers the desired bulletproofing.

peetucket: I can't tell if this is intended to be a performance improvement to reduce the number of queries or a simplification/refactor?

lib/web_of_science/contributions.rb
darrenleeweber: The priority of the previous query was on the match-priority, which resulted in a return priority.  AFAICT, the new query doesn't impose a match-priority up front, but sorts the results by that priority.  It gets all the results in one hit, to get one result.  So the change in the query does hit the db once and that seems like a performance improvement in terms of IO perhaps - but the db search algorithm itself might have potentially more work to do on the query, depending on the case?  If it matched on say WOS-UID, the previous code was faster.  If it matched on the last identifier, the new code could be faster.  The table is indexed by type & value, so that helps.  I'm not sure what the details of the execution times could be - it would be interesting to benchmark it.  On this branch, did the profiling results improve for this change?

peetucket: I would prefer the combined query approach since the columns we are querying on are indexed - mysql should be well optimized to do this plus we avoid more ActiveRecord overhead.

lib/web_of_science/process_records.rb
atz: Mainly the latter.  I don't think it makes sense to expect local DB failures here (compared to expecting potential network failures for remote requests).  Performance improvement would be from avoiding unnecessary methods and rescue contexts (relatively small compared to reducing queries).

atz:
120 line delta is pretty manageable and we already agreed that consolidating queries would improve performance.

lib/web_of_science/contributions.rb
atz: The query may pull more rows prior to the SORT and LIMIT.  But we if we aren't worried about the performance of the DB pulling a single row based on two indexed fields, we shouldn't suddenly become worried when that is increased to MAX 4 rows.

PR #717, opened 2018-02-03
Add a validator for identifier_type presence

atz:
Obviously, the pub ID record is useless if it doesn't identify what kind of ID it is.

We would want to add validation for `identifier_value` as well, but, ironically, that is blocked because it breaks dozens of lines of specs for code designed to clean up Pub IDs.  So the code meant to clean up values is what prevents us from ensuring values are even there.  :\

TODO: We could also choose to enumerate valid values for type.

### DIFF
--- a/app/models/publication_identifier.rb
+++ b/app/models/publication_identifier.rb
@@ -1,8 +1,8 @@
 class PublicationIdentifier < ActiveRecord::Base
   belongs_to :publication, required: true, inverse_of: :publication_identifiers
+  validates :identifier_type, presence: true
 
-  # A pub_hash[:identifier] entry
-  # @return [Hash]
+  # @return [Hash<Symbol => String>] an entry for pub_hash[:identifier]
   def identifier
     ident_hash = {}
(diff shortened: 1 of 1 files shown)


atz:
@darrenleeweber: there is no migration here.

darrenleeweber:
Technically it's not a migration, right.  I suppose the point I was trying to make is checking whether existing data will validate.  There is a script we are yet to run to clean up the identifiers; it worked on a prod-db without this validation, want to be sure that it works with this validation or to be sure this PR is only merged after the data is cleaned up.

atz:
Every prod pub ID does indeed have a type (because I deleted the ones that didn't months ago).

darrenleeweber:
But prod has been running for months, so changes months ago may not be true today or tomorrow.  Whatever bug allowed them to get in there, may have allowed more to get in there since those changes were made to delete them months ago.

atz:
`PublicationIdentifier.where(identifier_type:nil).count` is `0` on both prod and cap-qa.

PR #720, opened 2018-02-05
delayed_jobs should actually run

peetucket:
1. daemons gem must be declared in gemfile for delayed_jobs process to start
2. start delayed_job on cap deploys
3. restart delayed_job on reboot


peetucket:
just added a commit to add a delayed_job check to okcomputer

this was all part of the patch-party work when I realized we had delayed_job but it wasn't running and wasn't automatically restarted

no jobs currently in the -prod queue anyway, so that API endpoint is probably lightly or unused at the moment

PR #721, opened 2018-02-06
[WIP] WOS harvests could use background jobs

darrenleeweber:
Connected to #673
- once a record is retrieved from the WOS-API, process it in the background and move on
- this should allow for decoupling/concurrent retrieval and db-integration processes
  - i.e. the processing of retrieved records will not block retrieving more records
  - each WOS-record processing is orthogonal to all others
    - if they share any identifiers, the first one saved wins, that's OK and would happen anyway
  - the WOS record retrievals should be serial, because they all use single instances for:
    - `WebOfScience.client`, `WebOfScience.queries`, `WebOfScience.harvester`
    - there should be only one worker assigned to the `:wos_author_harvest` queue

### Update
- rebased on `wos-jobs-revised`, #727


atz:
Perhaps we want to persist the Links AMR responses, but it doesn't seem important and certainly shouldn't be coupled w/ introduction of asynchronous architecture.

PR #722, opened 2018-02-07
Explicit order where required

atz:
The error seems legitimately related to ruby hash order (default: order of key insertion), but the solution is to specify order for Savon/Gyoku's XML builder to use.


darrenleeweber:
For some reason, when I use this branch and remove the VCR fixutes, my local copy cannot pass specs.  For some reason, my system cannot generate the same XML request, using the same code and same bundle install.  (Maybe there are libxml or other gem or XML library differences somewhere, possibly related to native versions?)

PR #723, opened 2018-02-07
WebOfScience WSDL stubs

darrenleeweber:
Try to lighten the load of VCR recordings for WOS specs.  The intention of this PR is to provide two generic methods that stub the retrieval of WSDL data for the authentication and search endpoints of the WOS-API.  It's beyond the scope of this PR to apply them everywhere, which can be done in follow on PRs.

- if it were possible, it would be better to initialize the `WebOfScience::Client` with these files every time so that the auth and search Savon clients could be initialized with them.  AFAICT, this is not an option for the Savon client, it only accepts a WSDL URL.  If I'm wrong about this (and I hope so), then these fixture files could be used as such and the nature of this PR would change substantially - the focus could shift from using these fixtures in specs only to using them all the time.

### Testing
- wos-client tests with `:vcr` enabled shows recordings of the WSDLs, the content of the following files includes the WSDL when those requests are _not_ stubbed.
```
$ tree fixtures/vcr_cassettes/WebOfScience_Client/
fixtures/vcr_cassettes/WebOfScience_Client/
├── _authenticate
│   └── authenticates_with_the_service.yml
├── _search
│   └── works.yml
├── _session_close
│   ├── resets_the_client.yml
│   └── when_there_are_no_matches_returned_for_SessionID
│       └── creates_a_logger_and_works.yml
└── _session_id
    └── works.yml
```
- when the WSDL requests are stubbed, with `:vcr` enabled, there are no recordings:
```
$ ls -l fixtures/vcr_cassettes/WebOfScience_Client
ls: cannot access 'fixtures/vcr_cassettes/WebOfScience_Client': No such file or directory
```

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -168,18 +168,5 @@ def default_institution
   )
 end
 
-
-WOS_AUTH_WSDL = File.read('spec/fixtures/wos_client/authenticate_wsdl.xml')
-def stub_wos_auth_wsdl
-  raise unless WebOfScience::Client::API_VERSION == '3.0'
-  stub_request(:get, WebOfScience::Client::AUTH_WSDL).
-    to_return(status: 200, body: WOS_AUTH_WSDL)
-end
-
-WOS_SEARCH_WSDL = File.read('spec/fixtures/wos_client/search_wsdl.xml')
-def stub_wos_search_wsdl
-  raise unless WebOfScience::Client::API_VERSION == '3.0'
-  stub_request(:get, WebOfScience::Client::SEARCH_WSDL).
-    to_return(status: 200, body: WOS_SEARCH_WSDL)
-end
-
+require_relative 'lib/web_of_science/wsdl'
+WebOfScience::WSDL.fetch
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -167,3 +167,19 @@ def default_institution
     Agent::AuthorAddress.new(Settings.HARVESTER.INSTITUTION.address.to_hash)
   )
 end
+
+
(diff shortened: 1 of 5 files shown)


spec/spec_helper.rb
```diff
@@ -167,3 +167,19 @@ def default_institution
     Agent::AuthorAddress.new(Settings.HARVESTER.INSTITUTION.address.to_hash)
   )
 end
+
+
+WOS_AUTH_WSDL = File.read('spec/fixtures/wos_client/authenticate_wsdl.xml')
+def stub_wos_auth_wsdl
+  raise unless WebOfScience::Client::API_VERSION == '3.0'
```
atz: Why this `raise`?  Why would it break at say, `3.1`?

darrenleeweber: I need some way to check or flag that these fixture files need to be updated; it's important that they accurately record what it currently at the URL.  This is the best idea that I came up with so far.

atz: Not sure I follow.  If the WSDL changes on the back end, nothing guarantees that `WebOfScience::Client::API_VERSION` changes to match.  I guess if the theory is we use that variable to signal changes, it might be OK, but it also precludes any kind of "minor" SemVer compatibility.

darrenleeweber: Agreed, it's far from perfect.  The only real check is to retrieve it again and do a diff on the fixture file.  If necessary, this PR could add a method to do that once (and only once) when the specs are run.  If we go that way, that change will remove the check on the version number.

darrenleeweber:
The latest commit pushed fetches the WSDL files and writes them to fixtures, once, so that the GET stubs will be using the latest files.  If those fixture files ever change, it will be a useful flag that something in the API could have changed (or at least the file format has changed or something).  Changes in the WSDL fixture files should be infrequent.

PR #725, opened 2018-02-08
add a new name variant of Last Name,FirstInitialMiddleInitial when searching

### LINKED ISSUE #728: Add a name variant of Last, FirstMiddle when searching by name

peetucket:
Fixes #728

Yes, you probably need to add the 1 variant “Gentles, AJ”.  That would probably work.  The name search is matching straight text.  The wildcard “\*” could be used (“Gentles,A*”), but then you pick up all records with Gentles, A…, which probably would produce numerous false hits.
 
Thanks.
Rob

---------------------------- 
We noticed that for a particular faculty member use case, there were a few publications that were in the WoS database but weren't being returned by our name query (query shown below along with the missed publications).  Just curious if you had any ideas of if we need to do something better with our query structure.  For example, I'm wondering if we need to add another name variant like "Gentles, AJ".

Faculty member:
Gentles, Andrew J.
 
Query sent:
AU=("Gentles,Andrew" or "Gentles,A" or "Gentles,Andrew" or "Gentles,A,J") AND AD=(stanford)
 
Missed Publications:
Data normalization considerations for digital tumor dissection
10.1186/s13059-017-1257-4
WOS:000404744800002
 
Low BUB1 expression is an adverse prognostic marker in gastric adenocarcinoma
10.18632/oncotarget.19357
WOS:000412066700048


peetucket:
Found this test failing even before I made my change:

rspec ./spec/lib/web_of_science/record_spec.rb:92

Also, not sure why the author query is constructed in two places...

atz:
What's the wildcard here?  Didn't seem to make it across your copy/paste.

peetucket:
His wildcard suggested an asterix character (fixed my copy and paste) but I don't think that is wise as it may have too many false positives.  But adding "Last, FirstMiddle" (note first and middle initial are appended to each other) makes sense.

atz:
I would expect this to invalidate a bunch of VCR cassettes, so it's sorta odd that it doesn't.

peetucket:
Good point. I guess it depends on what we cache?  If it didn’t turn up any more results for the queries in the cassettes would they be the same?  It seems like we should be generating a different query request as a result regardless.

atz:
Yeah, specifically, I expected the cassettes to be altered because our outgoing query would have changed.  VCR is not fully semantic in this regard though: it thinks the [METHOD and URI](https://relishapp.com/vcr/vcr/v/3-0-3/docs/request-matching) are sufficient to match a request to a cassette.  That is useless for SOAP XML, since literally every request POST would be equivalent to the next.  The only differences between requests are in the body.  

We should pursue using `match_requests_on` to include `:body`.  Without it, our strategy of using VCR in WoS tests is questionable.

peetucket:
--- does this change also apply to alt-names?

yes, it should apply to any name search against WoS ... I'm trying to understand the difference between the author_name class and the query class since we seem to be constructing a name query in both.  Do you know if one is unused now?

PR #727, opened 2018-02-09
Wos jobs revised and store Links AMR data in WoS Source Record

darrenleeweber:
Extracted from #721 

- this refactors existing jobs that support the authors-controller API
  - it decouples the SW and WOS harvest jobs for _one_ author, checking settings first

### TODO
- other places that need non-blocking _single_ author harvests could use these jobs
  - e.g. some wos-rake tasks that need to harvest for _one_ author, so they are non-blocking
  - e.g. a wos-rake task could facilitate admin tasks to process a single author:record that could be missing from a harvest or used to recover from or debug failures to integrate a wos-src-record into a publication/contribution

### Changes
- refactored jobs for SW and WOS, to decouple them and check settings up front
```
app/jobs/
├── author_harvest_base_job.rb
├── science_wire
│   └── author_harvest_job.rb
└── web_of_science
    ├── author_harvest_job.rb
    └── author_record_job.rb
```
- WOS-src-record gets `identifiers` field
  - it's a serialized Hash, mediumtext field
    - this captures all the Links-API data, which was not saved before
  - the source_fingerprint respects an AR update
- WOS-record gets utils for WOS-src-record
  - it can find or create and update a matching WOS-src-record based on UID
- WOS-process-record gets a WOS-src-record now
  - it can be used in an AR-active-job because it now accepts a WOS-src-record
  - given an author and a WOS-src-record:
    - it checks for contributions, adds publication, publication-identifiers and contribution
    - it now uses the ProcessPubmed module to handle those details
  - note that WOS-process-records is not involved
- WRT the harvester:
  - it uses WOS-process-records for most harvest paths, no changes there
  - it used WOS-process-record in the DOI, so the only change in this PR is to the DOI retrieval, i.e.
```diff
-      WebOfScience::ProcessRecord.new(author, record).execute
+      WebOfScience::ProcessRecord.new(author, record.source_record_find_or_create).execute
```


atz:
Here's my reservations.

### Introducing asynchronicity 
- context: in an app that literally has not seen async used (yes, we have a Job class defined and wired in, but we weren't actually running a job worker and logs do not show it having been used -- it would have been broken if it was used).
- timing: in the last hours of development after many months.
- complexity: guaranteed cost.
- purpose: no demonstrated need and it seems likely to make performance worse.   
- **fake scalability**: async is great when you can split up a task and have many workers run in parallel.  But because of current WOS throttling limits, we can never do that.  Running multiple workers will break the app.  To me, it is inadvisable to apply a pattern introducing complexity and architectural requirements while its primary benefit is precluded and actually harmful!

### Data
- more serialized Hashes in the database, as though we hadn't learned from the pains of working with `pub_hash`.  Obviously, persisting the links data is agreeable and fine, but this seems like another ill-advised shortcut.

### Code
- *still* propagating the `NotificationManager` anti-pattern.
- IMO, either this code should be integrated with the app, or it should be rejected.  It should not be accepted with the decision to switch over to it deferred, particularly at this late phase of development.  I don't want major architecture hanging out in `master` doing nothing.

app/jobs/web_of_science/author_harvest_job.rb
```diff
@@ -0,0 +1,19 @@
+module WebOfScience
+  class AuthorHarvestJob < ActiveJob::Base
+    queue_as :wos_author_harvest
+
+    # Performs an asynchronous harvest and save for a given Author
+    # @param [Author] author
+    # @param [Hash] options
+    # @return [void]
+    def perform(author, options = {})
+      return unless Settings.WOS.enabled
```
atz: I think the protection should be against even enqueuing a job if the setting is disabled.  No point filling up a queue just to figure out 10,000 times asynchronously that nothing should happen.  (This could block other jobs that actually might do something.  Just naming a different `queue_as` isn't sufficient to know how the backend is processing.)

darrenleeweber: Done
- required changes in the AuthorsController
- safe guards are still required in the jobs, esp if an API endpoint is not functional, to avoid retries

darrenleeweber:
Responses to @atz reservations:

### Introducing asynchronicity 
- context: in an app that literally has not seen async used (yes, we have a Job class defined and wired in, but we weren't actually running a job worker and logs do not show it having been used -- it would have been broken if it was used).
  - the API already has an active job - it must be enabled to support the API
  - any oversight to run/monitor the active job is a failure to be fixed
  - the failure was recently identified and a fix was merged
- timing: in the last hours of development after many months.
  - irrelevant, this PR has no change to the harvest process that is already merged
  - it's important to make any necessary WOS-src-record changes before doing the first harvest
  - this is a core piece of the entire wos-api work cycle, fixes are to be expected and better now than later
- complexity: guaranteed cost.
  - irrelevant, the entire wos-api work cycle is _all_ about more complexity
  - more harvesting systems are more complex, this work cycle adds one for WOS-API
- purpose: no demonstrated need and it seems likely to make performance worse.
  - the API already requires these jobs, we need to add and support the WOS-API
- **fake scalability**: async is great when you can split up a task and have many workers run in parallel.  But because of current WOS throttling limits, we can never do that.  Running multiple workers will break the app.  To me, it is inadvisable to apply a pattern introducing complexity and architectural requirements while its primary benefit is precluded and actually harmful!
  - this comment suggests a lack of understanding about how retrieval and processing are decoupled
  - retrieval (and saving a WOS-src-record retrieved) will be throttled by the WOS-API
  - subsequent processing of a WOS-src-record has no throttle limitations

### Data
- more serialized Hashes in the database, as though we hadn't learned from the pains of working with `pub_hash`.  Obviously, persisting the links data is agreeable and fine, but this seems like another ill-advised shortcut.
  - point is made, but there can be more identifiers than we want to define specific fields for
  - it's important to preserve _all_ the identifiers from the source record, rather than retrieve links again
  - the key identifiers that will be used to index records have specific fields already

### Code
- *still* propagating the `NotificationManager` anti-pattern.
  - this is _not_ the concern of this PR to change a pattern that could touch many parts of the code base
  - please document this concern in a different issue
- IMO, either this code should be integrated with the app, or it should be rejected.  It should not be accepted with the decision to switch over to it deferred, particularly at this late phase of development.  I don't want major architecture hanging out in `master` doing nothing.  
  - these jobs already exist in a part of the API
  - this PR refactors the jobs and supporting execution code

atz:
Thanks for adding checks to the controller.  

The API previously had an `ActiveJob` that was **never used**.  While the code dependency was there, obviously no *actual* load was put on that infrastructure.  In that sense, this PR *introduces* an operational dependency on asynchronous jobs where previously only the appearance of one existed.

One payoff of my recommending a base class was to utilize a common `rescue_from` declaration.  You have dismissed the requested changes, but not implemented that.  The current version is still susceptible to deserialization errors that will not be caught by the `rescue` block.  While deserialization from DB is not a common source of errors in synchronous code (`Hashie::Mash` issues aside), it is decidedly more common in async.  It's not hypothetical, we have experienced it on multiple projects.  This is part of what `rescue_from` helps us catch.  

I think you misunderstand how jobs execute.  Singleton pattern doesn't matter when you have completely different rails instances.  Singletons in different processes know nothing of each other and the number of workers is **not** constrained here. They would have to be in reality to avoid throttling problems.  Worker patterns achieve scalability by increasing the number of workers available.  But we cannot have WoS requests spawned from more than one process without maxing out the number of authorized sessions, meaning a concurrent real-time request would fail.  

### How many queues?
I see:
- `sw_author_harvest`
- `wos_author_harvest`
- `wos_author_record`
- `default` (the base class, presumably not actually queued from this code)

The current back-end is `DelayedJob`.  Can you specify what the `DelayedJob` configuration is that would improve performance without breaking from throttling?  It seems like any benefit is dependent on that and the biggest risks are enmeshed in getting it right.  

We only have **MAX 2** WOS auth'd sessions to put in play.  Assuming (at least!) one of those is for real-time requests, we cannot have multiple WOS queues process simultaneously.  We cannot even run these jobs, and realtime and scheduled harvests simultaneously.  What am I missing that makes adding processes workable?

darrenleeweber:
- `wos_author_harvest` will use a WOS-client
- `wos_author_record` does not use a WOS-client (it works on local-db content only)

atz:
@darrenleeweber: Barring some new information (like an increase in WoS number of concurrent sessions), please confirm our understanding of a **MAX 2** limit.  

Count of processes that would compete to fit under than limit:

1. realtime request
2. scheduled harvest
3. the **first** asynchronous WOS-harvesting worker
4. the **next** asynchronous WOS-harvesting worker
5. etc.

So how would this be safe?

darrenleeweber:
I'm out of time to pursue all the angles on this, sorry.  There are some useful changes in this PR that might be worth merging, with changes made in one or more followup PRs to deal with additional concerns.  If I had time, I would look into options for handling throttle limits and such on active jobs.  I recently noticed that the python celery protocol has parameters for such things, e.g.
- http://docs.celeryproject.org/en/latest/userguide/tasks.html#Task.rate_limit

Using active job with a sequential working seems to be possible, see
- https://stackoverflow.com/questions/34739958/delayed-job-how-to-process-job-sequentially-one-after-another

atz:
Yes, sequential (single queue) processing is possible (but not accomplished here).  But even the **first** async Job puts the count of possible sessions over the MAX 2 limit.  So it isn't a fix.

atz:
Closing with prejudice.

PR #733, opened 2018-02-12
WebOfScience::Retriever - refactor to behave as Enumerable

darrenleeweber:
- this is a time-boxed PR to explore an Enumerable pattern for the WOS-retriever
  - the `next_batch` method is still public
  - some routine enumerable methods are enabled
  - broken specs have been fixed

### DIFF
--- a/lib/web_of_science.rb
+++ b/lib/web_of_science.rb
@@ -1,4 +1,5 @@
 # rubocop:disable Style/ClassVars
+require 'set'
 
 # Web of Science (WOS) utilities
 module WebOfScience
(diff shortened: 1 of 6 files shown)


darrenleeweber:
I prefer to leave the `next_batch` and `next_batch?` as a public methods.  Given that this is time-boxed, I don't want to track down everything that would break if that public API were to change today.  Having said that, yes, in theory, it's possible they could be made private by a later PR.  I just don't want to tackle that here in this PR, which seems to be working afer a couple hours work on it.

atz:
Now that there is conflict, I would more likely just rebuild this than resolve/salvage it.

PR #742, opened 2018-02-20
update cap_profile_id checked since previous referenced a no longer active cap profile

peetucket:
darren hardy doesn't work here anymore and the previous ID would start failing when we go to -prod


atz:
CI seems to need a new VCR cassette or something?

PR #747, opened 2018-02-26
[WIP] remove null values in addresses; add in :name value into author section of pub-hash for wos records

peetucket:


peetucket:
speculative - unneeded for now

PR #752, opened 2018-02-27
we need to quote the title search provided by the user

### LINKED ISSUE #755: Quote and escape title search strings when construting queries

peetucket:
fixes #755
 to avoid issues with boolean operators in the title messing up the search...If you do a title search that includes a boolean operator, e.g. 
```GET https://sulcap-dev.stanford.edu/publications/sourcelookup?title=Nutrition and Mesenteric Issues in Pediatric Cardiac Critical Care.```

you will get a 500, which comes from the Clarivate API.  The reason is that if you don't add quotes around the title when constructing the query to clarivate it will interpret the "and" and "or" in the title incorrectly.


atz:
Makes sense.  We probably need to purge or escape any quotes inside the search term also, or it will break the DSL semantics on the back-end.

PR #756, opened 2018-02-28
deal with situations where the full_name is missing but the display_name is provided

### LINKED ISSUE #757: Deal with missing full_name when parsing medline names
There are some MEDLINE records containing "anononymous" names that currently fail unless we guard against this. For an example, do a title search for "Cancer Research" on sul-pub-cap-dev:

http://sul-pub-cap-dev.stanford.edu/publications/sourcelookup.json?title=Cancer Research

peetucket:
fixes #757 

There are some MEDLINE records containing "anononymous" names that currently fail unless we guard against this.  For an example, do a title search for "Cancer Research" on sul-pub-cap-dev:

`http://sul-pub-cap-dev.stanford.edu/publications/sourcelookup.json?title=Cancer Research`


spec/lib/web_of_science/map_names_spec.rb
```diff
@@ -58,6 +62,21 @@
     it_behaves_like 'contains_author_data'
   end
 
+  context 'MEDLINE records with anonymous names' do
+    let(:pub_hash_class) { described_class.new(medline_record_anon) }
+    let(:pub_hash) { pub_hash_class.pub_hash }
+
+    it 'works with WOS records with anonymous names' do
+      expect(pub_hash_class).to be_an described_class
```
atz: This test can never detect a "failure".  `AnyClass.new` must return an instance of the class or **Ruby** is broken.

spec/lib/web_of_science/map_names_spec.rb
```diff
@@ -58,6 +62,21 @@
...
+      expect(pub_hash[:author]).not_to be_nil
+      expect(pub_hash[:authorcount]).not_to be_nil
```
atz: If this is an integer, a `be > 0` matcher would be more accurate.

spec/lib/web_of_science/map_names_spec.rb
peetucket: expect 1 == 1

PR #759, opened 2018-02-28
add pmcid to pub_hash when available from pubmed

### LINKED ISSUE #758: PubMedCentralID is missing (PMCID) and is not being added to the identifier block for Pubmed supplemented records
For an example, see SulPubID = 347251
This has a PMID of 26394642

The Pubmed Source Record has correctly parsed the pmc identifier into the pub_med supplementary part of the pub_hash.  But that ID was not integrated into the actual publication pub_hash (likely because there was already an identifier block in the original pub_hash and the reverse_update didn't overwrite it).

https://github.com/sul-dlss/sul_pub/blob/master/lib/web_of_science/process_pubmed.rb#L40

https://github.com/sul-dlss/sul_pub/blob/master/lib/web_of_science/process_record.rb#L92

Note: it is unclear why we have two methods that seem to do the same thing: `pubmed_addition` in the ProcessPubmed module and `pubmed_addition` in the ProcessRecord class.  Perhaps one of them is never called.

peetucket:
fixes #758 

mimics what was done for SW before here: https://github.com/sul-dlss/sul_pub/blob/master/app/models/publication.rb#L277-L286

Notes:
- why are there two methods that do this?  do we know?  can we eliminate one?

### DIFF
--- a/lib/web_of_science/process_pubmed.rb
+++ b/lib/web_of_science/process_pubmed.rb
@@ -37,7 +37,10 @@ module WebOfScience
       if pubmed_record.nil?
         pubmed_cleanup(pub)
       else
-        pub.pub_hash.reverse_update(pubmed_record.source_as_hash)
+        pubmed_hash = pubmed_record.source_as_hash
+        pub.pub_hash.reverse_update(pubmed_hash)
+        pmc_id = pubmed_hash[:identifier].detect { |id| id[:type] == 'pmc' }
+        pub.pub_hash[:identifier] << pmc_id if pmc_id
         pub.pubhash_needs_update!
         pub.save!
       end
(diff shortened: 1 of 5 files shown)


atz:
@peetucket So we already seem to have **59122** `pmc` values in `publication_identifiers` table:
```ruby
PublicationIdentifier.group(:identifier_type).count
=> {"doi"=>197977, "isbn"=>2507, "legacy_cap_pub_id"=>7751, "pmc"=>59122, "PMID"=>263203, "PublicationItemID"=>380357, "SULPubId"=>8208, "WoSItemID"=>335964}
```
Shouldn't we be storing them there and relying on the regular rebuilding to put them in pub_hash?

atz:
After some review, we find that callbacks should be creating the PublicationIdentifier entries when the pub is saved.  Adding a test would confirm it.

peetucket:
Updates tests to confirm publication identifiers are being persisted from the pub_hash for the new pmc value

PR #760, opened 2018-03-02
Inproceeding detection

### LINKED ISSUE #754: Document type mapping is not consistent between SW and WoS
document type mappings should be the same from SW API to WoS API for the same publication

```https://sulcap-dev.stanford.edu/publications/sourcelookup?title=ESTIMATION OF THE AVERAGE WARFARIN MAINTENANCE DOSE IN SAUDI POPULATION```

produces for WOS:000267428100504

` "type": "article",`

```https://sulcap.stanford.edu/publications/sourcelookup?title=ESTIMATION OF THE AVERAGE WARFARIN MAINTENANCE DOSE IN SAUDI POPULATION```

produces for the same publication (happens to be SWID: 31814578 but same WOS ID):

`"type": "inproceedings",`

atz:
Fix #754

### DIFF
--- a/lib/web_of_science/map_pub_hash.rb
+++ b/lib/web_of_science/map_pub_hash.rb
@@ -1,30 +1,21 @@
 module WebOfScience
-
   # Map WOS record data into the SUL PubHash data
-  class MapPubHash < Mapper
+  class MapPubHash
+    attr_reader :pub
+    alias pub_hash pub
 
-    # Map WOS record data into the SUL PubHash data
-    # @return [Hash]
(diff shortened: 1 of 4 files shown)


atz:
@peetucket Because it didn't add accomplish anything logic-wise (other than a single param check).

- We didn't use `uid` 
- We didn't use `database` 
- We didn't use `logger` 
- We override `pub_hash` as expected
- We override `extract`, but called `super` to extract things we don't care about.

That's literally everything in the class: we use none of it.

PR #761, opened 2018-03-03
Add :name key to author block to mimic what is in SW records

### LINKED ISSUE #762: Return a :name value in WoS records to mimic old SW pub hash format

peetucket:
fixes #762 

* add name key to author block, which mimics what we are returning for SW records, this helps the Profiles team
* parse some WoS records that return first and middle names in the first name field to separate them correctly
* adds more specific tests for name parsing

### DIFF
--- a/lib/web_of_science/map_names.rb
+++ b/lib/web_of_science/map_names.rb
@@ -88,18 +88,28 @@ module WebOfScience
         name[:given_name] = given
         name[:first_name] = first if first.present?
         name[:middle_name] = middle if middle.present?
+        name[:name] = "#{name[:last_name]},#{name[:first_name]},#{name[:middle_name]}" # full name in the older style format used by Profiles/CAP
         name
       end
 
       # Parse the WOS names and return a Hash compatible with Csl::AuthorName
       # @return [Hash]
       def wos_name(name)
+        # look for the case where the first_name is the initials, with first and middle initials combined
+        # e.g. first_name = "RB", should be first_name = "R", middle_name = "B"
         match = name[:first_name].to_s.match(/\A([A-Z])([A-Z])/)
         if match
-          # first_name is the initials, with first and middle initials combined
           name[:first_name] = match[1]
           name[:middle_name] ||= match[2]
         end
+        # look for the case where the first name includes the middle initial optionally followed by a period
+        # e.g. first_name = "Russel B.", should be first_name = "Russel", middle_name = "B"
+        match = name[:first_name].to_s.match(/[\s][A-Z][.]?\z/)
+        if match
+          name[:first_name].delete!(match[0])
+          name[:middle_name] = match[0].strip[0]
+        end
+        name[:name] = "#{name[:last_name]},#{name[:first_name]},#{name[:middle_name]}" # full name in the older style format used by Profiles/CAP
         name
       end
   end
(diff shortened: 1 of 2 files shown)


lib/web_of_science/map_names.rb
```diff
@@ -88,18 +88,28 @@ def medline_name(name)
         name[:given_name] = given
         name[:first_name] = first if first.present?
         name[:middle_name] = middle if middle.present?
+        name[:name] = "#{name[:last_name]},#{name[:first_name]},#{name[:middle_name]}" # full name in the older style format used by Profiles/CAP
```
atz: Are you OK w/ the possibility that this becomes `'Anonymous,,'`?  That seems inelegant, at best.

peetucket: Yeah, its not great, but this is the format they are parsing on their end for display.  Since first and middle name are blank in this case, I'd rather leave it like that and let their parsing decide what to do (they must already handle cases of blank middle names anyway).

spec/lib/web_of_science/map_names_spec.rb
```diff
@@ -31,8 +31,9 @@
   end
 
   shared_examples 'contains_author_data' do
-    it 'has an authors' do
-      expect(pub_hash[:author]).not_to be_nil
+    it 'has authors with a name key in hash' do
+      expect(pub_hash[:author].size).to be > 0
+      pub_hash[:author].each { |author_name| expect(author_name[:name]).not_to be_nil }
```
atz: Looping like this is a bit of an anti-pattern.  Better rspec convention would be:
```ruby
expect(pub_hash[:author]).to all(match a_hash_including(name: be_present))
```

lib/web_of_science/map_names.rb
atz: OK, nevermind, I see you tested exactly that below.

spec/lib/web_of_science/map_names_spec.rb
peetucket: fixed and squashed commit

lib/web_of_science/map_names.rb
```diff
@@ -88,18 +88,28 @@ def medline_name(name)
...
+        # e.g. first_name = "RB", should be first_name = "R", middle_name = "B"
         match = name[:first_name].to_s.match(/\A([A-Z])([A-Z])/)
         if match
-          # first_name is the initials, with first and middle initials combined
           name[:first_name] = match[1]
           name[:middle_name] ||= match[2]
         end
+        # look for the case where the first name includes the middle initial optionally followed by a period
+        # e.g. first_name = "Russel B.", should be first_name = "Russel", middle_name = "B"
+        match = name[:first_name].to_s.match(/[\s][A-Z][.]?\z/)
```
atz: This regex/delete combo should be improved:
- regex has 3x as many braces as necessary (if indeed even one is required)
- the `[A-Z]` pattern fails to match Unicode characters   
- the `delete!` is dangerously inaccurate.  

You should use the regex to do the substitution (deletion) to make sure you don't end up with something like this:
```ruby
x = 'RUSSEL R. R.'
=> "RUSSEL R. R."
match = x.match(/[\s][A-Z][.]?\z/)
=> #<MatchData " R.">
x.delete!(match[0])
=> "USSEL"
x
=> "USSEL"
```

atz: To fix the above issues, I'd recommend something like:
```ruby
name[:first_name].gsub!(/\s[[:upper:]]\.?\z/) { |s| name[:middle_name] = s.strip; '' } # block returns empty, removing match
```

I confirmed that `[[:upper:]]]` matches unicode chars w/:
```ruby
'Ångström'.gsub(/[[:upper:]]/) {|s| puts s; '' }
Å
=> "ngström"
```

PR #772, opened 2018-03-07
Rake task to assign explicit WOSUIDs to an author

### LINKED ISSUE #479: Create WOS rake task to harvest by a list of publications by WosItemID for a cap_profile
The `sw:wos_profile_id_report` is a rake task that is used periodically (every one to two months).  It is a manual load of publications by `WosItemID` to a given `cap_profile_id`.  Implement this functionality for the `wos` rake tasks using the Web of Science API calls.

### Notes
- a previous call to the WOS API using a `WosItemID` without a database prefix returned multiple results from different databases (WOS and CCC, IIRC).  Without knowing, up front, what database a `WosItemID` belongs to, it is possible that it could return 0, 1 or many results.
- this was split out from #195

atz:
Exection is like:
```ruby
rake wos:assign_uids[98765] < input.txt
```

With output like:
```
Reading WosUIDs from STDIN...
10 UIDs to assign to Author 496 (cap_profile_id: 98765)
10 newly associated Contributions
```
Closes #479

### DIFF
--- a/lib/tasks/wos.rake
+++ b/lib/tasks/wos.rake
@@ -1,4 +1,17 @@
 namespace :wos do
+  # @example rake wos:assign_uids[98765] < input.txt
+  desc 'Assign WoS records (via UIDs on STDIN) to a cap_profile_id'
+  task :assign_uids, [:cap_profile_id] => :environment do |_t, args|
+    author = Author.find_by(cap_profile_id: args[:cap_profile_id])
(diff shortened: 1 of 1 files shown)


peetucket:
Presumably duplicates will not be created by the "process_uids" method?   Because it is helpful to just dump a list of UIDs in a file and run the task without having to worry if some of them are already in the profile.

atz:
Right, but right now in fact, the intended behavior will not be accomplished at all because of #774.

PR #776, opened 2018-03-07
create a connection between publication and web_of_science_source_record

peetucket:


atz:
Usually the symmetric relationship for a `has_one` is `belongs_to`, though in this case it is weird because the association does not involve a true primary key on either end.  

I think we would need add a DB-level `unique` constraint on the fields in question for `has_one <-> has_one` to work out.

atz:
The reason we can't just use the `uid <-> wos_uid` linkage is because timing.  The source record will be saved with the `uid` before the Pub that results from it, so the source record cannot have `uid` point to a Pub.

peetucket:
Yeah, realized that after the tests failed - not a normal activerecord association

also pondering if applying db constraints may lead to unforeseen consequences

atz:
Closing in favor of #779

PR #777, opened 2018-03-08
Report prod as the env rather than production

jcoyne:
For consistency with all the other projects. https://github.com/sul-dlss/DeveloperPlaybook/blob/master/best-practices/deployment.md#stage-names

### DIFF
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -31,3 +31,7 @@ set :linked_dirs, fetch(:linked_dirs, []).push(
 set :whenever_identifier, ->{ "#{fetch(:application)}_#{fetch(:stage)}" }
 
 set :delayed_job_workers, 2
+
+# honeybadger_env otherwise defaults to rails_env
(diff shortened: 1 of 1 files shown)


atz:
@jcoyne The playbook specifies the names of cap stage files, not honeybadger environments.

jcoyne:
This is a consistency issue for the department.  This is one of the changes we agreed to make in the June 17th, 2017 patch party, but it never happened for sul_pub.  You can see that it is indeed using `production` when you go here: https://app.honeybadger.io/projects/50046/deploys

Other projects that have this:
https://github.com/sul-dlss/exhibits/blob/master/config/deploy.rb#L45
https://github.com/sul-dlss/SearchWorks/blob/master/config/deploy.rb#L34
https://github.com/sul-dlss/argo/blob/master/config/deploy.rb#L80
https://github.com/sul-dlss/hydrus/blob/master/config/deploy.rb#L39
etc...

Issue #774, opened 2018-03-07
authorship_api 404s for known WOSUIDs


atz:
Repeated example from `production.log`:
```
Started POST "/authorship.json" for 10.35.49.120 at 2018-03-06 16:45:34 -0800
POST Contribution JSON: 
{"cap_profile_id":"35324","featured":false,"status":"approved","visibility":"private","wos_uid":"WOS:000364146300042"}
The WOS:000364146300042 publication was not found either locally or at WebOfScience.
Started POST "/authorship.json" for 10.35.49.120 at 2018-03-06 16:45:36 -0800
POST Contribution JSON: 
{"cap_profile_id":"35324","featured":false,"status":"approved","visibility":"private","wos_uid":"WOS:000364146300042"}
The WOS:000364146300042 publication was not found either locally or at WebOfScience.
Started POST "/authorship.json" for 10.35.49.120 at 2018-03-06 16:45:38 -0800
POST Contribution JSON: 
{"cap_profile_id":"35324","featured":false,"status":"approved","visibility":"private","wos_uid":"WOS:000364146300042"}
The WOS:000364146300042 publication was not found either locally or at WebOfScience.
Started POST "/authorship.json" for 10.35.49.120 at 2018-03-06 16:45:40 -0800
POST Contribution JSON: 
{"cap_profile_id":"35324","featured":false,"status":"approved","visibility":"private","wos_uid":"WOS:000364146300042"}
The WOS:000364146300042 publication was not found either locally or at WebOfScience.
Started POST "/authorship.json" for 10.35.49.120 at 2018-03-06 16:45:44 -0800
POST Contribution JSON: 
{"cap_profile_id":"35324","featured":false,"status":"approved","visibility":"private","wos_uid":"WOS:000364146300042"}
The WOS:000364146300042 publication was not found either locally or at WebOfScience.
```

Corresponding lines from `web_of_science.log` don't seem to suggest any problem:
```
I, [2018-03-06T16:45:34.980010 #10247]  INFO -- : WebOfScience::Harvester - author 17853 -  - 1 UIDs for search
I, [2018-03-06T16:45:34.982041 #10247]  INFO -- : WebOfScience::Harvester - author 17853 -  - 1 UIDs without contributions
I, [2018-03-06T16:45:36.699597 #10247]  INFO -- : WebOfScience::Harvester - author 17853 -  - 1 UIDs for search
[...]

PR #790, opened 2018-03-19
Tests for matching_publication and non-wos indirect matches

atz:
Because there are hundreds of lines of untested code and some hamstringing complexity even building fixtures, the level of confidence I have that these tests reflect our real scenario is lower than I'd like.  But they reflect it more closely than **no tests**.

My specific outstanding question is about the *timing* of pubmed additions.  The tests inject PMID ahead of time, but the actual `process_records` code only does `pubmed_additions` as the last step
(which has been true since its beginning). I suspect that is too late for the indirect matching query we need to work.

### DIFF
--- a/spec/lib/web_of_science/process_records_spec.rb
+++ b/spec/lib/web_of_science/process_records_spec.rb
@@ -8,6 +8,9 @@ describe WebOfScience::ProcessRecords, :vcr do
   let(:uids) { records.uids }
   let(:new_pubs) { Publication.where(wos_uid: uids) }
   let(:new_pub) { new_pubs.first }
+  let(:wos_records_links) do
+    { 'WOS:000288663100014' => { 'pmid' => '21253920', 'doi' => '10.1007/s12630-011-9462-1' } }
+  end
 
   before do
     allow(WebOfScience).to receive(:logger).and_return(Logger.new('/dev/null'))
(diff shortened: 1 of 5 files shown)


spec/lib/web_of_science/process_records_spec.rb
```diff
@@ -157,24 +156,75 @@
...
+        end
+        it 'adds new Contribution' do
+          expect { processor.execute }.to change { author.contributions.count }.from(0).to(1)
+          expect(Publication.find_by(wos_uid: records.first.uid)).not_to be_nil
+        end
+        it 'sets wos_uid on the Pub' do
+          expect { processor.execute }.to change { pub.reload.wos_uid }.from(nil).to(uid) # failing
+        end
+        it 'associates source record to existing Pub' do
+          expect { processor.execute }.to change { WebOfScienceSourceRecord.find_by(uid: uid).publication }.from(nil).to(pub) # failing
```
peetucket: tests no longer fail presumably so comments can be removed?

atz: right

peetucket:
fixes #404 ?

atz:
Yes, I think it does fix #774, but given that we've been going down rabbit holes a few times around different aspects of this problem, I'll be more sure of that once we test it on real data.

PR #794, opened 2018-03-24
More rails like way of detecting when author is changed if identity changes

peetucket:
It would be very useful to trigger author.changed? for any changes to author identities so we know when to re-harvest after a cap author update for that author...

This code will consider changes to the author_idenitity associated models as a change to the author model, thus eliminating the need to keep track of if alt identities changed in an instance variable.

This feels cleaner and more like how it should happen in a rails project.

Connected to #974


peetucket:
Indifferent, not sure it is really making the code less complex, so closing.

PR #802, opened 2018-03-30
do not throw an exception when a pmid cannot be found in a search

### LINKED ISSUE #801: Search by PMID produces an exception if no results are returned
See https://app.honeybadger.io/projects/50046/faults/37246705

It looks like the PubmedHarvester, as called from here: https://github.com/sul-dlss/sul_pub/blob/master/app/controllers/publications_controller.rb#L58

fails to check if pub exists before returning pub.pub_hash here:

https://github.com/sul-dlss/sul_pub/blob/master/app/controllers/publications_controller.rb#L58

It is quite possible that you can issue a search for a non-existent pmid, which should return nil or empty array instead of throwing an exception, which could be used by the controller so it can return an empty result set.

peetucket:
fixes #801

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -266,7 +266,7 @@ GEM
       activerecord (>= 4.2, < 5.2)
       request_store (~> 1.1)
     parallel (1.12.1)
-    parser (2.5.0.4)
+    parser (2.5.0.5)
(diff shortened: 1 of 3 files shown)


atz:
Added a commit trying to fix build failure due to dependency resolution

PR #820, opened 2018-04-14
[WIP] on manual searches, prefer pubmed over wos

peetucket:
Seems odd that WoS is the preferred source over Pubmed when searching manually by PMID.  This would swap the preferred order, though without specific stakeholder guidance, reluctant to push forward.


peetucket:
will recreate as a draft PR

PR #844, opened 2018-04-24
parse the year from another spot in the pubmed source record if missing in the first place

### LINKED ISSUE #843: Some Pubmed source record data has the year in a different location
Improve parsing of pubmed source records for year.

e.g. pmid = 29279863, sul_pub_id = 457360

This is a Pubmed Source Record, and the year is in a different location in the pubmed source record.  When users search for it by title, we return the record from WoS, which includes the year.  When they add it, they were getting the existing record from Pubmed, where the year wasn't parsed, therefore the citation was not correct.  I fixed this one manually, but it would be good to improve the parsing of Pubmed source records.

peetucket:
fix #843

### DIFF
--- a/app/models/pubmed_source_record.rb
+++ b/app/models/pubmed_source_record.rb
@@ -123,7 +123,19 @@ class PubmedSourceRecord < ActiveRecord::Base
     record_as_hash[:author] = author_list.map { |a| author_to_hash(a) }.compact
 
     record_as_hash[:mesh_headings] = mesh_headings unless mesh_headings.blank?
-    record_as_hash[:year] = publication.xpath('MedlineCitation/Article/Journal/JournalIssue/PubDate/Year').text unless publication.xpath('MedlineCitation/Article/Journal/JournalIssue/PubDate/Year').blank?
+    year_xpaths = [
+      'MedlineCitation/Article/Journal/JournalIssue/PubDate/Year',
+      'MedlineCitation/Article/ArticleDate/Year',
+      'PubmedData/History/PubMedPubDate[@PubStatus="accepted"]/Year'
+    ]
+    # look for a year in all of the xpath locations above in order
+    #  stop after the first produces something that looks like a year
+    year_xpaths.each do |path|
+      match = publication.xpath(path).text.match(/[12][0-9]{3}/)
+      next unless match
+      record_as_hash[:year] = match.to_s
+      break
+    end
 
     record_as_hash[:type] = Settings.sul_doc_types.article
 
--- a/spec/fixtures/pubmed/pubmed_record_29279863.xml
+++ b/spec/fixtures/pubmed/pubmed_record_29279863.xml
@@ -1,0 +1,351 @@
+<PubmedArticle>
+    <MedlineCitation Status="In-Data-Review" Owner="NLM">
+        <PMID Version="1">29279863</PMID>
+        <DateRevised>
+            <Year>2017</Year>
+            <Month>12</Month>
+            <Day>29</Day>
+        </DateRevised>
+        <Article PubModel="Electronic-eCollection">
+            <Journal>
+                <ISSN IssnType="Electronic">2373-2822</ISSN>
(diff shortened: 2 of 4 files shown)


peetucket:
may want to add another test here...

spec/fixtures/pubmed/pubmed_record_29279863.xml
```diff
@@ -0,0 +1,351 @@
...
+                        <Affiliation>Department of Pharmacology and Physiology, University of Rochester Medical Center, Rochester, NY 14642.</Affiliation>
+                    </AffiliationInfo>
+                </Author>
+            </AuthorList>
+            <Language>eng</Language>
+            <PublicationTypeList>
+                <PublicationType UI="D016428">Journal Article</PublicationType>
+            </PublicationTypeList>
+            <ArticleDate DateType="Electronic">
+                <Year>2017</Year>
```
atz: @peetucket This seems to be a more reliable representation of just the year than the language dependent `YYYY Mon-Mon` field targeted.  Is there a reason this isn't preferred?

peetucket: Seams a better fit.  Note sure if the DateType="Electronic" is important enough to select for, or if just /ArticleDate/Year is safe but I would probably go more broad to avoid missing a year with a different DateType.

peetucket: updated to look in that location and added another location that seems reasonable too -- updated to allow for multiple locations to be specified, with processing stopping at the first that one that appears to be a year

app/models/pubmed_source_record.rb
```diff
@@ -123,7 +123,19 @@ def source_as_hash(publication = Nokogiri::XML(source_data).xpath('//PubmedArtic
     record_as_hash[:author] = author_list.map { |a| author_to_hash(a) }.compact
 
     record_as_hash[:mesh_headings] = mesh_headings unless mesh_headings.blank?
-    record_as_hash[:year] = publication.xpath('MedlineCitation/Article/Journal/JournalIssue/PubDate/Year').text unless publication.xpath('MedlineCitation/Article/Journal/JournalIssue/PubDate/Year').blank?
+    year_xpaths = [
```
atz: 👍  I appreciate the enumeration of xpaths.

app/models/pubmed_source_record.rb
```diff
@@ -123,7 +123,19 @@ def source_as_hash(publication = Nokogiri::XML(source_data).xpath('//PubmedArtic
...
+      'MedlineCitation/Article/Journal/JournalIssue/PubDate/Year',
+      'MedlineCitation/Article/ArticleDate/Year',
+      'PubmedData/History/PubMedPubDate[@PubStatus="accepted"]/Year'
+    ]
+    # look for a year in all of the xpath locations above in order
+    #  stop after the first produces something that looks like a year
+    year_xpaths.each do |path|
+      if !publication.xpath(path).blank? && publication.xpath(path).text.match(/[12][0-9]{3}/)
```
atz: We can convert this to reuse the initial regex match though, like:
```ruby
next if publication.xpath(path).blank?
match = publication.xpath(path).text.match(/[12][0-9]{3}/)
next unless match
record_as_hash[:year] = match.to_s
break
```

Or more cleanly, we don't need to `break` the loop explicitly if we rely on how Nokogiri works and use `find`, like:
```ruby
match = year_xpaths.map { publication.xpath(path).text }.find { |x| x.match(/[12][0-9]{3}/}
```
We can count on a `blank?` Document node to still return `''` to `.text`, neither breaking nor matching the regex.  Example:
```ruby
> Nokogiri::XML::Document.new.xpath('foo').text
=> ""
```
I think _some_ form of not regex-ing twice seems useful.

peetucket: I think your strategy works -- see what I just pushed (your original suggestion just had some typos and missing characters)

PR #848, opened 2018-04-24
timeframes for updated/new author harvests and setting that determines if authors poller runs harvests on changed authors

### LINKED ISSUE #841: When running harvests from the nightly cap authors poller job, be sure to use a limited timeframe, and a different one for new vs updated authors
For new users, use 1 year
For updated users, use 4 weeks

### LINKED ISSUE #842: Add an option to only run harvests for new users from cap authors poller job
Currently we run an immediate harvest for all updated or new users from the nightly authors poller.  There are many updated users, and some of these updates could be minor.  Add a config option to easily disable immediate harvests for updated users.

peetucket:
also:
* cleanup some logging
* add default timeframes to settings
* use timeframes when harvesting updated or new authors in cap author poller

fixes #841
fixes #842

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -24,6 +24,7 @@ CAP:
   AUTHORSHIP_API_PATH: /cap-api/api/cap/v1/authors
   AUTHORSHIP_API_PORT: 443
   AUTHORSHIP_API_URI: https://cap-uat.stanford.edu
+  HARVEST_ON_CHANGE: true
 
 ## PubMed Auth Config
 PUBMED:
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -53,6 +53,8 @@ WOS:
   AUTH_CODE: secret
   LOG: log/web_of_science.log
   LOG_LEVEL: warn
+  update_timeframe: 4week # default WoS symbolicTimeSpan for updated authors/biweekly harvests
+  new_author_timeframe: 52week # default WoS symbolicTimeSpan for new authors
 
 DOI:
   BASE_URI: https://dx.doi.org/
@@ -60,8 +62,6 @@ DOI:
 HARVESTER:
   USE_MIDDLE_NAME: true
   USE_AUTHOR_IDENTITIES: false
-  UPDATE_TIMEFRAME: 4week # default WoS symbolicTimeSpan for updated authors/biweekly harvests
-  NEW_AUTHOR_TIMEFRAME: 52week # default WoS symbolicTimeSpan for new authors
   INSTITUTION:
     name: Stanford University
     address:
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -60,6 +60,8 @@ DOI:
 HARVESTER:
   USE_MIDDLE_NAME: true
   USE_AUTHOR_IDENTITIES: false
+  UPDATE_TIMEFRAME: 4week # default WoS symbolicTimeSpan for updated authors/biweekly harvests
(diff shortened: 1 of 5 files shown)


peetucket:
This test is failing but it may be due to test setup (where the author is created first and thus the check for an existing author is being triggered):
`rspec ./spec/lib/cap/authors_poller_spec.rb:130 # Cap::AuthorsPoller#process_record with an author retrieved from the CAP API harvests for new authors marked harvestable`

lib/cap/authors_poller.rb
```diff
@@ -212,19 +204,20 @@ def process_record_for_new_author(cap_profile_id, record)
...
-        queue_author_for_harvest(author)
+        queue_author_for_harvest(author,new_record: true)
         author.save!
         @new_author_count += 1
       end
 
       # Add author to job queue if it's harvestable and new/changed
       # @param [String] `skip_message` logs this message if it skips adding author to queue
-      def queue_author_for_harvest(author)
-        if (author.new_record? || author.changed?) && author.harvestable?
```
peetucket: NOTE:  author.new_record? will always returning false here, because the `process_record_for_new_author` method calls `Author.fetch_from_cap_and_create` which calls `Author.create!`.  This saves the record to the database.  Thus by the time we reach this method, we have no idea if the authors is new or not.  Hence passing this information in to the method.

lib/cap/authors_poller.rb
```diff
@@ -212,19 +204,20 @@ def process_record_for_new_author(cap_profile_id, record)
...
-          @new_or_changed_authors_to_harvest_queue << author.id
+      def queue_author_for_harvest(author,options)
+        if author.harvestable?
+          @new_authors_to_harvest_queue << author.id if options[:new_record]
+          @changed_authors_to_harvest_queue << author.id if Settings.CAP.HARVEST_ON_CHANGE && !options[:new_record] && author.changed?
         else
-          @no_sw_harvest_count += 1
-          logger.info "Author marked as not harvestable or did not change. Skipping Author.find_by(cap_profile_id: #{author.cap_profile_id})"
```
peetucket: seems overkill to log the fact that we didn't harvest because something did *not* change.  In fact, probably overkill even to log the fact that nothing was harvested

spec/lib/cap/authors_poller_spec.rb
```diff
@@ -128,14 +130,14 @@
       it 'harvests for new authors marked harvestable' do
         expect(author).to receive(:harvestable?).and_return(true)
         subject.process_record(author_record)
-        queue = subject.instance_variable_get('@new_or_changed_authors_to_harvest_queue')
+        queue = subject.instance_variable_get('@new_authors_to_harvest_queue')
```
peetucket: This test was only ever working before because we weren't differentiating between new or changed authors when building the queue.   Previously the queue was changing because because it was considered a changed author, not a new author.

config/settings.yml
```diff
@@ -59,6 +60,8 @@ DOI:
 HARVESTER:
   USE_MIDDLE_NAME: true
   USE_AUTHOR_IDENTITIES: false
+  UPDATE_TIMEFRAME: 4week # default WoS symbolicTimeSpan for updated authors/biweekly harvests
```
atz: These are attributes of WOS harvest only, not `HARVESTER`, right?

peetucket: yeah that's right

PR #849, opened 2018-04-25
do not bother throwing an expcetion in pubmed_additions if we cannot …

### LINKED ISSUE #799: pubmed_additions throws an exception for WOS records that exist from sciencewire harvesting
From https://app.honeybadger.io/projects/50046/faults/37129449

Due to the fact that the pubmed_additions method looks for publications only by the wos_uid column (see https://github.com/sul-dlss/sul_pub/blob/master/lib/web_of_science/process_pubmed.rb#L13), and it is possible for publications that were already harvested from Sciencewire to have a WOS source record but to *not* have the wos_uid column populated and still be considered match, the following line of code will not find the actual publication, and will generate an exception: 

https://github.com/sul-dlss/sul_pub/blob/master/lib/web_of_science/process_pubmed.rb#L16

This is because uid_to_pub will be nil and the call to `.first` blows up.

This is actually ok in this scenario, since the sciencewire publication already has the necessary pubmed additions, but we should either 

* catch this better and short-circuit so we don't keep throwing exceptions

OR

* use the matching_publication method instead to find the pub records (https://github.com/sul-dlss/sul_pub/blob/master/lib/web_of_science/process_records.rb#L141-L156) ... this would correctly find the sciencewire publication ... and this method should be available to us already without having to move it anywhere

peetucket:
…find the record using the wos_uid column in the pub database

fixes #799

### DIFF
--- a/lib/web_of_science/process_pubmed.rb
+++ b/lib/web_of_science/process_pubmed.rb
@@ -13,7 +13,8 @@ module WebOfScience
       uid_to_pub = Publication.where(wos_uid: present_recs.map(&:uid)).group_by(&:wos_uid)
       present_recs.each do |record|
         begin
-          pub = uid_to_pub[record.uid].first || raise("No Publication matches UID #{record.uid}")
+          next if uid_to_pub[record.uid].blank?
+          pub = uid_to_pub[record.uid].first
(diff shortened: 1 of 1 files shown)


peetucket:
add a test?

PR #850, opened 2018-04-27
Nuke sciencewire

### LINKED ISSUE #1004: Remove all sciencewire code no longer needed

atz:
We keep the sourcerecord model and the various classes related to parsing.  We get rid of the SW client, harvester, _et al._ with their related rake tasks and tests.  The sourcerecord class drops all the methods that had to do with remote retrieval.

Then the rest of the code, tests and VCR cassettes catch up with the fact those components are gone, mostly by removing the blocks that previously were gated by `Settings.SCIENCEWIRE.enabled` and converting cassettes to using WoS or Pubmed.

fixes #1004


peetucket:
Nice work, but propose we keep this around just a bit longer until things settle down with the WoS harvesting -- not necessarily because the Sciencewire client is likely not even an option anymore, but on the principle of don't change a lot of stuff all at once.

atz:
Yeah, I had some of this prepared ahead of time (the deletions), but a lot of the details were more complex to disentangle than expected, mostly because the use of VCR tied unrelated tests to ScienceWire.  I just don't want us to drift away from the project and come back in a year have there be a bunch of zombie deadweight code mucking everything up.

PR #860, opened 2018-05-02
any changes to number of alternate identities should trigger changed?…

peetucket:
… to be true on the author model, which triggers auto harvesting on author poller


atz:
Are you *extremely* certain this doesn't interfere with `ActiveRecord`'s `dirty?` semantics?  Because the Samvera community (us in particular) suffered for years due to half-replicating AR's expectations w/ `dirty?`.  I really don't want to repeat that experience.

peetucket:
@atz Updated as we discussed

PR #890, opened 2018-06-13
log enhancements for a complete harvest job

### LINKED ISSUE #894: Increase frequency of CAP harvesting and decrease lookback window
- [x] increase frequency in schedule.rb from bi-weekly to every 3 days
- [x] decrease lookback window to 7 days for regular harvest

peetucket:
Fixes #894

### DIFF
--- a/lib/harvester/base.rb
+++ b/lib/harvester/base.rb
@@ -7,13 +7,11 @@ module Harvester
     def harvest_all(options = {})
       total = authors_query.count
       count = 0
-      n = 0
       start_time = Time.zone.now
       logger.info "***** Started a complete harvest for #{total} authors at #{start_time}"
-      authors_query.find_in_batches(batch_size: batch_size).each do |batch|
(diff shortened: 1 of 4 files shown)


lib/harvester/base.rb
```diff
@@ -7,12 +7,18 @@ class Base
     def harvest_all(options = {})
       total = authors_query.count
       count = 0
-      logger.info "Started a complete harvest for #{total} authors at #{Time.zone.now}"
+      n = 0
+      start_time = Time.zone.now
+      logger.info "***** Started a complete harvest for #{total} authors at #{start_time}"
```
atz: This logging logic doesn't belong in the base class, IMO.  Also, you can use `each_with_index` instead of declaring and incrementing a variable yourself.

peetucket: Good point, switched to each_with_index.

Why don't you like this in the base class?  harvest_all is only implemented here, so i don't know where else to put global logging for a full scheduled harvest (which is very useful for monitoring)

PR #914, opened 2018-08-06
Update dependencies

SaravShah:


peetucket:
we have dependabot hooked up to this project that does dependency update PRs for us -- i've been merging those which has not put this PR into conflict --- not sure its worth resolving

PR #940, opened 2018-10-12
remove sciencewire lookup when adding by pmid

peetucket:
current production bug: https://app.honeybadger.io/projects/50046/faults/39602176#notice-summary

* sciencewire should *not* be in the path when looking up by and adding pubs by pmid
* short circuit a couple places where sw records are fetched, just in case that codepath is exercised somewhere else

we should also allocate some time time to review and merge #850

### DIFF
--- a/app/models/publication.rb
+++ b/app/models/publication.rb
@@ -57,7 +57,7 @@ class Publication < ActiveRecord::Base
   end
 
   def self.find_or_create_by_pmid(pmid)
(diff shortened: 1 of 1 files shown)


app/models/sciencewire_source_record.rb
```diff
@@ -93,11 +93,13 @@ def self.get_sciencewire_source_record_for_pmid(pmid)
   end
 
   def self.get_sciencewire_source_record_from_sciencewire(pmid)
+    return nil unless Settings.SCIENCEWIRE.enabled
```
atz: These method should be **removed** as in #850.  We're never enabling Sciencewire again, so gating on a setting suggests functionality that doesn't exist.

PR #1007, opened 2019-03-18
Update dependencies

jcoyne:


peetucket:
FYI - I believe all of the test failures are due to the `csl-styles` gem update.  This changed the format of the citations generated.  Given the fact that I have not heard we should be updating citation formats going forward, I think we should pin to the current version with a comment indicating that updating this gem will change the format of citations and we should only do this with stakeholder feedback.

jcoyne:
@peetucket pinning csl-styles was just the thing. Build passed.

jcoyne:
comment added

PR #1012, opened 2019-03-18
Instructions and changes for running test with MySQL in Docker.

justinlittman:


config/database.yml
```diff
@@ -22,10 +22,14 @@ default: &default
 development:
   <<: *default
   database: sulbib_development
+  host: 127.0.0.1
```
jcoyne: can we put these in default?

README.md
```diff
@@ -40,6 +40,16 @@ bundle exec rake db:create
 bundle exec rake db:migrate
 ```
 
+Alternatively, you can use Docker:
+```
+docker run --rm -e MYSQL_ALLOW_EMPTY_PASSWORD=true -p 3306:3306 -d mysql:8
+```
+and then:
+```
+bundle exec rake db:create RAILS_ENV=test
```
jcoyne: The modern way to do this is `RAILS_ENV=test ./bin/rails db:create`

jcoyne: but, we aren't using rails 5 yet. Sigh.

config/database.yml
justinlittman: Yes. I had thought default was using unix sockets, but mistaken.

PR #1014, opened 2019-03-19
Added comparing existing author identities against imported author id…

### LINKED ISSUE #974: Auto detection of changes to alternate identities for immediate harvesting
Currently when a user changes to their primary profile information, this change is detected via a nightly job, via an API call from the library to the Profiles system.  This in turn triggers an immediate harvest to pick up any new potential publications.  We are also able to detect if a user adds or removes an entry in the alternate identities import settings nightly (i.e. adds or removes a new name or institution) and trigger an immediate harvest.  However, if a user simply changes one of these rows, we are not able to detect this change and do not run an immediate harvest.  This is because:

The Profiles API call we make to request “changed” profiles on a nightly basis returns a large number of rows regardless of the “change” timeframe requested.  For example, we run a job each night requesting profiles changed within the last 24 hours, but routinely receive > 4000 rows returned, which doesn’t seem correct.  
The rows for alternate names in the import settings are not uniquely identified or ordered,, and so we are unable to detect if any particular row is actually different.  Currently, we just replace all all of the alternate identities for any user that the Profiles API indicates is different and count the rows -- if the number is different than what we started with, we know a new identity has been added or removed and a harvest is needed.  

Since we can’t re-harvest many profiles each night, we do what we can to confirm if the main part of the profile settings (name and email) have changed or if the total number of alternate identities has changed and harvest only those profiles.  

In order to be more intelligent about this, we either:

- Need the Profiles API to return only profiles that have actually changed (to ensure this is a more realistic and smaller number).  We can then just re-harvest all of these profiles without worrying about what part of the profile or import settings have changed.
- Need the Profiles API to mark each alternate identity row with a unique identifier, so we can use this to track changes to specific rows, and trigger a harvest if any of one of the alternate identity rows have changed.
- Use a composite key on our alternate identities that is recomputed on each update, which can be used to detect if the data has changed.

Requested for FY2019

some potentially useful work done in #794

justinlittman:
…entities.

closes #974

### DIFF
--- a/app/models/author.rb
+++ b/app/models/author.rb
@@ -1,3 +1,5 @@
+require 'set'
+
 class Author < ActiveRecord::Base
   has_paper_trail on: [:destroy]
   validates :cap_profile_id, uniqueness: true, presence: true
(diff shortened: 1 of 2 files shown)


justinlittman:
@peetucket requested your review because want to make sure that this satisfied the tix requirements.

peetucket:
I believe we can close WIP #794 if we accept this approach -- please take a quick look to see.

PR #1024, opened 2019-03-25
Made pubmed a full-fledged harvester.

### LINKED ISSUE #973: Harvest publications from Pubmed Central by querying via name and institution
Topline harvesting from Pubmed, similar to WoS.  Requested for FY 2019

- [x] The PubMed client needs to have the ability to query PubMed for publications by name and institution and return an array of results.
- [x] The PubMed harvester needs to be refactored and subclassed from the Base::Harvester class, and have the relevant methods added to perform regular harvests.
- [x] Move the PubMed harvester to a sub-folder to match WoS.
- [x] De-duplication code needs to be written and/or confirmed as working to ensure duplicate publications (i.e. same PMID) are not added when harvesting is performed using WoS and PMID.
- [x] Logging needs to be added for PubMed harvester.
- [x] New tests need to be written for PubMed harvesting.
- [x] Scheduling of regular Pubmed harvesting needs to be added to the cron jobs via the schedule.rb file (same as WoS regular harvesting).
- [x] We need to be hook into the Pubmed harvesting from the nightly cap author polling (where we harvest nightly when authors are added or changed) -- may want to create a new `do_harvest` method in `AuthorsPoller` class which encompasses both WoS and Pubmed and then call this method instead of the WoS only method when we are ready to harvest
- [ ] Update wiki pages with new rake tasks for harvesting
- [ ] Smoke test

Resources:
- [Pubmed API Documentation](https://www.ncbi.nlm.nih.gov/books/NBK25501/) 
- [Searching a database with API](https://www.ncbi.nlm.nih.gov/books/NBK25500/#chapter1.Searching_a_Database)
- [Query syntax](https://www.ncbi.nlm.nih.gov/books/NBK25499/)

Example search by name query for Pubmed (post):
```
https://eutils.ncbi.nlm.nih.gov/entrez/eutils/esearch.fcgi?db=pubmed&retmode=xml&api_key=ACTUAL_API_KEY_HERE&field=author&term=Casciotti, Karen OR Casciotti, K
```

So we we can use the `Settings.PUBMED.BASE_URI` but need a new setting analogous to the `FETCH_PATH`, and then need to create a new query in `PubmedClient` to send the query above, which returns PMIDs, for which we can then get full records using the existing `.fetch_records_for_pmid_list` method in the client.

justinlittman:
closes #973

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -33,6 +33,10 @@ PUBMED:
   API_KEY: a_fake_pubmed_key
   BASE_URI: https://eutils.ncbi.nlm.nih.gov
   ARTICLE_BASE_URI: https://www.ncbi.nlm.nih.gov/pubmed/
+  # In number of days
+  regular_harvest_timeframe: 21
+  update_timeframe: null # default pubmed reldate for updated authors resulting from nightly cap polling
+  new_author_timeframe: null # default pubmed reldate for new authors resulting from nightly cap polling (null == for all time)
 
 ## Sciencewire Auth Config
 SCIENCEWIRE:
--- a/lib/cap/authors_poller.rb
+++ b/lib/cap/authors_poller.rb
@@ -37,6 +37,7 @@ module Cap
       logger.info 'new authors to harvest: ' + @new_authors_to_harvest_queue.to_s
       logger.info 'changed authors to harvest: ' + @changed_authors_to_harvest_queue.to_s if Settings.CAP.HARVEST_ON_CHANGE
       do_wos_harvest
+      do_pubmed_harvest
       log_stats
       logger.info 'Finished authorship import'
     rescue => e
@@ -55,6 +56,15 @@ module Cap
       end
       Author.where(id: @changed_authors_to_harvest_queue).find_in_batches(batch_size: 250) do |authors|
         WebOfScience.harvester.harvest(authors, symbolicTimeSpan: Settings.WOS.update_timeframe)
+      end
+    end
+
+    def do_pubmed_harvest
+      Author.where(id: @new_authors_to_harvest_queue).find_in_batches(batch_size: 250) do |authors|
+        Pubmed.harvester.harvest(authors, reldate: Settings.PUBMED.new_author_timeframe)
+      end
+      Author.where(id: @changed_authors_to_harvest_queue).find_in_batches(batch_size: 250) do |authors|
+        Pubmed.harvester.harvest(authors, reldate: Settings.PUBMED.update_timeframe)
       end
     end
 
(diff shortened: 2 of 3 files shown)


justinlittman:
So I tried to closely follow the pattern of WOS harvester.

lib/pubmed/harvester.rb
```diff
@@ -0,0 +1,95 @@
+module Pubmed
+  # Application logic to harvest publications from Pubmed;
+  # This is the bridge between the Pubmed API and the SUL-PUB application.
+  # This class is responsible for processing Pubmed API response data
+  # to integrate it into the application data models.
+  class Harvester < ::Harvester::Base
+    # # @param [Enumerable<Author>] authors
+    # # @param [Hash] options
+    # # @return [void]
+    # def harvest(authors, options = {})
```
jcoyne: Cruft?

lib/web_of_science/harvester.rb
```diff
@@ -4,22 +4,22 @@ module WebOfScience
...
-        author_success += 1
-      end
-      logger.info("#{self.class} - completed harvest - #{author_success} of #{count} processed")
-    rescue StandardError => err
-      NotificationManager.error(err, "harvest(authors) failed - #{author_success} of #{count} processed", self)
-    end
+    # # @param [Enumerable<Author>] authors
+    # # @param [Hash] options
+    # # @return [void]
+    # def harvest(authors, options = {})
```
jcoyne: Cruft?

config/settings.yml
```diff
@@ -30,9 +30,12 @@ CAP:
 PUBMED:
   LOG: log/pubmed.log
   FETCH_PATH: /entrez/eutils/efetch.fcgi?db=pubmed&retmode=xml
+  SEARCH_PATH: /entrez/eutils/esearch.fcgi?db=pubmed&retmode=xml&retmax=100000
   API_KEY: a_fake_pubmed_key
   BASE_URI: https://eutils.ncbi.nlm.nih.gov
   ARTICLE_BASE_URI: https://www.ncbi.nlm.nih.gov/pubmed/
+  # In number of days
+  regular_harvest_timeframe: 21
```
peetucket: Can we set this value to 'nil' or something else to eliminate the time filtering constraint?  This would allow us to mimic the behavior of the WoS harvesting, which does a full harvest for all time for new or updated users.  

e.g. from settings.yml, these are the WoS values that used in the `lib/cap/authors_poller.rb` class that we will need to add to

 ``` 
update_timeframe: null # default WoS symbolicTimeSpan for updated authors resulting from nightly cap polling
new_author_timeframe: null # default WoS symbolicTimeSpan for new authors resulting from nightly cap polling (null == for all time)
```

justinlittman: This setting is currently only used in `harvest_authors_update` rake task, so it won't act as a default elsewhere.

justinlittman: I added pubmed to `lib/cap/authors_poller.rb` with suggested defaults.

jcoyne:
@justinlittman this now conflicts with master. Can you rebase please?

justinlittman:
Something went badly wrong ...

PR #1025, opened 2019-03-25
load all dependencies in lib directory by default

peetucket:
fix a problem where rails 5 does not load all dependencies by default in production

seems to auto-load all subfolders as well with explicitly setting that

see https://blog.bigbinary.com/2016/08/29/rails-5-disables-autoloading-after-booting-the-app-in-production.html


config/application.rb
```diff
@@ -11,10 +11,8 @@ class Application < Rails::Application
     # Initialize configuration defaults for originally generated Rails version.
     config.load_defaults 5.0
 
-    # Custom directories with classes and modules you want to be autoloadable.
-    # config.autoload_paths += %W(#{config.root}/extras)
-    config.autoload_paths += %W(#{config.root}/lib)
-    config.autoload_paths += Dir["#{config.root}/lib/**/"]
+    config.enable_dependency_loading = true
```
jcoyne: I believe this is not thread-safe.  Can we do `eager_load_paths` on the line below instead?

peetucket: yes, done and rebased

peetucket:
Also, I moved `lib/bibtex/bibtex_identifiers.rb` --> root of `lib` folder since its current folder location does not match up to the fact that it isn't in a module and therefore was not autoloading

config/application.rb
jcoyne: Okay, but now we don't need `enable_dependency_loading`

peetucket: removed

PR #1034, opened 2019-04-03
harvest all sources for better efficiency

peetucket:
- create a new harvester class subclassed from the base class that runs `process_author` for both WoS and Pubmed ...  running through the main `harvest_all` loop in the harvest base class only once for both sources will improve efficiency since we only query and iterate through all authors once (as opposed to querying and iterating over all others once for WoS and once for Pubmed for all authors)
- refactor the cap author poller class to take advantage of this new class instead of calling each source harvester separately
- create new rake tasks for harvesting authors for both sources
- update the cron jobs used in `schedule.rb` to use this new rake task 
- update the `author_harvest_job` class to use the new all sources class
- remove the 'harvest-alternate-names' parameter from the immediate harvest authors controller API call, it is not supported and wasn't working anyway
- some small clarifications to inline method documentation

### DIFF
--- a/app/controllers/authors_controller.rb
+++ b/app/controllers/authors_controller.rb
@@ -5,7 +5,7 @@ class AuthorsController < ApplicationController
   # request an immediate harvest of this user's profile
   # POST /authors/:cap_profile_id/harvest.json
   def harvest
-    if AuthorHarvestJob.perform_later(author_params[:cap_profile_id], harvest_alternate_names: alt_names)
+    if AuthorHarvestJob.perform_later(author_params[:cap_profile_id])
       render json: {
         response: "Harvest for author #{params[:cap_profile_id]} was successfully created."
       }, status: :accepted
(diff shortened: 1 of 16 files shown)


justinlittman:
@peetucket Can you explain why the efficiency justifies the additional complexity? How expensive is querying authors twice?

Not trying to be difficult, but I think we should be aiming to reduce the size and complexity of this codebase.

peetucket:
> @peetucket Can you explain why the efficiency justifies the additional complexity? How expensive is querying authors twice?

There are a few reasons:

1. The most common harvesting scenario is to harvest an author for all available sources -- having to run multiple methods/rake tasks for each other seems counter-intuitive.  I think it makes more sense to run a single "harvest" job for an author and have it encapsulate the complexity of the fact that there are multiple sources.
2. If we end up adding another source in the future (e.g. dimensions), the call to have this new source be part of a regular author harvest gets added in only one location (the new `AllSources` class) instead of having to add it to the all the places where an author harvest is triggered (regular cron job, the authors poller, an API call).
3. There are ~32k authors authors that are iterated over every three days.  This will iterate over them once instead of twice.  Not sure how expensive it is - I didn't mention reasons 1 & 2 in my PR, but now that I write them up, I feel they more important than 3.

peetucket:
Note, this PR missed one change in the `AuthorHarvestJob`, I need to work on this still.  It also needs to be done as a PR into master if we continue.

peetucket:
Also open to other simpler ideas to achieve this...

justinlittman:
Thanks. Agree that 1 and 2 are more compelling than 3.

PR #1039, opened 2019-04-04
prefer pubmed central when searching by pmid

peetucket:
Seems odd that WoS is the preferred source over Pubmed when searching manually by PMID. This would swap the preferred order, though without specific stakeholder guidance, reluctant to push forward.

Would be good to validate this approach with the Profiles team.  Perhaps we want WoS first since most publications are WoS in the database?  Still seems odd.


peetucket:
Speculative improvement - unclear if it makes any.

PR #1042, opened 2019-04-05
log all_sources harvesting into own log file

peetucket:
otherwise all sources info gets logged into default rails logger which is very noisy

also, this makes the all_sources harvester analogous to the WoS and Pubmed harvester in the way it is called.

also move log_info method to the base harvester class to avoid duplication

### DIFF
--- a/lib/all_sources.rb
+++ b/lib/all_sources.rb
@@ -1,0 +1,16 @@
+# rubocop:disable Style/ClassVars
+
+# All Sources harvesting utilities
+module AllSources
+
+  # @return [AllSources::Harvester]
+  def self.harvester
+    @@harvester ||= AllSources::Harvester.new
+  end
+
+  def self.logger
+    @@logger ||= Logger.new(Settings.HARVESTER.LOG)
+  end
+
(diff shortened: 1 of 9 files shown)


lib/all_sources.rb
```diff
@@ -0,0 +1,16 @@
...
+# All Sources harvesting utilities
+module AllSources
+
+  # @return [AllSources::Harvester]
+  def self.harvester
+    @@harvester ||= AllSources::Harvester.new
+  end
+
+  def self.logger
+    @@logger ||= Logger.new(Settings.HARVESTER.LOG)
```
jcoyne: Does this logger need to rotate?  Is that something we should do here or something we need to have ops do?

peetucket: ops handles all log rotations, but i believe i need to let them know about the new file that is being generated (though this one will be quite small)...I'll do that

peetucket: log rotation confirmed with ops - it will happen based on the .log extension of the file and its location (alongside other log files)

PR #1047, opened 2019-04-24
fix issue with our content-type response being incorrect

peetucket:
- two API calls convereted from grape to rails were incorrectly responding with content-type:text/plain instead of application/json causing a production issue with profiles
- don't need to protect from forgery for API only controllers (causing warnings in logs)


justinlittman:
Approve pending green.

PR #1050, opened 2019-04-29
show last deploy time in revision check of okcomputer

peetucket:
potentially useful when debugging to see when the app was last deployed easily in the status check area


jcoyne:
This info is already in honeybadger. Do we need it in a non-standard place too?

peetucket:
you're right, never mind, not really needed

PR #1053, opened 2019-05-06
saving the pub should come outside the find_or_create_by block

### LINKED ISSUE #1052: Blank authorships being created for pubmed harvesting
Since ~April 23, a large number of publications are being created with a blank `authorship` part of the pub_hash, even though there is a contribution associated with the publication.  This coincides with an error being reported by the Profiles team when updating publications on their end.

peetucket:
fixes #1052 

When creating new publications, the `Author#assign_pub` method doesn't correctly update the pub_hash due to the save calls being inside the `find_or_create_by` contribution block.  I have confirmed this on my laptop manually by testing the Pubmed harvesting code with the change in this PR (it creates the authorship block in the pub_hash correctly) and without this change (it fails to add authorship just like in prod).

This isn't a problem with WoS harvesting since the creation of new publications occurs in a different point in the code and doesn't rely on the `Author#assign_pub` method:
https://github.com/sul-dlss/sul_pub/blob/master/lib/web_of_science/process_records.rb#L93-L108

- [ ] need a test for this?

### DIFF
--- a/app/models/author.rb
+++ b/app/models/author.rb
@@ -171,9 +171,9 @@ class Author < ActiveRecord::Base
         cap_profile_id: cap_profile_id,
         featured: false, status: 'new', visibility: 'private'
       )
-      pub.pubhash_needs_update! # Add to pub_hash[:authorship]
-      pub.save! # contrib.save! not needed
     end
+    pub.pubhash_needs_update! # Add to pub_hash[:authorship]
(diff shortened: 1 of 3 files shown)


justinlittman:
@peetucket Assigning to you to review unit test.

peetucket:
> @peetucket Assigning to you to review unit test.

looks good, thanks for the doing that

Issue #1057, opened 2019-05-15
Cleanup/remedidate large number of pubmed publications created in May 9 and 13 2019 harvests


peetucket:
PR that has a rake task to purge these publications here: #1058 
Underlying issue ticket: #1059

Since we deployed code that was merged on May 6-7 (v3.2.4 and v3.2.5) there were lots and lots of new publications being created with harvests, which seems very wrong.

```
Contribution.where('created_at > ?','2019-05-14').size
=> 73777

Publication.where('created_at > ?','2019-05-14').size
=> 73000
```

With nearly all of these being "pubmed" provenance (and likely from the pubmed harvester).  You can tell where publications came from by examining the pub_hash[:provenance] attribute in the each row (yes, you need to iterate, its not a top level model attribute).  If the associated contribution has a status of `new`, it likely came from a harvest (since manually entered publications are automatically are set to status of `approved`).

```
Contribution.where('created_at > ? and status=? and visibility=?',Time.parse('05-05-2019'),'new','private').size
=> 218937
```

peetucket:
Not sure how to fix this data, as rolling back the database will destroy manual publication entries too that occurred during this time.  Could possibly:

1. Look at all new contributions created since May 6 with a status of 'new' and visibility of 'private'.
2. Grab all publication IDs.
3. For any new publications with these IDs that were also created since May 6, destroy the publication. 
4. For those publications created earlier, just destroy the contribution, but update the publication's pub_hash

peetucket:
How to find authors with lots of new publications in a certain timeframe:

```
cutoff_pubs = 200 # more than this number of new pubs is a problem
start_timeframe = Time.parse('April 25 2019')
end_timeframe = Time.parase('May 14 2019')

authors = Contribution.where('status = ? and visibility = ? and created_at > ? and created_at < ?', 'new', 'private', start_timeframe, end_timeframe).pluck(:author_id).uniq;nil

author_info = []
authors.each do |author_id|
  author = Author.find(author_id)
  new_pubs = author.contributions.where('status = ? and visibility = ? and created_at > ? and created_at < ?', 'new', 'private', start_timeframe, end_timeframe).size
  if new_pubs > cutoff_pubs
     author_info << {cap_profile_id: author.cap_profile_id,name:"#{author.last_name}, #{author.first_name}",new_pubs:new_pubs,num_authors_with_same_initial:num_authors_with_same_initial}
  end
end;nil
```

justinlittman:
[...]

PR #1058, opened 2019-05-15
cleanup task to remove pubs and contribs harvested after a date

peetucket:
- proposed cleanup task for bad harvest (run per author)
- log the pub ids we deleted so we can provide to profiles team
- destroy dependent models when we delete a publication (which is very rare and only happens in remediations like this, but if we don't destroy the associated WebOfScienceSourceRecord for example, mysql will throw a unique key constraint violation)

see https://github.com/sul-dlss/sul_pub/issues/1057#issuecomment-492480310 for logic

### DIFF
--- a/lib/tasks/cleanup.rake
+++ b/lib/tasks/cleanup.rake
@@ -1,3 +1,5 @@
+require 'csv'
+
 namespace :cleanup do
   desc 'Merge contributions FROM duped_cap_profile_id INTO primary_cap_profile_id'
   # Use case: a single author has two author rows with publications associated with each.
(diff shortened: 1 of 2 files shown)


lib/tasks/cleanup.rake
```diff
@@ -124,4 +124,56 @@ namespace :cleanup do
...
+
+ desc 'Remove all new contributions and related publications with no other contributions after a specific date -- rarely used for bad harvests'
+  # Use case: a harvest goes awry and you need to cleanup the contributions/publications that resulted.  This should be rare.
+  #  Input a date and a provenance.  Any publications created with that provenance after or equal to that date that also had a single
+  #  contribution created and whose status is 'new' (to avoid manually added publications) after that date will be destroyed, along with the related source record.
+  #  NOTE: you should only do this if CAP has not yet updated on their end or else you will be out of sync.
+  # RAILS_ENV=production bundle exec rake cleanup:harvest['May 6 2019','pubmed'] # will remove all contributions in the 'new' state after the given date and with the given provenance
+  task :harvest, [:start_date, :provenance] => :environment do |_t, args|
+
+    start_date = args[:start_date]
```
jcoyne: Can this code be moved into a class that the rake task calls?  That would make it easier to test

peetucket: makes sense, sort of hoping that is something that only needs to be run this one time though, and also just looking for general feedback on this approach to cleanup of potential bad publications having been created

PR #1060, opened 2019-05-16
Added handling for institutions that have ampersands when querying pu…

### LINKED ISSUE #1059: Determine why pubmed harvester is creating lots of publications
Related to #1057

justinlittman:
…bmed.

closes #1059

### DIFF
--- a/spec/lib/pubmed/query_author_spec.rb
+++ b/spec/lib/pubmed/query_author_spec.rb
@@ -37,5 +37,37 @@ describe Pubmed::QueryAuthor do
         expect(query_author.send(:term)).to eq('((Altman Russ[Author]) OR (Altman R[Author])) AND (Stanford University[Affiliation])')
       end
     end
+
+    context 'with a user with alternate identities with different institutions' do
+      before do
+        AuthorIdentity.create(
+          author: author,
+          first_name: 'R',
+          middle_name: 'B',
+          last_name: 'Altman',
+          institution: 'Amherst College'
+        )
+      end
+
+      it 'generates the correct term string' do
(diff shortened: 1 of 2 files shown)


spec/lib/pubmed/query_author_spec.rb
```diff
@@ -37,5 +37,37 @@
...
+          author: author,
+          first_name: 'R',
+          middle_name: 'B',
+          last_name: 'Altman',
+          institution: 'William & Mary'
+        )
+      end
+
+      it 'generates the correct term string' do
+        expect(query_author.send(:term)).to eq('((Altman Russ[Author]) OR (Altman R[Author])) AND (Stanford University[Affiliation] OR William Mary[Affiliation] OR William and Mary[Affiliation])')
```
jcoyne: Are we confident the extra `and` isn't going to screw up the query parser?   Is there a way to designate "William and Mary" is one token rather than 3?

justinlittman: Determined this approach by testing as described in #1059, which isn't to say that there are not other ways of formulating the query.

jcoyne: I see we're doing `\"oregon\"[MeSH Terms]` would it be possible to do `\"William & Mary\"[Affiliation]`?

justinlittman: I tried various variations with escaped quoting, escaping the ampersand, and using parentheses to no avail.

jcoyne: Thanks @justinlittman

peetucket:
```
author=Author.find(37959)
WebOfScience::QueryAuthor.new(author).send(:institutions)
=> ["stanford", "oregon health & science", "washington"]
```

PR #1061, opened 2019-05-16
pubmed harvest disable option; tweak to search by pmid in pub model first

peetucket:
- allow Pubmed harvesting to be disabled with a setting flag (default to true, override in specific environment as needed)
- when searching for existing pubmed pubs via pmid, search pmid column in publication model directly first to improve performance and ensure we find pubs that may not have the associated publication_identifiers values (legacy pubs)
- small tweak to remediation task to store pmids we are processing before starting to delete them in case something timesout in the middle and crashes
- remove now unused sciencewire enabled flag

If accepted and deployed, we should temporarily override the new `PUBMED.harvest_enabled` option and set to `false` in both -prod and -stage settings.yml on https://github.com/sul-dlss/shared_configs until we can find why pubmed harvests produced so many new publications

### DIFF
--- a/app/models/publication.rb
+++ b/app/models/publication.rb
@@ -67,9 +67,10 @@ class Publication < ActiveRecord::Base
   end
 
   def self.find_by_pmid_pub_id(pmid)
-    Publication.includes(:publication_identifiers)
-               .where("publication_identifiers.identifier_type": 'pmid', "publication_identifiers.identifier_value": pmid)
-               .find_by('wos_uid IS NOT null OR sciencewire_id IS NOT NULL OR pmid IS NOT null')
+    find_by_pmid(pmid) ||
+      Publication.includes(:publication_identifiers)
+                 .where("publication_identifiers.identifier_type": 'pmid', "publication_identifiers.identifier_value": pmid)
+                 .find_by('wos_uid IS NOT null OR sciencewire_id IS NOT NULL OR pmid IS NOT null')
   end
 
   # Queries publication_identifiers (like the poorly-named find_by_xxx methods above)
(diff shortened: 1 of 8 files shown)


app/models/publication.rb
```diff
@@ -67,9 +67,10 @@ def self.find_by_doi(doi)
   end
 
   def self.find_by_pmid_pub_id(pmid)
-    Publication.includes(:publication_identifiers)
-               .where("publication_identifiers.identifier_type": 'pmid', "publication_identifiers.identifier_value": pmid)
-               .find_by('wos_uid IS NOT null OR sciencewire_id IS NOT NULL OR pmid IS NOT null')
+    find_by_pmid(pmid) ||
```
peetucket: why not just quickly look to see if we can find the publication via the `.pmid` column in the publication model itself before digging into PublicationIdentifier association?

config/settings.yml
```diff
@@ -29,6 +29,7 @@ CAP:
 ## PubMed Auth Config
 PUBMED:
   LOG: log/pubmed.log
+  harvest_enabled: true # this only enables/disabling regular harvests, it will still allow manual entries via pubmed and record augmentation
   FETCH_PATH: /entrez/eutils/efetch.fcgi?db=pubmed&retmode=xml
```
aaron-collier: Should this be `enabled` to stay consistent with WOS?

peetucket: right now its not consistent though...if you set WOS.harvest to disabled, then you aren't allowing any WOS client connections at all (not just harvesting), but pubmed disabled is only for harvesting...not ideal but a stopgap to kill just Pubmed harvesting while allowing single record augmentation

PR #1066, opened 2019-05-20
ensure all & is stipped from pubmed queries

### LINKED ISSUE #1063: Author identities with & still have a problem with pubmed queries
#1059 did not completely fix the issue

there are permutations of institutions with an & with different spacing that causes problems.

```
author=Author.find_by(cap_profile_id:'63608')
author.author_identities  # note name redacted for this ticket comment
=> [#<AuthorIdentity:0x0000000007db59a0
  id: 1496227,
  author_id: 49036,
  first_name: "First Name",
  middle_name: nil,
  last_name: "Last Name",
  institution: "Texas A&M University",
  start_date: Tue, 01 Sep 2009,
  end_date: Wed, 31 Dec 2014,
  created_at: Sat, 20 Apr 2019 11:53:16 UTC +00:00,
  updated_at: Sat, 20 Apr 2019 11:53:16 UTC +00:00>]
pmids=Pubmed::QueryAuthor.new(author, {}).pmids
==> 100000
query = Pubmed::QueryAuthor.new(author, {});nil
query.send(:term)
=> "((First Name Last Name[Author])) AND (Stanford University[Affiliation] OR Texas A&M University[Affiliation] OR Texas AandM University[Affiliation])"
```

peetucket:
fixes #1063 by just blunt force changing & to and everywhere with no other variants

### DIFF
--- a/lib/pubmed/query_author.rb
+++ b/lib/pubmed/query_author.rb
@@ -37,14 +37,11 @@ module Pubmed
     end
 
     def affiliation_terms(institution)
-      terms = []
       if institution.include?('&')
-        terms << "#{institution.gsub('& ', '')}[Affiliation]"
-        terms << "#{institution.gsub('&', 'and')}[Affiliation]"
+        "#{institution.gsub('&', 'and')}[Affiliation]"
       else
-        terms << "#{institution}[Affiliation]"
+        "#{institution}[Affiliation]"
       end
-      terms
     end
 
     def parse_response(response)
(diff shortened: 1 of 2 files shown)


peetucket:
this PR just strips the & completely with very little logic, it only produces a single variant for each institution based on what the user entered (previously, we were producing two variants for each institution, but leaving some cases with an &, causing problems, see #1063).

definitely open to more intelligent and better way of doing this that ensures nothing is left with an &

PR #1072, opened 2019-05-29
do not save publications if not needed when assigning a publication

### LINKED ISSUE #1071: Do not save publications if no changes occurred
The #assign_pub method in the `author` model (https://github.com/sul-dlss/sul_pub/blob/master/app/models/author.rb#L167-L181) saves the publication each time even if no contributions were added.  This leads to lots of updated publications during Pubmed harvesting (because of this method, which is called each time: https://github.com/sul-dlss/sul_pub/blob/master/lib/pubmed/harvester.rb#L41) which may causes problems with the Profiles API calls (since they have big updates each time).

peetucket:
fixes #1071

### DIFF
--- a/app/models/author.rb
+++ b/app/models/author.rb
@@ -170,14 +170,17 @@ class Author < ActiveRecord::Base
         return
     end
     raise 'Author must be saved before association' unless persisted?
-    pub.contributions.find_or_create_by!(author_id: id) do |contrib|
+
+    contribution = pub.contributions.find_or_initialize_by(author_id: id) do |contrib|
       contrib.assign_attributes(
         cap_profile_id: cap_profile_id,
         featured: false, status: 'new', visibility: 'private'
       )
     end
-    pub.pubhash_needs_update! # Add to pub_hash[:authorship]
-    pub.save! # contrib.save! not needed
+    return true unless contribution.new_record?
+    contribution.save!
+    pub.pubhash_needs_update!
+    pub.save!
   end
 
   private
--- a/app/models/author.rb
+++ b/app/models/author.rb
@@ -177,10 +177,11 @@ class Author < ActiveRecord::Base
         featured: false, status: 'new', visibility: 'private'
       )
     end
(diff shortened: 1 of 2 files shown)


app/models/author.rb
```diff
@@ -170,14 +170,17 @@ def assign_pub(pub)
...
+
+    contribution = pub.contributions.find_or_initialize_by(author_id: id) do |contrib|
       contrib.assign_attributes(
         cap_profile_id: cap_profile_id,
         featured: false, status: 'new', visibility: 'private'
       )
     end
-    pub.pubhash_needs_update! # Add to pub_hash[:authorship]
-    pub.save! # contrib.save! not needed
+    return true unless contribution.new_record?
```
jcoyne: This method is documented as returning a `Contribution`, but here you're returning a `Boolean`.

spec/models/author_spec.rb
```diff
@@ -195,6 +195,16 @@
...
     end
 
+    it 'does not create contrib or save publication when not needed' do
+      # assign a pub
+      subject.assign_pub(pub)
+      expect(subject.contributions.count).to eq(1)
+      # assign it again, pub should not change!
+      subject.assign_pub(pub)
+      expect(subject.contributions.count).to eq(1)
+      expect(pub).not_to be_changed
```
jcoyne: `changed?` always returns false after a save, so I'm not sure what you're trying to test for here.

app/models/author.rb
peetucket: ah yes, missed that -- turns out it wasn't actually returning a contribution before either (and the code doesn't care so it's never using the return value anyway), but fixed it to match the signature

spec/models/author_spec.rb
peetucket: interesting...didn't realize this - made an adjustment to the test (basically trying to confirm that pub does not get re-saved if we re-assign that publication, so that is causing unnecessary touches of the publication updated_at)

PR #1076, opened 2019-06-03
add a script to update the publication.updated_at column based on the last_contribution_date

peetucket:
fixes data problems created by #1071


peetucket:
Shouldn't be needed anymore, its been a while since we've been creating bad data.

Issue #1081, opened 2019-06-11
Consider other ways of constructing name queries


peetucket:
We have tried to write our queries to be as inclusive as possible for as many users as possible, but the algorithm doesn't always work out.  For example, Dr James Chen has data like this:

Primary identity = James K. Chen
Alternate identity  = J K Chen

We then produce all of the variants of "Last, First", "Last, First, Middle" and "Last, First, Middle Initial" for all of the entered names (just two in his case) and "or" them together.  For him, this looks like:  "Chen,James" OR "Chen,James,K." OR "Chen,James,K" OR "Chen,J" OR "Chen,J,K"  

If we were to skip the "Last, First" variant (i.e. dropping the middle name), this would miss cases were people have entered a middle name in their profile, but then don't publish with it.  

For example, "Jackson, Samuel Paul" may publish under "Jackson, Samuel" … so to get these publications, we would need a variant that doesn't include the middle name or else we'd miss those publications.  

In cases like Professor Chen, since he had to use "J" as his first name in the alternate identity, our algorithm then adds this variant without the middle initial.

One possibility is to *not* add this variant without the middle name for just the alternate identities (but leave it for the primary identity).  This may still present cases where the user is better served by it.

peetucket:
Other ideas: 

* conditionally skip the search that removes middle name / initial when authors have specified the middle initial specifically in their pub import settings.
* adding a flag in the settings in Profiles that the user can check if they want the pub import / harvest to be limited to the specified name and then alter the search criteria to not add/remove variants of the name

peetucket:
Notes from Grace: 

```
WoS does not use commas in the author index and also only uses initials for first and middle name.   Only one citation had his middle initial.   So here’s the format you should use:
 
AU=(McKeown N OR McKeown NW) AND (AD=Stanford Univ OR OG=Stanford University)
```

PR #1082, opened 2019-06-11
do not remove middle name variants from the query when the user has a middle name entered

peetucket:
See #840 for a different approach to query alteration.  They both touch the same code.

Possible option for #1081 	

Rather than blindly remove the middle name variant for all queries, let's only remove if the user really has not entered a middle name.  It should improve results for users with middle names.

Requires some coordination with Profiles team to validate the approach


peetucket:
Not sure this makes sense

PR #1107, opened 2019-10-04
do not fail if a user has no valid alt identities

peetucket:
if a user has only one alternate identity and has no first name (or something like a dot for a first name), we will end up with a nil in the `names` method of wos `query_author` class, and this will throw a nil exception if we don't catch it like this (and instead just return an empty array, which is what we want)


peetucket:
rubocop is currently targeting ruby 2.2 and fails on this...bumping to target ruby 2.5 produces a slew of other complaints

thoughts?  autogen ?

peetucket:
Added a test for this.  Targeted rubocop to ruby 2.5 to allow this new syntax through, and then created a new rubocop todo to get past the new/updated ruby cops.

PR #1122, opened 2019-10-30
add new deploy targets for Profiles team to test against

peetucket:
Based on the new VMs created here: https://github.com/sul-dlss/operations-tasks/issues/2061

Will be used by the Profiles team for testing their Oracle upgrade against while we maintain the current stack of VMs.  Later we can pull down these two instances/targets and use these new ones instead: `sul-pub-cap-dev` and `sul-pub-cap-qa`


peetucket:
Once testing is complete, we can remove the older (current) targets of `sul-pub-cap-dev` and `sul-pub-cap-qa`

Issue #1150, opened 2020-03-19
Improve error messaging reported to Honeybadger


peetucket:
This exception reported an interesting and useful error from our code, but the actual message was buried in the "context" and easy to miss.  Can we put this in a more noticeable message to honeybadger? 

HB error: https://app.honeybadger.io/projects/50046/faults/49773157

The place in the code reporting it: https://github.com/sul-dlss/sul_pub/blob/master/lib/notification_manager.rb#L28-L30

jmartin-sul:
a bit more from the honeybadger alert, in case it takes us longer than our HB account's retention period to look into this issue:


<img width="1137" alt="Screen Shot 2020-03-19 at 6 37 09 PM" src="https://user-images.githubusercontent.com/7741604/77129341-eb6c8480-6a10-11ea-89ed-ab3812cc8d9e.png">
<img width="1153" alt="Screen Shot 2020-03-19 at 6 37 26 PM" src="https://user-images.githubusercontent.com/7741604/77129342-ee677500-6a10-11ea-864f-142832556e2d.png">
<img width="1137" alt="Screen Shot 2020-03-19 at 6 37 55 PM" src="https://user-images.githubusercontent.com/7741604/77129343-ef000b80-6a10-11ea-910f-83a7d608259d.png">

peetucket:
closed by #1153

PR #1153, opened 2020-03-23
Fix issue with first names; add better exception for HB

### LINKED ISSUE #1149: Authors with a period for a first name cause a blank name query clause to be generated
This author currently has a period for a first name, causing a blank name clause `AU` to be generated.  We should at least put their last name (which does exist) in the name clause to avoid creating too many publication results or even better, skip the query entirely if it would be nothing.

```
author=Author.find(53490)
puts "CAP_PROFILE_ID: #{author.cap_profile_id}";puts "NUM PUBS: #{author.contributions.count}";puts "PRIMARY AUTHOR NAME: #{author.cap_first_name} #{author.cap_last_name}";author_query = WebOfScience::QueryAuthor.new(author);puts "ALL NAMES: #{author_query.send(:names).join(";")}";puts "INSTITUTIONS: #{author_query.send(:institutions).join(";")}";puts "WOS: #{author_query.send(:author_query)[:queryParameters][:userQuery]}";pm_query = Pubmed::QueryAuthor.new(author, {});puts "Pubmed: #{pm_query.send(:term)}";nil
CAP_PROFILE_ID: 69115
NUM PUBS: 1
PRIMARY AUTHOR NAME: . Rituraj
ALL NAMES:
INSTITUTIONS: stanford
WOS: AU=() AND AD=("stanford")
Pubmed: () AND (Stanford University[Affiliation])
=> nil
```

peetucket:
## Why was this change made?

fixes #1149 and #1150

Basically we want to guard against invalid queries when harvesting so we don't create too many publications.

This PR also adds a guard against too many publications for WoS harvesting the same way we currently guard against too many publications for pubmed harvesting (just in case we are missing other edge cases that would generate too many publications).

Basically, look at the name query for any given authors, and if's blank call it invalid, and then skip over the author when harvesting.

It also adds a new exception class for harvesting errors so its easier to see what the actual exception is in HB (previously an error from harvesting was reported as "StandardError" and you had to dig down into the "context" to see what actually was being reported, which is easy to miss).

## Does this change affect how this application integrates with other services?
If so, please confirm change was tested on stage and/or test added to sul-dlss/infrastructure-integration-test.

### DIFF
--- a/spec/factories/author.rb
+++ b/spec/factories/author.rb
@@ -37,10 +37,21 @@ FactoryBot.define do
     preferred_last_name { 'Edler' }
     preferred_middle_name { 'Jim' }
     emails_for_harvest { 'alice.edler@stanford.edu' }
-  end
 
-  factory :inactive_author, parent: :author do
-    active_in_cap { false }
+    trait :blank_first_name do
+      official_first_name { '' }
+      preferred_first_name { '' }
+    end
+
+    trait :space_first_name do
+      official_first_name { ' ' }
+      preferred_first_name { ' ' }
+    end
+
+    trait :period_first_name do
+      official_first_name { '.' }
+      preferred_first_name { '.' }
+    end
   end
 
   factory :author_with_alternate_identities, parent: :author do
@@ -52,32 +63,6 @@ FactoryBot.define do
         create(:author_identity, author: author)
       end
     end
-  end
-
-  factory :period_first_name_author, parent: :author do
-    active_in_cap { true }
-    cap_import_enabled { true }
-    official_first_name { '.' }
-    official_last_name { 'Dude' }
-    official_middle_name { '' }
-    preferred_first_name { '.' }
-    preferred_last_name { 'Dude' }
-    preferred_middle_name { '' }
-    email { 'bad.dude@stanford.edu' }
-    emails_for_harvest { 'bad.dude@stanford.edu' }
-  end
-
-  factory :blank_first_name_author, parent: :author do
-    active_in_cap { true }
-    cap_import_enabled { true }
-    official_first_name { '' }
-    official_last_name { 'Dude' }
-    official_middle_name { '' }
-    preferred_first_name { '' }
-    preferred_last_name { 'Dude' }
-    preferred_middle_name { '' }
-    email { 'dude@stanford.edu' }
-    emails_for_harvest { 'dude@stanford.edu' }
   end
 
   # Public data from
--- a/spec/factories/author.rb
+++ b/spec/factories/author.rb
@@ -54,6 +54,32 @@ FactoryBot.define do
     end
   end
 
+  factory :period_first_name_author, parent: :author do
+    active_in_cap { true }
+    cap_import_enabled { true }
+    official_first_name { '.' }
+    official_last_name { 'Dude' }
+    official_middle_name { '' }
+    preferred_first_name { '.' }
+    preferred_last_name { 'Dude' }
+    preferred_middle_name { '' }
+    email { 'bad.dude@stanford.edu' }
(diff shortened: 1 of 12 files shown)


spec/factories/author.rb
```diff
@@ -54,6 +54,32 @@
     end
   end
 
+  factory :period_first_name_author, parent: :author do
```
jcoyne: Have you considered using traits for this?

spec/lib/pubmed/harvester_spec.rb
```diff
@@ -21,6 +22,7 @@
       let(:pmid) { '30833575' }
       before do
         allow(Pubmed::QueryAuthor).to receive_message_chain(:new, :pmids).and_return([pmid])
```
jcoyne: How about returning a stub QueryAuthor from new rather than several stub chains?

spec/lib/pubmed/harvester_spec.rb
```diff
@@ -33,8 +35,26 @@
 
     context 'when author query has too many publications' do
       let(:lotsa_pmids) { Array(1..Settings.PUBMED.max_publications_per_author) }
+      let(:query_author) { instance_double(Pubmed::QueryAuthor) }
+
+      before do
+        allow(Pubmed::QueryAuthor).to receive(:new).with(author, {}).and_return(query_author)
+        allow(query_author).to receive(:pmids).and_return(lotsa_pmids)
```
jcoyne: You can incorporate these lines into the instance_double call

spec/lib/pubmed/harvester_spec.rb
```diff
@@ -33,8 +35,26 @@
...
+      end
+    end
+
+    context 'when author query is invalid' do
+      let(:query_author) { instance_double(Pubmed::QueryAuthor) }
+
       before do
-        allow(Pubmed::QueryAuthor).to receive_message_chain(:new, :pmids).and_return(lotsa_pmids)
+        allow(Pubmed::QueryAuthor).to receive(:new).with(author, {}).and_return(query_author)
+        allow(query_author).to receive('valid?').and_return(false)
```
jcoyne: Roll this into the instance double

spec/lib/web_of_science/harvester_spec.rb
```diff
@@ -136,4 +136,34 @@
...
     end
   end
+
+  context 'when author query has too many publications' do
+    let(:lotsa_uids) { Array(1..Settings.WOS.max_publications_per_author) }
+    let(:query_author) { instance_double(WebOfScience::QueryAuthor) }
+
+    before do
+      allow(WebOfScience::QueryAuthor).to receive(:new).with(author, {}).and_return(query_author)
+      allow(query_author).to receive(:uids).and_return(lotsa_uids)
```
jcoyne: These lines can roll into the instance_double

spec/factories/author.rb
peetucket: just learned about this, adjusted to use

spec/lib/pubmed/harvester_spec.rb
peetucket: adjusted

spec/lib/pubmed/harvester_spec.rb
peetucket: done

spec/lib/web_of_science/harvester_spec.rb
peetucket: done

spec/lib/pubmed/harvester_spec.rb
peetucket: done

PR #1185, opened 2020-08-05
check for manual publication before updating instead of ids

### LINKED ISSUE #1184: Manually updating a publication with an identifier fails
We should use the provenance to decide if a publication can be manually updated or not instead of the presence of identifiers..

Report: a user trying to update a manual publication gets this error:

```
"error":"This record may not be modified.  If you had originally entered details for the record, it has been su
perceded by a central record."
```

i was able to reproduce with this publication: "Meeting U.S. Solid Oxide Fuel Cell Targets Joule  Whiston, M. M. 2019"   which is SulpubID `756842` and cap_profile_id `205881`   it has a wos_uid, even though it is provenance `cap`, and it should allowed to be updated

peetucket:
## Why was this change made?

Fixes #1184 

Some users cannot update their manual publications because our guard cause was looking at identifiers in the record instead of the provenance.  We later allowed model record level identifiers to be added to a publication, which was confusing our logic about which publications could be updated (and refusing to allow a legit manual publication to be updated).

Also discovered that the tests related to updating manual publications were not working as we expected.  We thought we were using a POST to generate a publication before updating it, but this was in fact not actually happening, and that POST was returning a 406, but our tests just happened to be crafted in a way that it worked anyway.  I changed the tests to use a factory to setup the test data instead of a POST within the test itself, and then confirmed the data was changed as expected in the test.  I also added a new test for this change.

Added a new test, and refactored existing related tests.

None.

### DIFF
--- a/app/controllers/publications_controller.rb
+++ b/app/controllers/publications_controller.rb
@@ -101,8 +101,8 @@ class PublicationsController < ApplicationController
       head :gone
       return
     end
-    if old_pub.sciencewire_id.present? || old_pub.pmid.present? || old_pub.wos_uid.present?
-      render json: { "error": 'This record may not be modified.  If you had originally entered details for the record, it has been superceded by a central record.' }, status: :forbidden, format: 'json'
+    if old_pub.harvested_pub? # only manually entered (i.e. non-harvested) publications may be updated with this method
+      render json: { "error": "This record SulPubID #{old_pub.id} may not be modified.  If you had originally entered details for the record, it has been superceded by a central record." }, status: :forbidden, format: 'json'
       return
     elsif !validate_or_create_authors(new_pub[:authorship])
       render json: { "error": 'You have not supplied a valid authorship record.' }, status: :not_acceptable, format: 'json'
--- a/app/models/publication.rb
+++ b/app/models/publication.rb
@@ -230,15 +230,19 @@ class Publication < ActiveRecord::Base
   end
 
   def sciencewire_pub?
-    provenance == 'sciencewire'
+    provenance == Settings.sciencewire_source
   end
 
   def pubmed_pub?
-    provenance == 'pubmed'
+    provenance == Settings.pubmed_source
   end
 
   def wos_pub?
-    provenance == 'wos'
+    provenance == Settings.wos_source
+  end
+
+  def harvested_pub?
+    provenance != Settings.cap_provenance
   end
 
   def authoritative_pmid_source?
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -108,7 +108,6 @@ sul_doc_types:
 
 batch_source: batch
 cap_provenance: cap
-manual_source: manual
 pubmed_source: pubmed
 sciencewire_source: sciencewire
 wos_source: wos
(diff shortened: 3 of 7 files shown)


app/models/publication.rb
```diff
@@ -230,15 +230,19 @@ def year
...
   end
 
   def wos_pub?
-    provenance == 'wos'
+    provenance == Settings.wos_source
+  end
+
+  def cap_pub?
+    provenance == Settings.cap_provenance
   end
```
peetucket, line 246: a new helper method for testing if this a manual publication by looking at the provenance

app/models/publication.rb
```diff
@@ -230,15 +230,19 @@ def year
...
 
   def pubmed_pub?
-    provenance == 'pubmed'
+    provenance == Settings.pubmed_source
   end
 
   def wos_pub?
-    provenance == 'wos'
+    provenance == Settings.wos_source
```
peetucket, line 241: we have these strings in the settings already, let's use them instead of duplicating

app/controllers/publications_controller.rb
```diff
@@ -101,8 +101,8 @@ def update
       head :gone
       return
     end
-    if old_pub.sciencewire_id.present? || old_pub.pmid.present? || old_pub.wos_uid.present?
-      render json: { "error": 'This record may not be modified.  If you had originally entered details for the record, it has been superceded by a central record.' }, status: :forbidden, format: 'json'
+    if !old_pub.cap_pub? # only manually entered publications may be updated like this
```
ndushay, line 104: NOT?

mjgiarlo: PR LGTM but it'd be good to have this question answered before we merge. There seems to be an important notion here, that of whether a publication is "manual" or "manually entered" but this is a case where the language of the code does not match the language of the domain, so a reviewer without intimate sul_pub knowledge can not tell if this line is doing EXACTLY what you want or EXACTLY what you want to avoid. Of course, I trust you as the code author, but I also know that we all make silly typos and mistakes like getting boolean conditions the exact opposite, so ... not merging for now, but leaving this long-winded comment which someone should put on a T-shirt sometime.

app/models/publication.rb
mjgiarlo: so if `cap_pub?` is `true`, it's a manual publication? or if it's `false`, it's a manual publication? would you consider naming the method so that this is clearer? I'm thinking about the inebriation test.

peetucket: yeah, cap_pub? == true is a manual publication (basically just matching the string in the provenance) .... since its only used in one place in our new caluse, I can name the method something like "harvested_pub?", which will change the guard clause to a true instead of false

peetucket:
Adjusted helper method to make the guard clause make more sense when read

config/settings.yml
```diff
@@ -108,7 +108,6 @@ sul_doc_types:
 
 batch_source: batch
 cap_provenance: cap
-manual_source: manual
```
peetucket, line 111: could not find this setting referenced anywhere in the code

peetucket:
@ndushay do you want to have another look before I merge?

PR #1192, opened 2020-09-10
disable pubmed lookups via config

peetucket:
## Why was this change made?

Created in response to #1193 

We already have a configuration setting to disable pubmed harvesting, which we used when that was causing us problems with our search logic.  However, if the Pubmed API goes down completely, it would be useful to also explicitly disable any pubmed lookups during that time to avoid exceptions in our code.  These other lookups include:

- a user searching for a publication manually via a PMID
- augmentation of WoS harvested records with Pubmed data

Updated/added tests


spec/lib/pubmed/fetcher_spec.rb
```diff
@@ -95,8 +96,11 @@
...
 
-    context 'WOS enabled' do
-      before { allow(Settings.WOS).to receive(:enabled).and_return(true) }
+    context 'Pubmed and WOS both enabled' do
+      before do
+        allow(Settings.WOS).to receive(:enabled).and_return(true)
+        allow(Settings.PUBMED).to receive(:lookup_enabled).and_return(true)
+      end
 
       it 'searches only WOS' do
```
ndushay, line 105: if they are both enabled, shouldn't it search both?

ndushay (changes requested):
this looks good except for a question on one spec.

spec/lib/pubmed/fetcher_spec.rb
peetucket: its a bit confusing, but not in this case

The reason is that

1. You can search for publications by PMID in both WoS and Pubmed Central.
2. WoS is currently the preferred service when you search by PMID (something that is actually changed by one of the draft PRs, but would require verification by the Profiles team)
3. This test has a result coming from WoS for the PMID being searched.

So in the case when both are enabled, you'll get a result from WoS, then stop and never hit PMID.

Really this scenairo would have another test that leaves them both enabled, but simulate returning nothing from WoS.  Then we *would* expect the Pubmed client to run.

peetucket: lemme take a stab at writing such a test

peetucket: and will clarify what is happening in the test too with a better test name

peetucket: actually the very next test does test this scenario of no results found at WoS and verifies it hits Pubmed.  let me word for better clarity though

ndushay (approved):

Issue #831, opened 2018-04-18
Duplicate keys errors in web of science source records


peetucket:
Errors like this keep showing up in the logs during harvesting.  We should understand why to see if any mitigation is necessary:

```
E, [2018-04-18T05:07:47.834058 #21220] ERROR -- : WebOfScience::ProcessRecords: Author: 92207, ProcessRecords failed
E, [2018-04-18T05:07:47.834111 #21220] ERROR -- : #<ActiveRecord::RecordNotUnique: Mysql2::Error: Duplicate entry '351586' for key 'index_web_of_science_source_records_on_publication_id': UPDATE `web_of_science_source_records` SET `publication_id` = 351586, `updated_at` = '2018-04-18 12:07:46' WHERE `web_of_science_source_records`.`id` = 9509>
E, [2018-04-18T05:07:47.834172 #21220] ERROR -- : /opt/app/pub/sul-pub/shared/bundle/ruby/2.3.0/gems/mysql2-0.4.10/lib/mysql2/client.rb:120:in `_query'
/opt/app/pub/sul-pub/shared/bundle/ruby/2.3.0/gems/mysql2-0.4.10/lib/mysql2/client.rb:120:in `block in query'
/opt/app/pub/sul-pub/shared/bundle/ruby/2.3.0/gems/mysql2-0.4.10/lib/mysql2/client.rb:119:in `handle_interrupt'
/opt/app/pub/sul-pub/shared/bundle/ruby/2.3.0/gems/mysql2-0.4.10/lib/mysql2/client.rb:119:in `query'
/opt/app/pub/sul-pub/shared/bundle/ruby/2.3.0/gems/activerecord-4.2.10/lib/active_record/connection_adapters/abstract_mysql_adapter.rb:317:in `block in execute'
/opt/app/pub/sul-pub/shared/bundle/ruby/2.3.0/gems/activerecord-4.2.10/lib/active_record/connection_adapters/abstract_adapter.rb:484:in `block in log'
/opt/app/pub/sul-pub/shared/bundle/ruby/2.3.0/gems/activesupport-4.2.10/lib/active_support/notifications/instrumenter.rb:20:in `instrument'
/opt/app/pub/sul-pub/shared/bundle/ruby/2.3.0/gems/activerecord-4.2.10/lib/active_record/connection_adapters/abstract_adapter.rb:478:in `log'
[...]

PR #1199, opened 2020-09-22
do a better job of finding existing matching publication records when adding then manually

### LINKED ISSUE #1198: Incorrect error message shown to user when trying to add a new publication manually
I have a ticket from Mohammad Shahrokh Esfahani (profile id = 63608) where he reports that he has issues trying to add publications to his profile.  In my investigation of the issue, I found that it is only certain publications that cannot be added. 
 
He provided me with the following two examples.  I did confirm his findings using his production profile and mine as well.  I found the same issue trying to add the publication to his profile as I did to mind for these publications:
 
·        Dynamic Risk Profiling Using Serial Tumor Biomarkers for Personalized Outcome Prediction CELL
 
·        Circulating Tumor DNA Analysis for Detection of Minimal Residual Disease After Chemoradiotherapy for Localized Esophageal Cancer GASTROENTEROLOGY
 
 
Please see the attached screenshots.  The message received when selecting the ‘+add’ option is:
 
“Unable to update this publication right now.  Please try again later.”
 
For other publications for Mohammad, the add is successful, as you can see on his profile and from the following example of:
 
·        Functional significance of U2AF1 S34F mutations in lung adenocarcinomas Nature Communications
 
Please see the attached screenshot that shows the successful add for this publication.
 
Another issue which I noticed for his profile is that he has, according to the number shown, 1548 new publications in his Profile Inbox.  The review page only shows 100 of these, but of those, I am not finding Esfahani listed as an author.  Not sure if this is related to the above issue, but this is a problem.

peetucket:
## Why was this change made?

Fixes #1198

Test suite and manually on sul-cap-dev

None

### DIFF
--- a/app/controllers/authorships_controller.rb
+++ b/app/controllers/authorships_controller.rb
@@ -221,7 +221,8 @@ class AuthorshipsController < ApplicationController
   # @return [Publication]
   def get_publication_via_wos!(author, wos_uid)
     WebOfScience.harvester.author_uid(author, wos_uid)
-    pub = Publication.find_by(wos_uid: wos_uid)
+    # we find the publication by looking for any matching publication for this WOS record (could be MEDLINE or WOS ID)
+    pub = WebOfScienceSourceRecord.find_by(uid: wos_uid)&.record&.matching_publication
     unless pub
       log_and_error!("The publication with WOS_UID:#{wos_uid} was not found either locally or at WebOfScience.")
       false
--- a/app/controllers/authorships_controller.rb
+++ b/app/controllers/authorships_controller.rb
@@ -229,12 +229,10 @@ class AuthorshipsController < ApplicationController
 
     # we find the publication by looking for any matching publication for this WOS source record (could be MEDLINE or WOS ID, use all valid identifiers)
     pub = wossr.record.matching_publication
+    return pub if pub
 
-    unless pub
-      log_and_error!("A matching publication record for WOS_UID:#{wos_uid} was not found in the publication table.")
-      false
-    end
-    pub
+    log_and_error!("A matching publication record for WOS_UID:#{wos_uid} was not found in the publication table.")
+    false
   end
 
 end
--- a/app/controllers/authorships_controller.rb
+++ b/app/controllers/authorships_controller.rb
@@ -221,10 +221,17 @@ class AuthorshipsController < ApplicationController
   # @return [Publication]
   def get_publication_via_wos!(author, wos_uid)
     WebOfScience.harvester.author_uid(author, wos_uid)
-    # we find the publication by looking for any matching publication for this WOS record (could be MEDLINE or WOS ID)
-    pub = WebOfScienceSourceRecord.find_by(uid: wos_uid)&.record&.matching_publication
+    wossr = WebOfScienceSourceRecord.find_by(uid: wos_uid)
+    unless wossr
+      log_and_error!("A WebOfScienceSourceRecord was not found for WOS_UID:#{wos_uid}.")
+      return false
+    end
+
+    # we find the publication by looking for any matching publication for this WOS source record (could be MEDLINE or WOS ID, use all valid identifiers)
+    pub = wossr.record.matching_publication
+
     unless pub
-      log_and_error!("The publication with WOS_UID:#{wos_uid} was not found either locally or at WebOfScience.")
+      log_and_error!("A matching publication record for WOS_UID:#{wos_uid} was not found in the publication table.")
       false
     end
     pub
--- a/lib/web_of_science/record.rb
+++ b/lib/web_of_science/record.rb
@@ -128,7 +128,6 @@ module WebOfScience
     # Does record have an associated publication? (based on matching PublicationIdentifiers)
     # Note: must use unique identifiers, don't use ISSN or similar series level identifiers
     # We search for all PubIDs at once instead of serial queries.  No need to hit the same table multiple times.
-    # @param [WebOfScience::Record] record
     # @return [::Publication, nil] a matched or newly minted publication
     def matching_publication
       Publication.joins(:publication_identifiers).where(
--- a/lib/web_of_science/record.rb
+++ b/lib/web_of_science/record.rb
@@ -124,5 +124,29 @@ module WebOfScience
         rec.database = database
       end
     end
+
+    # Does record have a contribution for this author? (based on matching PublicationIdentifiers)
+    # Note: must use unique identifiers, don't use ISSN or similar series level identifiers
+    # We search for all PubIDs at once instead of serial queries.  No need to hit the same table multiple times.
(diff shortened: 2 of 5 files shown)


app/controllers/authorships_controller.rb
```diff
@@ -221,7 +221,8 @@ def get_publication_via_sciencewire!(sw_id)
   # @return [Publication]
   def get_publication_via_wos!(author, wos_uid)
     WebOfScience.harvester.author_uid(author, wos_uid)
-    pub = Publication.find_by(wos_uid: wos_uid)
+    # we find the publication by looking for any matching publication for this WOS record (could be MEDLINE or WOS ID)
+    pub = WebOfScienceSourceRecord.find_by(uid: wos_uid)&.record&.matching_publication
```
jcoyne, line 225: This set of conditionals is hard to understand.  Can you make the logic do something like this, so we better understand where the nils might come from?
```ruby
wossr = WebOfScienceSourceRecord.find_by(uid: wos_uid)
if wossr
  pub = wossr.record.matching_publication
...
```

jcoyne (changes requested):

lib/web_of_science/record.rb
```diff
@@ -124,5 +124,29 @@ def find_or_create_model
         rec.database = database
       end
     end
+
+    # Does record have a contribution for this author? (based on matching PublicationIdentifiers)
+    # Note: must use unique identifiers, don't use ISSN or similar series level identifiers
+    # We search for all PubIDs at once instead of serial queries.  No need to hit the same table multiple times.
+    # @param [WebOfScience::Record] record
+    # @return [::Publication, nil] a matched or newly minted Contribution
+    def matching_publication
```
jcoyne, line 133: 👏

peetucket:
Overiew:

We run an a harvest at some point and get this pub, it is in the MEDLINE collection at the Web of Science:

```
MEDLINE:123456
DOI:1234
```

This generated a new Publication record for an author, along with a new Web of Science source record with MEDLINE:123456, and associates the publication with the source record in the 'wos_uid'.

Later a different author comes and searches for that publication by Title: 'Title of the pub'

We find it, but now it is also in the Web of Science Core Collection, and thus returns a different WOS identifier, but the same DOI:

```
WOS:67890
DOI:1234
```

The new author clicks the add button, and we try to add this publication to their profile with the new identifier: WOS:67890

Our code is smart and via the `matching_publication` method, it realizes it is the same as the MEDLINE record (the DOIs are the same), so instead of making a new publication, it associates the existing MEDLINE:123456 generated publication with the new author.  This does not generate a new publication record because it doesn't need to.  It does happen generate a new Web of Science Source Record for WOS:67890 because we happen to store these records before doing the associations, but that doesn't matter, its just extra data.  I don't want to make too many changes, so I am not going to worry about this extra source record.  This is an also an edge case that doesn't happen often.

Anyway, the original Publication has 'MEDLINE:123456' in the 'wos_uid' column of the publication table (this was the original publication record we created from the harvest).  And we did not go back and change the 'wos_uid' column to 'WOS:67890', the new publication record we are trying to add, because why would we?  The publication record was built from the original MEDLINE record and we just associated that the author, determining it was actually the same (via the DOI matching).  In fact, we used to change the 'wos_uid' column to the new record before and stopped because it was causing problems.  See https://github.com/sul-dlss/sul_pub/issues/831 and https://github.com/sul-dlss/sul_pub/commit/d0b160b8d5a0032802e63daede972c5653e43662

SO ... so far, this all seems fine.  The original publication was added to the new author record, and we didn't mess up any existing data.  So hhy is there a problem?

It's because right before we return to the user, we verify the new publication is associated by doing this:

`Publication.find_by(wos_uid: 'WOS:67890')`

here: https://github.com/sul-dlss/sul_pub/blob/master/app/controllers/authorships_controller.rb#L224

We basically are asking to find the publication by the new WOS identifier (not the original MEDLINE one).  We just said that we don't need to do this.  This returns NIL, and the user gets a message saying the publication can't be found.

I propose that instead we use the more intelligent logic from our `matching_publication` method, which will discover it (as it did before) by virtue of the fact that it has the same DOI.  I had to move this method to a different class in order to use it, but that move makes sense anyway.

**Note**:

Publication == a database row with parsed data from any number of original source records (depending on which API it came from) that will be linked to a user's profile via the join table

SourceRecord (e.g. WebOfScienceSourceRecord, PubmedSourceRecord, SciencewireSourceRecord) == cached raw XML that came back from an API call and may have been used to build a publication record

app/controllers/authorships_controller.rb
peetucket: Yeah, ok.  I don't think we needed to check for the second nil (a record should return a matching publication), but I'll break things out with different error messages based on what happened.

lib/web_of_science/record.rb
```diff
@@ -124,5 +124,29 @@ def find_or_create_model
         rec.database = database
       end
     end
+
+    # Does record have an associated publication? (based on matching PublicationIdentifiers)
+    # Note: must use unique identifiers, don't use ISSN or similar series level identifiers
+    # We search for all PubIDs at once instead of serial queries.  No need to hit the same table multiple times.
+    # @param [WebOfScience::Record] record
```
jmartin-sul, line 131:
```suggestion
```

i think this `param` comment can go now?

jmartin-sul (commented):
one small suggestion about what seemed like parameter comment cruft, but otherwise LGTM

lib/web_of_science/record.rb
peetucket: yes, good catch -- fixed

app/controllers/authorships_controller.rb
```diff
@@ -221,9 +221,17 @@ def get_publication_via_sciencewire!(sw_id)
...
+      log_and_error!("A WebOfScienceSourceRecord was not found for WOS_UID:#{wos_uid}.")
+      return false
+    end
+
+    # we find the publication by looking for any matching publication for this WOS source record (could be MEDLINE or WOS ID, use all valid identifiers)
+    pub = wossr.record.matching_publication
+
     unless pub
-      log_and_error!("The publication with WOS_UID:#{wos_uid} was not found either locally or at WebOfScience.")
+      log_and_error!("A matching publication record for WOS_UID:#{wos_uid} was not found in the publication table.")
```
jcoyne, line 234: Can this case get a test?
I how about doing `return pub if pub`. I'm not sure why https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/GuardClause isn't being enforced here.

jcoyne (changes requested):

app/controllers/authorships_controller.rb
peetucket: adjusted style and added a test case for the pub can't be found (and adjusted the other test case to catch the other specific error)

jcoyne (approved):

PR #1204, opened 2020-10-15
add new custom report

### LINKED ISSUE #1201: Custom regular report for Stanford Medicine Center for Improvement (SMCI)
Consider producing a custom report on a regular basis that includes both publications for users currently in profiles, as well as users not currently in profiles.

Most likely it will be a once a month report.  They need to have a citation or be able to build the citation they want.  Therefore, we were going to provide them with the most recent approved publications (from the last date received) with the following:

- First Published Date
- DOI URL
- Chicago Citation Format
- APA Citation Format
- MLA Citation Format

Additionally, we wanted to provide them with the separated fields that form the citation in the case they want to do their own formatting for the citation display on their website.  So that would be, for example:

- Title
- Article No.
- Pages
- DOI / DOI URL
- Journal Title
- Publisher
- Year Published
- Volume
- Issue
- Supplement Number
- Authors - First name, Middle Initial, Last Name

This would also need to include the author's details, when available:

- First Name
- Last Name
- Position
- sunet id
- profile id
- univ id
- email address

The thought is that possibly it would make sense, if you can pull the information for those without profiles, that you produce the full report for everyone that also includes the data for those with profiles (thus some of the fields noted above), so that the report is coming from one source and combined.  The number of researchers is from 94 - 130 from the current information we have.  Out of these there were around 33 that did not have profiles or active profiles and there a number that do not have full profiles that have the publication import option.  Also, we have staff members in the list who would also not have the publication import turned on by default, as this is only on by default for faculty and postdocs.

peetucket:
## Why was this change made?

Fixes #1201 

A specific custom report that combines publications from our database with WoS name queries for people not in our database.

Running the report on my laptop with various iterations of data.  New unit test to verify basic runnability.

None yet, but will add to the wiki.

### DIFF
--- a/lib/smci_report.rb
+++ b/lib/smci_report.rb
@@ -1,0 +1,220 @@
+# Exports all publications for the given sunets or cap_profile_ids in an 'approved' state
+#  and it also harvests publications by name for authors who do not have a sunet or cap_profile_id
+# This is a custom report for Stanford Medicine Center for Improvement (SMCI)
+#   specified in https://github.com/sul-dlss/sul_pub/issues/1201
+
+# October 2020
+
+# You must have an input csv file with the following columns (including a header row):
+#   sunetid,cap_profile_id,first_name,middle_name,last_name,institutions
+#
+#  For authors that exist in profiles, you can specify either a sunet or cap_profile_id (sunet preferred if both specified)
+#  For authors that do NOT exist in profiles, you MUST leave the cap_profile_id and sunet field columns blank, and then you must specify first name and last name
+#    and optionally middle name a comma delimited list of institutions (defaults to 'stanford' if no institutions provided)
+
+# You can run it on the Rails console with:
+#
+# smci = SMCIReport.new(input_file: 'tmp/input.csv', output_file: 'tmp/output_file.csv', date_since: '1/1/2000', time_span: '1year')
+# smci.run
+
+# You can also run as a rake task to avoid having to drop into the rails console:
+
+#
+# bundle exec rake sul:smci_export['tmp/authors.csv','tmp/results.csv','1/1/2000','10year']
+# bundle exec rake sul:smci_export['tmp/authors.csv','tmp/results.csv',,] # for all time
+
+# Note that in either case, the runtime may take a while (like 30 mins or longer) if you have a reasonable large (>100) number of authors
+#  so I'd suggest running in screen or nohup.
+
+# The 'input_file' parameter indicates where your input CSV file with the desired authors is.
+# The 'output_file' parameter species where the output CSV will be generated.
+
+# The 'date_since' parameter is optional and specifies the date after which publications for profiles authors will be exported
+#  The format for 'date_since'  is DD/MM/YYYY.
+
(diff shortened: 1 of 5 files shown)


lib/smci_report.rb
```diff
@@ -0,0 +1,205 @@
...
+#   specified in https://github.com/sul-dlss/sul_pub/issues/1201
+
+# October 2020
+
+# You must have an input csv file with the following columns (including a header row):
+#   sunetid,cap_profile_id,first_name,middle_name,last_name,institutions
+#
+#  For authors that exist in profiles, you can specify either a sunet or cap_profile_id (sunet preferred if both specified)
+#  For authors that do NOT exist in profiles, you MUST leave the cap_profile_id and sunet field columns blank, and then you must specify first name and last name
+#    and optionally middle name a comma delimtied list of institutions (defaults to 'stanford' if no institutions provided)
```
justinlittman, line 13: typo = delimtied

lib/smci_report.rb
```diff
@@ -0,0 +1,205 @@
...
+
+# Run with:
+#
+# smci = SMCIReport.new(input_file: 'tmp/input.csv', output_file: 'tmp/output_file.csv', date_since: '1/1/2000', time_span: '1year')
```
justinlittman, line 17: What is the context for executing these commands?

justinlittman: Also, mention that this is available as a rake task.

justinlittman (commented):
This seems like a reasonable, minimal effort approach.

Add documentation to README?

lib/smci_report.rb
peetucket: thanks, will fix

lib/smci_report.rb
peetucket: right, this was before there was a rake task...but I will indicate you can run this from the Rails console as well if you want

peetucket:
> This seems like a reasonable, minimal effort approach.
> 
> Add documentation to README?

Will add to the wiki, which is the place where this sort of custom scripty stuff is documented for this project.

lib/smci_report.rb
```diff
@@ -0,0 +1,220 @@
...
+    mesh = pub_hash[:mesh_headings] ? pub_hash[:mesh_headings].map { |h| h[:descriptor][0][:name] }.compact.reject(&:empty?).join('; ') : ''
+    if author # if this is a profiles author, then we will output this info as well
+      author_last_name = author.last_name
+      author_first_name = author.first_name
+      author_sunet = author.sunetid
+      author_cap_profile_id = author.cap_profile_id
+      author_univ_id = author.university_id
+      author_email = author.email
+    else
+      author_last_name = author_first_name = author_sunet = author_cap_profile_id = author_univ_id = author_email = ""
```
jcoyne, line 187: Can you add a test for this branch?

jcoyne (changes requested):

lib/smci_report.rb
peetucket: added a few more tests to hit this branch of the code

jcoyne (approved):

PR #1229, opened 2021-02-12
include new publications in smci report; show status in reports

peetucket:
## Why was this change made?

Based on continuing work on supporting campus research intelligence requests, the following changes to export reports are useful:

- Update two different reports that export publications to CSV so that they include the publication/contribution status (e.g. 'new' or 'approved').
- Update the SMCI report to include 'new' publications as well.


lib/smci_report.rb
```diff
@@ -109,13 +109,13 @@ def run
 
           if author # we found the author in our database, now get their publications
             contributions = Contribution.select('*')
-            contributions = contributions.where("author_id = ? and status in ('approved')", author.id)
+            contributions = contributions.where("author_id = ? and status in ('new','approved')", author.id)
```
jcoyne, line 112: can you do `contributions.where(author: author).where(status: %w[new approved])`?

lib/smci_report.rb
```diff
@@ -182,10 +182,10 @@ def run
   private
 
   def header_row
-    %w(title author_list pmid wos_uid sul_pub_id doi doi_url publisher journal volume articlenumber supplement issue pages mesh pub_year pub_date provenance profiles_author_last_name profiles_author_first_name profiles_author_sunet profiles_author_cap_profile_id profiles_author_employee_id profiles_author_email pub_harvested_date apa_citation mla_citation chicago_citation)
+    %w(title author_list pmid wos_uid sul_pub_id doi doi_url publisher journal volume articlenumber supplement issue pages mesh pub_year pub_date provenance profiles_author_last_name profiles_author_first_name profiles_author_sunet profiles_author_cap_profile_id profiles_author_employee_id profiles_author_email publication_status pub_harvested_date apa_citation mla_citation chicago_citation)
   end
 
-  def output_row(pub_hash, author, harvested_at)
+  def output_row(pub_hash, author, harvested_at, publication_status)
```
jcoyne, line 188: Can you use kwargs for this new field? That makes it easier to understand what we're passing in.

jcoyne (changes requested):

lib/smci_report.rb
peetucket: Yup, good call

lib/smci_report.rb
peetucket: Yes, that makes it easier to understand.  I also made several defaults for one spot where it is called that made it cleaned too.

jcoyne (approved):

PR #1243, opened 2021-04-01
remove unneeded environments; setup cron jobs appropriately

peetucket:
## Why was this change made?

We can remove the environments we created to help the Profiles team with their Oracle upgrade.  All tasks shown in https://github.com/sul-dlss/operations-tasks/issues/2512  Ops is taking care of decommissioning the servers we no longer need (and for which the deploy targets are being removed).  

Namely:
- rename the `cap-dev-a` (new server) deploy target to `cap-dev` (note that while `cap-dev` server is being decommissioned, we will use DNS to point from `cap-dev` to `cap-dev-a` so we can continue to deploy/ssh into the same server name)
- remove `cap-uat` deploy target as it's no longer needed
- adjust the `qa` deploy target to point to the new server

HOLD until we get the green light from the profiles team.

We need to follow up this PR with documentation update tasks and shared_config changes shown in the Ops ticket: https://github.com/sul-dlss/operations-tasks/issues/2512


peetucket:
closed in favor of #1245

PR #1245, opened 2021-04-01
update deploy targets

peetucket:
## Why was this change made?

We can remove the environments we created to help the Profiles team with their Oracle upgrade.  All tasks shown in https://github.com/sul-dlss/operations-tasks/issues/2512  Ops is taking care of decommissioning the servers we no longer need (and for which the deploy targets are being removed).  

Namely:
- remove the `cap-dev-a` (new server) deploy target and update the existing `cap-dev` target to point to the new server (note that while `cap-dev` server is being decommissioned, we will use also DNS to point from `cap-dev` to `cap-dev-a` so that we could continue to deploy/ssh into the same server name and get it go to the new server, but let's just be explicit here)
- remove `cap-uat` deploy target as it's no longer needed
- adjust the `qa` deploy target to point to the new server

HOLD until we get the green light from the profiles team.

We need to follow up this PR with documentation update tasks and shared_config changes shown in the Ops ticket: https://github.com/sul-dlss/operations-tasks/issues/2512


config/deploy/cap-dev.rb
```diff
@@ -1,5 +1,5 @@
 # see https://github.com/sul-dlss/sul_pub/wiki/Servers-Deployment-environment
-server 'sul-pub-cap-dev.stanford.edu', user: 'pub', roles: %w(web db app)
+server 'sul-pub-cap-dev-a.stanford.edu', user: 'pub', roles: %w(web db app)
```
mjgiarlo, line 2: Old dev-a also had the harvester_dev role. Should this also?

peetucket: yes, I added that ... thanks

mjgiarlo (approved):

PR #1276, opened 2021-05-11
add authors.cap_visibility to schema, populate from retrieved CAP authorship data

### LINKED ISSUE #1251: Add visibility for authors
When retrieving author information from CAP, also store visibility.

This will require:
* Adding column for storing CAP visibility to author model
* Enhancing Author Poller to store CAP visibility

jmartin-sul:
## Why was this change made?

closes #1251 

- [x] unit tests
- [x] deploy to qa and run `rake cap:poll` with a large value for its one param, `days_ago` (e.g. 365?), to see if actual visibility data fits the validation constraints imposed by this code change.

n/a

### DIFF
--- a/app/models/author.rb
+++ b/app/models/author.rb
@@ -3,12 +3,19 @@
 require 'set'
 
 class Author < ApplicationRecord
+  # Allowed values for visibility
+  VISIBILITY_VALUES = %w[public private stanford].freeze
+
   has_paper_trail on: [:destroy]
   validates :cap_profile_id, uniqueness: true, presence: true
 
   has_many :author_identities, dependent: :destroy, autosave: true
   has_many :contributions, dependent: :destroy
   has_many :publications, through: :contributions
+
+  # nil values allowed because we have historical records without visibility info, for which cap API
+  # will no longer have updated author info (e.g. for authors who are no longer at Stanford)
+  validates :cap_visibility, inclusion: { in: VISIBILITY_VALUES }, allow_nil: true
 
   # Provide consistent API for Author and AuthorIdentity
   alias_attribute :first_name, :preferred_first_name
@@ -122,7 +129,8 @@ class Author < ApplicationRecord
     seed_hash = {
       cap_profile_id: auth_hash['profileId'],
       active_in_cap: auth_hash['active'],
-      cap_import_enabled: auth_hash['importEnabled']
+      cap_import_enabled: auth_hash['importEnabled'],
+      cap_visibility: auth_hash['visibility']
     }
     profile = auth_hash['profile']
     seed_hash[:sunetid] = profile['uid'] || ''
--- a/db/migrate/20210511164010_add_cap_visibility_to_authors.rb
+++ b/db/migrate/20210511164010_add_cap_visibility_to_authors.rb
@@ -1,0 +1,5 @@
+class AddCapVisibilityToAuthors < ActiveRecord::Migration[6.0]
+  def change
+    add_column :authors, :cap_visibility, :string
+  end
+end
--- a/db/migrate/20210512190856_add_index_to_authors_cap_visibility.rb
+++ b/db/migrate/20210512190856_add_index_to_authors_cap_visibility.rb
@@ -1,0 +1,5 @@
+class AddIndexToAuthorsCapVisibility < ActiveRecord::Migration[6.0]
+  def change
+    add_index :authors, :cap_visibility
+  end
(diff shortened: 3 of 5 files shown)


jmartin-sul:
two question:
1. should the new column be indexed?  will we be searching on it?
2. when i ran `rails db:migrate` locally, a bunch of text columns went from being defined as `limit: 16777215` to `size: :medium` in my `db/schema.rb`.  should i commit that change?  e.g.:
```diff
-    t.text "bibtex_source_data", limit: 16777215
+    t.text "bibtex_source_data", size: :medium
```

jmartin-sul:
in person discussion:
1. sure index the new column.  prob won't hurt (at least not substantially), we might be searching/filtering on it, and table is prob only about 40k rows.  but we're likely to do e.g. `visibility = 'public'` queries.
2. check in the change from `db:migrate`
3. 365 days should be a good enough window for the test run on QA.  none of the error alerts from during the run yesterday evening seem related to this change ([a network timeout](https://app.honeybadger.io/projects/50046/faults/79342517), [an error in a WoS query we built](https://app.honeybadger.io/projects/50046/faults/79342739), [another invalid WoS query](https://app.honeybadger.io/projects/50046/faults/79343760)).
4. one more sanity check: query the column in QA to see if we still have any (and if so how many) null values for the new visibility column.

jmartin-sul:
* pushed a change so that the new col is indexed (as a separate migration file to prevent deployment trouble, since the first migration was already deployed to QA).
* that push also incorporated the other `db:migrate` changes about text column defs
* lots of authors still have no visibility info in QA, even after syncing back 365 days:
```ruby
[24] pry(main)> Author.pluck('DISTINCT cap_visibility')
=> [nil, "private", "public", "stanford"]
[25] pry(main)> Author.where(cap_visibility: nil).count
=> 100404
[26] pry(main)> Author.count
=> 137776
[27] pry(main)> 
```

I'm going to make this "ready for review", but happy to put back in draft and revisit in light of the lack of visibility info for most authors as indicated by the query above.

peetucket:
Verify in #sul-cap-collab about why some visibility values are returning nil and why we get 404 for some:

```
cap_profile_id='24277'
=> "24277"
[2] pry(main)> response = Cap::Client.new.get_auth_profile(cap_profile_id)
** [Honeybadger] Reporting error id=1aa22490-1db4-41e9-ad5f-a229ff2e1a0e level=1 pid=20119
Faraday::ResourceNotFound: the server responded with status 404
from /opt/app/pub/sul-pub/shared/bundle/ruby/2.7.0/gems/faraday-1.4.1/lib/faraday/response/raise_error.rb:22:in `on_complete'
[3] pry(main)> ** [Honeybadger] Success ⚡ https://app.honeybadger.io/notice/1aa22490-1db4-41e9-ad5f-a229ff2e1a0e id=1aa22490-1db4-41e9-ad5f-a229ff2e1a0e code=201 level=1 pid=20119
```

jmartin-sul:
switching to draft till we figure out the blank visibility issue

jmartin-sul:
another question from today's post-standup discussion:  should the new `cap_visibility` field have a default value?  should we allow it to be `NULL`?

jmartin-sul:
https://github.com/sul-dlss/sul_pub/issues/1251#issuecomment-840879481 --
> See discussions in slack for updates:
>
> https://stanfordlib.slack.com/archives/C0T8CJD3L/p1620945545013300
> and
> https://stanfordlib.slack.com/archives/C0T8CJD3L/p1620933705006200
>
> Bottom line: expectation is that the controlled vocabulary is ok, and we should not get nils back, but we will have nils in our database because we have all authors for all time, while their API only returns authors currently at Stanford. Still some discussion on what is coming back from their API endpoint.

i _think_ the code change in this PR implements what's described there.  i've updated the comment on the model validation to reflect that.  out today, but happy to revisit on monday if i've got that wrong.

peetucket:
Agreed, I think the code here accommodates the discussion in slack -- allowing nil for anyone we can't values for, but constraining values when they exist.  Our filter query would still work fine (we would be looking for only people with a "public" visibility to push to ORCID).

peetucket (approved):

peetucket:
I'm thinking we are ready to merge this, and then deploy (though maybe early in the day in a controlled way so we can deal with any wrinkles).  We can then separately run the update task with a large timeframe in prod too to update all visibilities there.

PR #1277, opened 2021-05-12
refactor Author#unique_first_initial? logic into slightly smaller helper methods

jmartin-sul:
## Why was this change made?

try to get rid of rubocop exception and make things slightly more readable

existing unit tests

n/a


justinlittman (approved):
You're going to have a field day on this code base @jmartin-sul.

PR #1281, opened 2021-05-12
remove another unused method

peetucket:
## Why was this change made?

Found another unused method (only called from a unit test)

```
$ git grep text_search_query
lib/agent/author_name.rb:    def text_search_query
spec/lib/agent/author_name_spec.rb:  describe '#text_search_query' do
spec/lib/agent/author_name_spec.rb:        expect(all_names.text_search_query).to eq '"abc" or "def" or "qrs" or "xyz"'
```

### DIFF
--- a/lib/agent/author_name.rb
+++ b/lib/agent/author_name.rb
@@ -50,10 +50,6 @@ module Agent
     # an 'or' conjuction is likely to generate results that mostly match this variant,
     # but additional variants might add something when using an 'ExactMatch' search.
     # @return [String] name(s) to be queried in an OR (disjunction) query
-    def text_search_query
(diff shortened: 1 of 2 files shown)


lib/agent/author_name.rb
```diff
@@ -50,10 +50,6 @@ def full_name
     # an 'or' conjuction is likely to generate results that mostly match this variant,
     # but additional variants might add something when using an 'ExactMatch' search.
     # @return [String] name(s) to be queried in an OR (disjunction) query
-    def text_search_query
```
justinlittman, line 53: Do you need to get rid of the comments too?

peetucket: I believe that comment also makes sense given what `text_search_query` returns

justinlittman (approved):

PR #1287, opened 2021-05-13
remove CI config cruft

jmartin-sul:
## Why was this change made?

removing a line of cruft leftover from when i was iterating on https://github.com/sul-dlss/sul_pub/pull/1286

CI

circle CI config

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,7 +7,6 @@ jobs:
         BUNDLE_JOBS: 3
         BUNDLE_RETRY: 3
         BUNDLE_PATH: vendor/bundle
-        MYSQL_DATABASE: circle_test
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -7,7 +7,6 @@ jobs:
         BUNDLE_JOBS: 3
         BUNDLE_RETRY: 3
         BUNDLE_PATH: vendor/bundle
-        MYSQL_DATABASE: circle_test
```
jmartin-sul, line 10: i ended up just passing the default sul_pub test DB name to the circle mysql docker image config, instead of passing the DB name into the app and having the app config pick it up.

justinlittman (approved):

PR #1308, opened 2021-05-18
Removed data dir.

justinlittman:
## Why was this change made?
Cruft.

NA

NA


jmartin-sul (approved):
mostly LGTM, with a couple small questions:
1. i did a quick `git grep 'app.data'` on this branch to see if there were any lingering references, and found one: `lib/tasks/batch_import.rake:    BibtexIngester.new.ingest_from_source_directory(Rails.root.join('app/data/bibtex_to_ingest'))`.  so... it seems like we should either leave a stub `app/data/bibtex_to_ingest` dir, or remove/modify that rake task?
2. related, if we did get rid of `lib/tasks/batch_import.rake`, should we also get rid of the `BibtexIngester` class?  i see only two two non-spec references, on in a comment in `lib/csl/role_mapper.rb`, and one actual usage in the aforementioned rake task.

if there should be follow-on changes related to the above questions, i'm indifferent as to whether they happen in this PR or in a subsequent PR.  if you don't feel changes are needed, or if you want to make changes in another PR, feel free to merge this one as-is.

peetucket:
Thanks for digging into the referecences John.  I think we can leave the task and ingester class for now, but certainly this data seems like it never had a place being in the repo, so its a good step to ditch.

peetucket (approved):

jmartin-sul:
`BibtexIngester` removed in #1317

PR #1310, opened 2021-05-18
Adds fetching work details to Orcid client.

### LINKED ISSUE #1290: Add retrieving work details to Orcid client
For mapping to a publication, the full work record is needed, not just the work summary.

justinlittman:
closes #1290

## Why was this change made?
It will be used to populate the publication.

Unit

NA

### DIFF
--- a/lib/orcid/client.rb
+++ b/lib/orcid/client.rb
@@ -10,10 +10,12 @@ module Orcid
     # @param [string] ORCID ID for the researcher
     # @return [Hash]
     def fetch_works(orcidid)
-      response = conn.get("/v3.0/#{base_orcidid(orcidid)}/works")
-      raise "ORCID.org API returned #{response.status}" if response.status != 200
+      get("/v3.0/#{Orcid.base_orcidid(orcidid)}/works")
+    end
 
-      JSON.parse(response.body).with_indifferent_access
+    # Fetch the details for a work
+    def fetch_work(orcidid, put_code)
+      get("/v3.0/#{Orcid.base_orcidid(orcidid)}/work/#{put_code}")
     end
 
     # Add a new work for a researcher.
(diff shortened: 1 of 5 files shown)


lib/orcid/client.rb
```diff
@@ -8,13 +8,26 @@ class Client
...
+
+    def get(url)
+      response = conn.get(url)
       raise "ORCID.org API returned #{response.status}" if response.status != 200
 
       JSON.parse(response.body).with_indifferent_access
     end
 
-    private
+    def base_orcidid(orcidid)
```
peetucket, line 28: we are going to need this kind of method when adding the ORCID-based query to the WoS client (was just taking a peek at that now) ... is there a better place to put it to make it easier to use in a different class?

justinlittman: In general, an ORCID ID is considered the entire URL: `https://orcid.org/0000-0003-3437-349X`, not just `0000-0003-3437-349X`. That being said, I have no idea what WOS accepts.

peetucket: WoS wants the base identifier as well without the prefix

justinlittman: Good catch. Extracting.

lib/orcid/client.rb
```diff
@@ -41,8 +43,11 @@ def add_work(orcidid, work, token)
 
     private
 
-    def base_orcidid(orcidid)
-      orcidid[-19, 19]
+    def get(url)
+      response = conn.get(url)
+      raise "ORCID.org API returned #{response.status}" if response.status != 200
```
peetucket, line 48: just double-checking ... will this give us an HB alert?

peetucket (approved):
Just have one question, otherwise looks good.

lib/orcid/client.rb
justinlittman: If not caught and used in an environment in which HB is configured, I'd expect so.

PR #1313, opened 2021-05-19
ORCID API OKComputer check

### LINKED ISSUE #1262: Write OKComputer check for ORCID API

jmartin-sul:
## Why was this change made?

fixes #1262

bonus copypasta fix for `spec/lib/mais_spec.rb` test description

we discussed after standup this morning, and people were OK with the VCR approach used in this PR, but i'm happy to switch to the pattern used for the `Mais` check if that's preferable (i.e. more explicitly mocking the `client.fetch_works` call/result -- i hadn't looked closely at `mais_spec.rb` until after that discussion). 

unit tests

added a setting for test orcid ID for the new check

### DIFF
--- a/spec/lib/mais_spec.rb
+++ b/spec/lib/mais_spec.rb
@@ -22,7 +22,7 @@ describe Mais do
         allow(client).to receive(:fetch_orcid_users).and_raise(StandardError, 'Fail!')
       end
 
-      it 'returns true' do
+      it 'returns false' do
         expect(described_class.working?).to be(false)
       end
     end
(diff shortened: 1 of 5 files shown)


spec/lib/mais_spec.rb
```diff
@@ -22,7 +22,7 @@
         allow(client).to receive(:fetch_orcid_users).and_raise(StandardError, 'Fail!')
       end
 
-      it 'returns true' do
+      it 'returns false' do
```
jmartin-sul, line 25: stumbled on this, seemed wrong, tried to fix

justinlittman (approved):

PR #1315, opened 2021-05-19
Query WoS by ORCID when harvesting

### LINKED ISSUE #411: When harvesting for a given author, search WoS by ORCIDs in addition to the usual name queries
In cases when we have an ORCID for the author, use this ORCID to issue an additional research ID query against WoS to improve the results

peetucket:
## Why was this change made?

Note: we can do some quick validation before merging.

Fixes #411 - query WoS by ORCID for authors with an ORCID (in addition to the name query)

Also, refactor the WOS query classes to reduce some duplication when adding the new queries.

Note: may need to update some documentation in the wiki about common queries... each search for "QueryAuthor" and update cases like this on this page: https://github.com/sul-dlss/sul_pub/wiki/Useful-console-commands-(harvest,-WoS,-Pubmed,-others)

```
# WoS
author_query = WebOfScience::QueryAuthor.new(author)
author_query.send(:author_query) # the query that will be sent to WoS, taking into account alternate names and institutions
```

This needs to be:

```
# WoS
author_query = WebOfScience::QueryAuthor.new(author)
author_query.name_query.send(:name_query) # the name query that will be sent to WoS, taking into account alternate names and institutions
author_query.orcid_query.send(:orcid_query) # the orcid query that will be sent to WoS (only makes sense if the user has an orcid, else returns an error)
```

To execute name or ORCID only searches:
```
author_query = WebOfScience::QueryAuthor.new(author)
names_uids = author_query.name_query.uids
orcid_uids = author_query.orcid_query.uids

```

New tests


lib/smci_report.rb
```diff
@@ -144,20 +144,16 @@ def run
                                                              middle_name: middle_name, institution: institution.strip)
             end
             # end check for institutions for this author
-            author_query = WebOfScience::QueryAuthor.new(author, symbolicTimeSpan: symbolicTimeSpan) # setup the WOS name query
-            uids = author_query.uids # now fetch all of the WOS_UIDs for the publications for this author by running the name search
-            logger.info(author_query.send(:author_query)) # log the name query being sent to WoS
+            name_query = WebOfScience::QueryName.new(author, symbolicTimeSpan: symbolicTimeSpan) # setup the WOS name query
+            uids = name_query.uids # now fetch all of the WOS_UIDs for the publications for this author by running the name search
+            logger.info(name_query.send(:name_query)) # log the name query being sent to WoS
```
peetucket, line 149: this class got renamed

lib/pubmed/query_author.rb
```diff
@@ -13,6 +13,8 @@ def initialize(author, options = {})
     # Find all pmids for an author
     # @return [Array<String>] pmids
     def pmids
+      return [] unless valid?
+
       resp = client.search(term, addl_args)
```
peetucket, line 18: no need to make the API call if it's not valid

lib/smci_report.rb
```diff
@@ -144,20 +144,16 @@ def run
...
-              params[:queryParameters].delete(:timeSpan)
-              params[:queryParameters][:order!] = %i[databaseId userQuery symbolicTimeSpan queryLanguage] # according to WSDL
-            end
-            retriever = WebOfScience.queries.search(params)
-            uids = retriever.merged_uids # now fetch all of the WOS_UIDs for the publications for this author by running the orcid search
+            author = Author.new(orcidid: orcid)
+            orcid_query = WebOfScience::QueryOrcid.new(author, symbolicTimeSpan: symbolicTimeSpan) # setup the WOS orcid query
+            uids = orcid_query.uids # now fetch all of the WOS_UIDs for the publications for this author by running the orcid search
+            logger.info(orcid_query.send(:orcid_query)) # log the orcid query being sent to WoS
           end
```
peetucket, line 157: we can use our shiny new query orcid class to search by orcid instead of running it manually

lib/web_of_science/queries.rb
```diff
@@ -123,6 +123,39 @@ def params_for_fields(fields)
...
+      end
+      params
+    end
+
+    # Use Settings.WOS.ACCEPTED_DBS to define collections without any fields retrieved
+    # @return [Array<Hash>]
+    def empty_fields
+      Settings.WOS.ACCEPTED_DBS.map { |db| { collectionName: db, fieldName: [''] } }
+    end
+
```
peetucket, line 158: these methods used to be defined in the `QueryAuthor` class, but since that class now servers to wrap the name and orcid queries and these methods are used by both of those classes, they got moved into a more general queries class

lib/web_of_science/query_author.rb
```diff
@@ -1,88 +1,29 @@
...
-    def quote_wrap(terms)
-      terms.reject(&:empty?).uniq.map { |x| "\"#{x.delete('"')}\"" }
-    end
-
-    # Use Settings.WOS.ACCEPTED_DBS to define collections without any fields retrieved
-    # @return [Array<Hash>]
-    def empty_fields
-      Settings.WOS.ACCEPTED_DBS.map { |db| { collectionName: db, fieldName: [''] } }
+      orcid_query.valid? || name_query.valid?
     end
```
peetucket, line 27: this class used to execute name queries only, but it now wraps both the name and orcid queries ... methods were either moved to the new name only query class or to the more generic queries class if they are also used in the orcid query class

lib/web_of_science/query_author.rb
```diff
@@ -1,88 +1,29 @@
...
 
-    # Find all WOS-UIDs for an author
+    # Find all WOS-UIDs for an author using both ORCID and Name queries, and then de-dupe
     # @return [Array<String>] WosUIDs
     # Implementation note: these records have a relatively small memory footprint, just UIDs
     def uids
-      queries.search(author_query).merged_uids
+      (orcid_query.uids + name_query.uids).uniq
     end
```
peetucket, line 22: we now join the results from the name and orcid queries and unique to prevent dupes

lib/web_of_science/query_orcid.rb
```diff
@@ -0,0 +1,38 @@
...
+
+    delegate :queries, to: :WebOfScience
+
+    attr_reader :orcid, :options
+
+    def orcid_query
+      queries.construct_uid_query("RID=(\"#{Orcid.base_orcidid(orcid)}\")", options)
+    end
+  end
+end
```
peetucket, line 38: a new class to run orcid queries

lib/web_of_science/query_name.rb
```diff
@@ -0,0 +1,64 @@
...
+      queries.construct_uid_query("AU=(#{quote_wrap(names).join(' OR ')}) AND AD=(#{quote_wrap(institutions).join(' OR ')})", options)
+    end
+
+    # @param [Array<String>] terms
+    # @param [Array<String>] the same terms, minus any empties or duplicates, wrapped in double quotes
+    def quote_wrap(terms)
+      terms.reject(&:empty?).uniq.map { |x| "\"#{x.delete('"')}\"" }
+    end
+  end
+end
```
peetucket, line 64: a new class to run name only queries (with the code all coming from what used to be in `QueryAuthor`)

spec/lib/web_of_science/harvester_spec.rb
```diff
@@ -13,7 +13,7 @@
 
   after(:all)  { savon.unmock! }
 
-  let(:author) { create :russ_altman }
+  let(:author) { create :russ_altman, :blank_orcid } # use a blank orcid, because these mocks work with the name query
   let(:wos_uids) { %w[WOS:A1976BW18000001 WOS:A1972N549400003] } # from wos_retrieve_by_id_response.xml
```
peetucket, line 17: this spec does not use VCR and is mocked to run only name queries, so will not behave correctly if we let it run ORCID queries as well -- since it is basically supposed to test the logic of the harvester class itself, it seems sufficient to verify it continues to make the call to our `QueryAuthor` class, which actually runs the underlying name or orcid queries as needed.  These class all have updated tests below that use VCR to test all of the new logic implemented here...

spec/lib/web_of_science/queries_spec.rb
```diff
@@ -140,6 +140,25 @@
...
+    end
+
+    it 'constructs a full query to send to WoS with provided options' do
+      query = wos_queries.construct_uid_query('AU=somebody', { symbolicTimeSpan: '2week' })
+      expect(query).to include(queryParameters: { databaseId: 'WOK', order!: %i[databaseId userQuery symbolicTimeSpan queryLanguage],
+                                                  queryLanguage: 'en', symbolicTimeSpan: '2week', userQuery: 'AU=somebody' })
+    end
+  end
+
   describe '#params_for_search' do
```
peetucket, line 162: these specs are moved from the `QueryAuthor` class since these methods moved above

spec/lib/web_of_science/query_author_spec.rb
```diff
@@ -21,122 +16,78 @@
...
-  context 'for a single alternate identity with invalid data' do
-    describe '#names' do
-      let(:author_one_identity) { create :author }
-      let(:bad_alternate_identity) { create :author_identity }
-
-      before do
-        bad_alternate_identity.update(first_name: '.')
-        author_one_identity.author_identities << bad_alternate_identity
-      end
+  describe '#uids with both the orcid and name queries invalid' do
```
peetucket, line 85: all of these tests were just moved to the new name specific class below where the methods were moved to, they were not actually deleted: `QueryName`

spec/lib/web_of_science/query_orcid_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+
+  describe '#uids with an invalid query' do
+    subject(:query_no_orcid) { described_class.new(author_blank_orcid) }
+
+    it 'returns an empty array' do
+      expect(query_no_orcid).not_to be_valid
+      expect(query_no_orcid.uids).to be_empty
+    end
+  end
+end
```
peetucket, line 52: tests for our new orcid query class

jmartin-sul (approved):
tried to review thoroughly, LGTM, seems like everything was accounted for.  still, since it's a large-ish PR with some refactoring, and since i'm relatively new to the code, might still be good to have a second set of eyes (i.e. @justinlittman?) give it a glance.

refraining from merging, as discussed after standup.

jmartin-sul:
@peetucket thanks for the inline explanatory comments, very helpful

peetucket:
Potential sanity check validation I thought of.  BUT i don't think this really works well in anything except production because the ORCIDs we have in the database in other environments are going to be Sandbox ORCIDs and thus produce no results when searching against WoS.  Still may be worth doing though just to ensure we don't get a big spike.

Alternatively, we can test a couple authors manually both in production and uat.  

```
# find the number of contributions in the last day
bundle exec rails c -e production
author = Author.find_by(cap_profile_id:'35324')
author.contributions.where('created_at > ?',Date.today-1.day).size
```

```
# run a harvest for that person
RAILS_ENV=production bundle exec rake harvest:author_update['35324']
```

```
# re-check the number of contributions in the last day
bundle exec rails c -e production
author = Author.find_by(cap_profile_id:'35324')
author.contributions.where('created_at > ?',Date.today-1.day).size
```

Sample full harvest test script:

1. Deploy branch to qa
2. See how many people have ORCID in the database.

```
Author.where('orcidid is not null').size
=> 140
```

3. Run a full harvest in screen:

```
RAILS_ENV=production bundle exec rake 'harvest:all_authors_update'
```

4. See how many contributions were added in the last day:

```
 Contribution.where('created_at > ?',Date.today-1.day).size
=> 68
```

5. Inspect for a given author with an ORCID:

```
author = Author.where('orcidid is not null').first
author.contributions.where('created_at > ?',Date.today-1.day).size
```

jmartin-sul:
i like that manual testing on uat idea

lib/smci_report.rb
justinlittman: Can we avoid calling `send`?

lib/web_of_science/query_orcid.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+module WebOfScience
+  # Use author orcid to find WOS publications for an Author
+  # fetch the wos uids for the given author
+  # e.g. WebOfScience::QueryOrcid.new(author).uids
+  class QueryOrcid
+    def initialize(author, options = {})
+      raise(ArgumentError, 'author must be an Author') unless author.is_a? Author
```
justinlittman, line 9: Why is this everywhere?

lib/web_of_science/query_orcid.rb
```diff
@@ -0,0 +1,38 @@
...
+
+      @orcid = author.orcidid
```
justinlittman, line 11: For consistency with other orcid code, can you call this `@orcidid`?

justinlittman (approved):
Approve pending consideration of comments.

lib/smci_report.rb
peetucket: The `name_query` method is private, really never used externally, but I thought it was useful to log in this reporting class so we could see what was actually run (we feed the data to the class from a spreadsheet, so its a bit different than what happens in the code, where it comes from the database).

lib/web_of_science/query_orcid.rb
peetucket: yeah, probably overkill, was copypasta.

i removed and left it only in the `QueryAuthor` class, which calls these sub-classes

lib/web_of_science/query_orcid.rb
peetucket: yup, changed

peetucket:
See #1344 for validation test idea

PR #1317, opened 2021-05-19
remove bibtext ingester and other unused classes

peetucket:
## Why was this change made?

Remove some unused classes and rake tasks.

Test suite


jmartin-sul (approved):
when `grep`ing, i did notice one other stray `BibtexIngester` reference:
```
lib/csl/role_mapper.rb:          # This is from BibtexIngester.convert_bibtex_record_to_pub_hash
```

but otherwise this LGTM, happy to merge as-is, and maybe clean that comment up in a follow on.

jmartin-sul:
er, actually, three references in comments, a couple were mis-cased:
```
% git grep -i BibtexIngester                                                                    main
lib/csl/bibtex_mapper.rb:  # Convert BibTexIngester content into a CSL format
lib/csl/bibtex_mapper.rb:    # Convert BibTexIngester authors into CSL authors
lib/csl/role_mapper.rb:          # This is from BibtexIngester.convert_bibtex_record_to_pub_hash
```

jmartin-sul:
small follow on for those comments:  https://github.com/sul-dlss/sul_pub/pull/1318

PR #1321, opened 2021-05-20
[HOLD] Adds task for adding works to ORCID.org.

### LINKED ISSUE #1261: Write task for adding publications to ORCID profile
The task will be provided with a list of SUNETID, ORCID ID, ORCID token, and permission type.

For each contribution for an approved publication that has not been added to ORCID profile and for an author that has ORCID permission type is “update” and CAP visibility is “public”:
1. Post a new work to ORCID API.
2. If the post is successful, update contribution with timestamp indicating that publication was added to ORCID profile for that author.
3. If the post is unsuccessful, log an error and Honeybadger alert.

Note:
* This will require adding a timestamp field to contributions model for added to ORCID profile.
* Existing sul-pub logging conventions for tasks should be followed.

justinlittman:
Note: hold for testing on cap-dev-a

closes #1261

## Why was this change made?
Support adding publications to ORCID.org for an author.

Sandbox/locally, unit

NA

### DIFF
--- a/lib/orcid.rb
+++ b/lib/orcid.rb
@@ -4,7 +4,11 @@
 module Orcid
   # @return [Orcid::Client]
   def self.client
-    Orcid::Client.new
+    @@client ||= Orcid::Client.new
+  end
+
+  def self.logger
+    @@logger ||= Logger.new(Settings.ORCID.LOG)
   end
 
   # Extract the ID part from an ORCID ID.
--- a/lib/orcid/add_works.rb
+++ b/lib/orcid/add_works.rb
@@ -1,0 +1,49 @@
+# frozen_string_literal: true
+
+module Orcid
+  # Service for adding works to a researcher's ORCID profile.
+  class AddWorks
+    # @param [Logger] logger
+    def initialize(logger: nil)
+      @logger = logger
+    end
+
+    # @param [Array<Mais::Client::OrcidUser>] orcid_users to add works for
+    # @return [Integer] count of works added.
+    def add_all(orcid_users)
+      count = orcid_users.map { |orcid_user| add_for_orcid_user(orcid_user) }.sum
+      logger&.info("#{self.class} Updated #{count} contributor records from #{orcid_users.size} ORCID users.")
+      count
+    end
+
+    # @param [Mais::Client::OrcidUser] orcid_user to add works for
+    # @return [Integer] count of works added.
+    def add_for_orcid_user(orcid_user)
+      return 0 unless orcid_user.update?
+
+      author = Author.find_by(sunetid: orcid_user.sunetid)
+      return 0 if author.nil? || author.cap_visibility != 'public'
+
+      logger&.info("#{self.class} - author #{author.id} - adding publications to #{orcid_user.orcidid}")
+
+      contributions = Contribution.where(author: author, status: 'approved', visibility: 'public', orcid_put_code: nil)
+      contributions.map { |contribution| add_work(author, contribution, orcid_user) ? 1 : 0 }.sum
+    end
+
+    private
+
+    attr_reader :logger
+
+    # @return [Boolean] true if work added.
+    def add_work(author, contribution, orcid_user)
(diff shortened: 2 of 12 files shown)


lib/orcid.rb
```diff
@@ -4,7 +4,11 @@
 module Orcid
   # @return [Orcid::Client]
   def self.client
-    Orcid::Client.new
+    @@lcient ||= Orcid::Client.new
```
peetucket, line 7: typo

lib/orcid/identifier_type_mapper.rb
```diff
@@ -0,0 +1,75 @@
...
+      ssrn
+      uri
+      urn
+      wosuid
+      zbl
+    ].freeze
+
+    # @return [String] ORCID id type or nil if no matching
+    def self.to_orcid_id_type(sul_pub_id_type)
+      id_type = ORCID_ID_TYPE_TO_SUL_PUB_ID_TYPE.invert.fetch(sul_pub_id_type, sul_pub_id_type)
```
peetucket, line 62: i'm not quite sure i understand what is happening here... and is there a reason doi is not in the `ORCID_ID_TYPE_TO_SUL_PUB_ID_TYPE` list?

lib/tasks/orcid.rake
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
```
peetucket, line 1: is the idea that we would have two separate schedules, one to update our author's table with ORCIDs that come from MAIS and one to push the works for all of those users?  both actions start with this same API call to MAIS right?

`Mais.client.fetch_orcid_users`

justinlittman: They are two completely independent tasks. However, both tasks start with retrieving orcid info from MAIS.

peetucket: Understood, though they could in theory be done at the same time (update users and push to ORCID as you go), since both tasks will be happening on a schedule and one is dependent on the other.

lib/orcid/identifier_type_mapper.rb
peetucket: never mind, i think i got what is happening

lib/tasks/orcid.rake
justinlittman: They're both independent. Pushing the works to Orcid does NOT used the orcid id stored in the db. Rather, it uses the orcid id and access token returned from MAIS.

jmartin-sul (approved):
a few small questions/suggestions, but nothing for which i'd hold up merging

lib/orcid/add_works.rb
```diff
@@ -0,0 +1,49 @@
...
+    private
+
+    attr_reader :logger
+
+    # @return [Boolean] true if work added.
+    def add_work(author, contribution, orcid_user)
+      work = Orcid::PubMapper.map(contribution.publication.pub_hash)
+      contribution.orcid_put_code = Orcid.client.add_work(orcid_user.orcidid, work, orcid_user.access_token)
+      contribution.save!
+      logger&.info("#{self.class} - author #{author.id} - adding publication #{contribution.publication.id} with put-code #{contribution.orcid_put_code}")
```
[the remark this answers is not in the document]
justinlittman: Fixed.

spec/lib/orcid/add_works_spec.rb
```diff
@@ -0,0 +1,146 @@
...
+    context 'when Author cap visibility is not public' do
+      let(:cap_visibility) { 'private' }
+
+      it 'skips' do
+        expect(contribution_count).to be_zero
+      end
+    end
+
+    context 'when Contribution is approved, visible, and does not have a put-code' do
+      let!(:contribution) { create :contribution, author: author, publication: publication }
```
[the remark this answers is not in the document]
justinlittman: Made rubocop a little unhappy, but passified with an inline disable.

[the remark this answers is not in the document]
jmartin-sul: thanks!

PR #1328, opened 2021-05-21
Mapping

justinlittman:
## Why was this change made?
To map from Orcid works to pub hashes and vice versa.

Unit

NA

### DIFF
--- a/lib/csl/citation.rb
+++ b/lib/csl/citation.rb
@@ -1,26 +1,12 @@
 # frozen_string_literal: true
 
 require 'citeproc/ruby'
-# You can install more recent styles using the 'csl-styles' gem or you can just
-# download the styles you need (just point CSL::Style.root to the directory and
-# you can load any style with just its name using CSL::Style.load).
-require 'csl/styles'
 
 module Csl
   class Citation
-    attr_reader :csl_renderer, :pub_hash
+    attr_reader :pub_hash
 
-    CSL_STYLE_APA = CSL::Style.load('apa')
-    CSL_STYLE_MLA = CSL::Style.load('modern-language-association')
-
-    CSL_STYLE_CHICAGO = CSL::Style.load('chicago-author-date')
-    CSL_STYLE_CHICAGO_ET_AL = begin
-      # Modify the bibliography attributes so it uses 'et al.' after 5 authors
-      style_et_al = CSL::Style.load('chicago-author-date')
-      style_et_al.bibliography.attributes['et-al-min'] = 1
-      style_et_al.bibliography.attributes['et-al-use-first'] = 5
-      style_et_al
-    end
+    delegate :to_apa_citation, :to_mla_citation, :to_chicago_citation, :to_bibtex, to: :renderer
 
     # @param pub_hash [Hash] a Publication.pub_hash
     def initialize(pub_hash)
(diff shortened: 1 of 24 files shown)


lib/csl/citation.rb
```diff
@@ -1,26 +1,12 @@
 # frozen_string_literal: true
 
 require 'citeproc/ruby'
-# You can install more recent styles using the 'csl-styles' gem or you can just
-# download the styles you need (just point CSL::Style.root to the directory and
-# you can load any style with just its name using CSL::Style.load).
-require 'csl/styles'
```
justinlittman, line 7: Moving rendering into its own class so that can be re-used.

lib/csl/mapper.rb
```diff
@@ -59,7 +59,7 @@ def csl_doc
         # Add journal information if it exists.
         if pub_hash.key?(:journal)
           if pub_hash[:journal][:name].present?
-            cit_data_hash['type'] = 'article-journal'
+            cit_data_hash['type'] = 'article-journal' unless cit_data_hash['type'] == 'inproceedings'
```
justinlittman, line 62: Arcadia noticed that we were incorrect setting type for `inproceedings`.

lib/orcid/add_works.rb
```diff
@@ -41,6 +41,9 @@ def add_work(author, contribution, orcid_user)
       contribution.save!
       logger&.info("#{self.class} - author #{author.id} - added publication #{contribution.publication.id} with put-code #{contribution.orcid_put_code}")
       true
+    rescue Orcid::PubMapper::PubMapperError => e
```
justinlittman, line 44: PubMapperErrors are worthy of logging, but not notifying.

lib/orcid/pub_author_mapper.rb
```diff
@@ -0,0 +1,85 @@
...
+        }
+      }
+    end
+
+    private
+
+    attr_reader :author_hash
+
+    def map_credit_name
+      clean_name(author_hash[:name].presence) ||
```
justinlittman, line 36: We take various strategies to constructing a name. The ordering of those strategies is deliberate.

lib/orcid/pub_identifier_mapper.rb
```diff
@@ -0,0 +1,84 @@
...
+
+    def map
+      ids = map_identifiers(pub_hash[:identifier], 'self') +
+            map_identifiers(pub_hash.dig(:journal, :identifier), 'part-of') +
+            map_identifiers(pub_hash.dig(:conference, :identifier), 'part-of') +
+            map_identifiers(pub_hash.dig(:series, :identifier), 'part-of')
+
+      raise PubMapper::PubMapperError, 'An identifier is required' if ids.empty?
+
+      clean_issns(ids)
```
justinlittman, line 25: A couple identifier pecularities cause us problems, so we clean them here.

lib/csl/citation.rb
```diff
@@ -31,43 +17,28 @@ def initialize(pub_hash)
     # @return [Hash]
     def citations
       {
-        apa_citation: to_apa_citation,
-        mla_citation: to_mla_citation,
-        chicago_citation: to_chicago_citation
+        apa_citation: renderer.to_apa_citation,
+        mla_citation: renderer.to_mla_citation,
+        chicago_citation: renderer.to_chicago_citation
```
peetucket, line 22: based on the delegation above in line 9, is calling `renderer.` here still necessary?

lib/csl/mapper.rb
peetucket: this is used when mapping pub_hash to the citation generator right?  so this basically would change for how we format citations going forward

lib/orcid/add_works.rb
peetucket: so is the idea that we monitor the logs for a bit to see the types of unexpected mapping issues that are occurring?

peetucket (commented):
Initial pass with some comments...

lib/csl/mapper.rb
justinlittman: I suspect it doesn't actually change any MLA, Chicago, or APA citations, but will only affect Bibtex citations (which are new).

lib/orcid/add_works.rb
justinlittman: We should keep in eye out, but this is primarily going to report issues like "this publication doesn't have an identifier." Any significant issues will be HB alerted.

lib/csl/citation.rb
justinlittman: Nope. Removed.

spec/lib/orcid/pub_identifier_mapper_spec.rb
```diff
@@ -246,6 +258,12 @@
 
       it 'ignores self identifier' do
```
jmartin-sul, line 259: is this test title still accurate?

jmartin-sul (changes requested):
only issues i noticed were a couple test titles that might need fixing, otherwise 👍

spec/lib/orcid/pub_identifier_mapper_spec.rb
justinlittman: Accurate, but not fully descriptive. Fixed.

spec/lib/orcid/add_works_spec.rb
```diff
@@ -68,12 +68,7 @@
       # Default factory pub_hash is missing an identifier so allowing factory to create publication.
       let!(:contribution) { create :contribution, author: author } # rubocop:disable RSpec/LetSetup
 
-      before do
-        allow(NotificationManager).to receive(:error)
-      end
-
       it 'notifies' do
```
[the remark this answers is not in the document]
justinlittman: Fixed and added `NotificationManager` check (since NotificationManager is mocked, I don't think it is overkill).

jmartin-sul (approved):

jmartin-sul:
@justinlittman / @peetucket -- wasn't sure if peter wanted to do more review, so i held off on merging, but approved, feel free to merge if sufficiently reviewed.

peetucket:
I'm ok without a re-review if you are both satisfied.

jmartin-sul:
> I'm ok without a re-review if you are both satisfied.

merged!

PR #1332, opened 2021-05-26
Changes pub_hash validation to support publications submitted by CAP.

### LINKED ISSUE #1331: Error when creating publication via create endpoint
## Backtrace

    line 88 of [PROJECT_ROOT]/app/controllers/publications_controller.rb: create

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/50046/faults/79596536)

justinlittman:
closes #1331

## Why was this change made?
CAP submitting an application was resulting in an error due to validation.

NA

pub_hash_schema.yml

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -111,7 +111,7 @@ RSpec/MultipleExpectations:
 # Offense count: 282
 # Configuration parameters: AllowSubject.
 RSpec/MultipleMemoizedHelpers:
-  Max: 18
+  Max: 19
 
 # Offense count: 288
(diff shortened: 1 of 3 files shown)


peetucket:
I added a commit to update the tests to add the `additionalProperties` and added a new test that better mimics what they are actually sending us.  I tested in localhost and it seems fine to me.  Thoughts?

peetucket (approved):
I think LGTM for testing in UAT

peetucket:
BTW - the updates to the tests cause them to fail on main and pass here, so that's a good sign.

justinlittman:
Looks good to me.

PR #1335, opened 2021-05-26
do not fail validation on pub_hash_schema, but alert HB instead

peetucket:
## Why was this change made?

Experiment to alert HB instead of failing validation on pub_hash schema validation errors (to prevent user errors)

Updated tests

### DIFF
--- a/app/validators/publication_validator.rb
+++ b/app/validators/publication_validator.rb
@@ -2,6 +2,10 @@
 
 class PublicationValidator < ActiveModel::Validator
   def validate(record)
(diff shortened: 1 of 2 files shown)


justinlittman (changes requested):
I would make this change in `app/validators/publication_validator` rather than here.

peetucket:
> I would make this change in `app/validators/publication_validator` rather than here.

oh, yes, that makes much more sense -- i missed that class and now i understand better how it works

justinlittman (approved):

PR #1350, opened 2021-06-08
use sandbox orcid account for checking in global config

peetucket:
## Why was this change made?

We have configured the sandbox ORCID in `config/settings.yml` and thus it is helpful to have a legitimate sandbox ORCID URL to check against so the okcomputer checks work as expected when deployed.  In production, we will override this setting with a valid production ORCID along with the valid production credentials for okcomputer checks to work there too.  Note that some VCR cassettes had to be updated (including some manual fudging for the 500 check) here.

Existing tests


spec/spec_helper.rb
```diff
@@ -112,6 +114,14 @@
     links_password if interaction.request.uri.include? Clarivate::LinksClient::LINKS_HOST
   end
 
+  # ORCID API filters
+  c.filter_sensitive_data('refresh_token') do |interaction|
+    if interaction.request.uri.include? Settings.ORCID.BASE_AUTH_URL
+      token_match = interaction.response.body.match(/"refresh_token":"(.*?)"/)
```
jmartin-sul, line 120: this might just be me forgetting some regex syntax, but what's the need for the `?` here?  shouldn't the `.*` match zero or more of any character?  is the intent of the `?` to make that match optional, or does it serve a different use from how i'm interpreting it?

jmartin-sul: the idea is to match anything that looks like `"refresh_token":"<MAYBE SOME TOKEN CHARACTERS>"`?  or something different?

jmartin-sul (approved):
question about a token filtering regex, but LGTM overall, so feel free to merge if that's actually a non-issue

spec/spec_helper.rb
peetucket: i think that may be the idea --- i just did a copy and paste job from the other uses of this same technique for the other APIs and it worked exactly as expected, so I went with it

peetucket: it basically ensures we don't have refresh_token's hanging out in vcr cassettes that get checked in... i tried to do something similar earlier for the MaIS API for the user authentication tokens, but that got more complicated and also caused problems with expectations in the tests, so I left it out

so for the MaIS API tests, we have to manually alter the cassettes unfortunately, though it hopefully will not be very often

PR #1351, opened 2021-06-08
enable cron schedule for orcid pushing in dev and uat

### LINKED ISSUE #1348: Turn on orcid push rake task for all users in cron
Use whenever gem to enable the rake task that pushes works for all others in cap-dev and run it daily to find out if there are any issues.

peetucket:
## Why was this change made?

Fixes #1348 -- note we are deliberately leaving this schedule off in production until we are complete with testing


jmartin-sul (approved):
neat trick, using the `harvester_*` roles for env specific cron jobs

PR #1353, opened 2021-06-09
tweaks to deleting orcid works

jmartin-sul:
## Why was this change made?

easy touchups for suggestions made when reviewing #1346.  tried to lump them in commits ordered from what i thought would be least to most controversial, in case any of the changes were undesired.

unit

n/a


peetucket (approved):
I think all of these changes make sense.

peetucket:
Reluctant to merge without a thumbs up from Justin L and also not urgent, so I am opting to wait for his confirmation before merging, but thanks for doing the work John.

PR #1379, opened 2021-08-05
look for year in pubmed data in alternate locations

### LINKED ISSUE #1378: Pubmed sometimes puts the year in a new spot
Pubmed puts the year into a few potential places in their source XML.  We are already looking at three, but some publications came through with the year in other places, so we should use those too.  

This came from a support request where a user didn't see the year in their citation.  After examining our code, I found the spot where we locate the year from the pubmed source xml:

Year lookup: https://github.com/sul-dlss/sul_pub/blob/main/app/models/pubmed_source_record.rb#L142-L155

We already look in a few different spots, but for the example publication (PMID "23388678"), I noticed the year in a some other locations.  We should just add these to our lookup table at the end.

```
pmsr = PubmedSourceRecord.find_by(pmid:'23388678')
puts pmsr.source_data

....
  <PubmedData>
    <History>
      <PubMedPubDate PubStatus="entrez">
        <Year>2013</Year>
        <Month>2</Month>
        <Day>8</Day>
        <Hour>6</Hour>
        <Minute>0</Minute>
      </PubMedPubDate>
      <PubMedPubDate PubStatus="pubmed">
        <Year>2013</Year>
        <Month>2</Month>
        <Day>8</Day>
        <Hour>6</Hour>
        <Minute>0</Minute>
      </PubMedPubDate>
      <PubMedPubDate PubStatus="medline">
        <Year>2013</Year>
        <Month>7</Month>
        <Day>10</Day>
        <Hour>6</Hour>
        <Minute>0</Minute>
      </PubMedPubDate>
    </History>
```

peetucket:
## Why was this change made?

Fixes #1378 -- some older pubmed data has the year in a different location in the XML than we were looking for, which was causing it be missing from our data structure (pub_hash), which was then preventing us from creating the citation correctly.  We should look for it in other places in the XML too.

Also, we should be able to rebuild the pub_hash for anything with a provenance of pubmed, even if it happens to have a `wos_uid` in the record, this allows for that.

New test added, existing tests passing.

### DIFF
--- a/app/models/publication.rb
+++ b/app/models/publication.rb
@@ -155,7 +155,7 @@ class Publication < ApplicationRecord
 
   # @return [Boolean] true if .save is successful
   def rebuild_pub_hash
-    raise 'rebuilding WOS records is unsupported, unimplemented' if wos_uid
+    raise 'rebuilding WOS records is unsupported, unimplemented' if wos_pub?
 
     if sciencewire_id
       sw_source_record = SciencewireSourceRecord.find_by_sciencewire_id(sciencewire_id)
--- a/app/models/pubmed_source_record.rb
+++ b/app/models/pubmed_source_record.rb
@@ -142,7 +142,10 @@ class PubmedSourceRecord < ApplicationRecord
     year_xpaths = [
       'MedlineCitation/Article/Journal/JournalIssue/PubDate/Year',
       'MedlineCitation/Article/ArticleDate/Year',
-      'PubmedData/History/PubMedPubDate[@PubStatus="accepted"]/Year'
+      'PubmedData/History/PubMedPubDate[@PubStatus="accepted"]/Year',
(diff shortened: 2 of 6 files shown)


app/models/publication.rb
```diff
@@ -155,7 +155,7 @@ def update_from_pubmed
 
   # @return [Boolean] true if .save is successful
   def rebuild_pub_hash
-    raise 'rebuilding WOS records is unsupported, unimplemented' if wos_uid
+    raise 'rebuilding WOS records is unsupported, unimplemented' if wos_pub?
```
peetucket, line 158: a record can have a provenance of pubmed, and still have a wos_uid added to it -- in this case, we can rebuild the pub_hash from the pubmed record if we want ... this is a better constraint check

justinlittman (approved):

PR #1386, opened 2021-08-23
Upgrade to Rails 6.1

jcoyne:


config/environments/test.rb
```diff
@@ -1,11 +1,13 @@
...
+
 Rails.application.configure do
   # Settings specified here will take precedence over those in config/application.rb.
 
-  # The test environment is used exclusively to run your application's
-  # test suite. You never need to work with it otherwise. Remember that
-  # your test database is "scratch space" for the test suite and is wiped
-  # and recreated between test runs. Don't rely on the data there!
-  config.cache_classes = true
+  config.cache_classes = false
```
justinlittman, line 9: Was this switch to `false` intentional?

config/initializers/filter_parameter_logging.rb
```diff
@@ -1,4 +1,6 @@
 # Be sure to restart your server when you modify this file.
 
 # Configure sensitive parameters which will be filtered from the log file.
-Rails.application.config.filter_parameters += [:password]
+Rails.application.config.filter_parameters += [
```
justinlittman, line 4: `:password` was dropped. Intentional?

justinlittman (changes requested):
Just a couple of questions and file conflict.

config/initializers/filter_parameter_logging.rb
jcoyne: yes, but we don't collect passwords, right?

config/environments/test.rb
jcoyne: yes.  this is the Rails default caching doesn't benefit us in test mode.

justinlittman (approved):

PR #1402, opened 2021-10-01
log warnings for errors; do not alert HB on invalid token

### LINKED ISSUE #1401: When the response is an invalid token, don't raise an error. Just log it.
When an error is raised we get these in Honeybadger, however this is believed to be caused by the user revoking our access to update their ORCID profile.   There is no action for us to take.

```
RuntimeError: ORCID.org API returned 401 ({  "error" : "invalid_token",  "error_description" : "Invalid access token: 4e04c1d2-c385-4e78-8ca0-6d5e2e29a03c"}) for: {"type":"journal-article","visibility":"public","title":{"title":{"value":"Comparison on two ways of phonon spectral property reconstruction from thermal spectroscopy experiment"}},"external-ids":{"external-id":[{"external-id-type":"doi","external-id-value":"10.1016/j.ijheatmasstransfer.2021.121119","external-id-url":"https://...
```
## Backtrace

    line 77 of [PROJECT_ROOT]/lib/orcid/client.rb: add_work
    line 41 of [PROJECT_ROOT]/lib/orcid/add_works.rb: add_work
    line 30 of [PROJECT_ROOT]/lib/orcid/add_works.rb: block in add_for_orcid_user

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/50046/faults/80989125)

peetucket:
## Why was this change made?

Fixes #1401 - do not alert HB if the user has revoked their ORCID token (we do not get notified about this, and it is an expected error that we cannot do anything about currently)

Needs a new test

### DIFF
--- a/lib/orcid/add_works.rb
+++ b/lib/orcid/add_works.rb
@@ -42,8 +42,8 @@ module Orcid
       contribution.save!
       logger&.info("#{self.class} - author #{author.id} - added publication #{contribution.publication.id} with put-code #{contribution.orcid_put_code}")
       true
-    rescue Orcid::PubMapper::PubMapperError => e
-      logger&.info("#{self.class} - author #{author.id} - did not add publication #{contribution.publication.id}: #{e.message}")
+    rescue Orcid::PubMapper::PubMapperError, Orcid::Client::InvalidTokenError => e
+      logger&.warn("#{self.class} - author #{author.id} - did not add publication #{contribution.publication.id}: #{e.message}")
       false
     rescue StandardError => e
       logger&.error("#{self.class} - author #{author.id} - error publication #{contribution.publication.id}: #{e.message}")
--- a/lib/orcid/client.rb
+++ b/lib/orcid/client.rb
@@ -5,6 +5,8 @@ require 'oauth2'
 module Orcid
   # Client for ORCID.org API.
   class Client
+    class InvalidTokenError < StandardError; end
+
     # Fetch the works for a researcher.
     # Model for the response: https://pub.orcid.org/v3.0/#!/Development_Public_API_v3.0/viewWorksv3
     # @param [string] ORCID ID for the researcher
@@ -68,6 +70,8 @@ module Orcid
       case response.status
       when 201
         response['Location'].match(%r{work/(\d+)})[1]
+      when 401
+        raise InvalidTokenError, "Invalid token for #{orcidid} - ORCID.org API returned #{response.status} (#{response.body}) for: #{work.to_json}"
       when 409
         match = response.body.match(/put-code (\d+)\./)
         raise 'ORCID.org API returned a 409, but could not find put-code' unless match
--- a/lib/orcid/client.rb
+++ b/lib/orcid/client.rb
@@ -71,7 +71,7 @@ module Orcid
       when 201
         response['Location'].match(%r{work/(\d+)})[1]
       when 401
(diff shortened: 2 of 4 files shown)


lib/orcid/add_works.rb
```diff
@@ -42,8 +42,8 @@ def add_work(author, contribution, orcid_user)
       contribution.save!
       logger&.info("#{self.class} - author #{author.id} - added publication #{contribution.publication.id} with put-code #{contribution.orcid_put_code}")
       true
-    rescue Orcid::PubMapper::PubMapperError => e
-      logger&.info("#{self.class} - author #{author.id} - did not add publication #{contribution.publication.id}: #{e.message}")
+    rescue Orcid::PubMapper::PubMapperError, Orcid::Client::InvalidTokenError => e
+      logger&.warn("#{self.class} - author #{author.id} - did not add publication #{contribution.publication.id}: #{e.message}")
```
peetucket, line 46: 1. catch this new exception as one that will not get an HB notification (we will throw the new exception below when necessary)
2. also, seems to make more sense to log this as a warn instead of an info to make it easier to grep this out later?

lib/orcid/client.rb
```diff
@@ -68,6 +70,9 @@ def add_work(orcidid, work, token)
       case response.status
       when 201
         response['Location'].match(%r{work/(\d+)})[1]
+      when 401
+        raise InvalidTokenError, "Invalid token for #{orcidid} - ORCID.org API returned #{response.status} (#{response.body}) for: #{work.to_json}"
+        false
```
peetucket, line 75: ORCID throws a 401 when the token is invalid

spec/lib/orcid/add_works_spec.rb
```diff
@@ -3,17 +3,25 @@
...
                                 '91gd29cb-124e-5bf8-1ard-90315b03ae12')
   end
 
+  # This is an author that has no currently valid publications that can be pushed to ORCID
+  let(:orcid_user_no_orcid_approved_contributions) do
+    Mais::Client::OrcidUser.new(author_no_orcid_approved_contributions.sunetid, author_no_orcid_approved_contributions.orcidid,
+                                ['/read-limited', '/activities/update', '/person/update'], '91gd29cb-124e-5bf8-1ard-90315b03ae12')
+  end
+
   describe '#add_for_orcid_user' do
```
peetucket, line 19: I discovered some of the existing tests were passing because the tokens were invalid and that was allowing them to pass.  This new data setup ensures they pass for the reasons the reasons stated in the test

spec/lib/orcid/add_works_spec.rb
```diff
@@ -64,12 +72,26 @@
...
+        expect(NotificationManager).not_to receive(:error)
+        error_message = "Orcid::AddWorks - author #{author.id}"\
+          " - did not add publication #{publication.id}: Invalid token for #{author.orcidid}"\
+          ' - ORCID.org API returned 401 '\
+          "({\n  \"error\" : \"invalid_token\",\n  \"error_description\" : \"Invalid access token: 91gd29cb-124e-5bf8-1ard-90315b03ae12\"\n})"
+        expect(logger).to receive(:warn).with(error_message)
+        expect(contribution_count).to be_zero
+      end
+    end
+
```
peetucket, line 87: the new test verifying invalid tokens are logged but not sent to HB (in this case via the NotificationManger class)

jmartin-sul (approved):

PR #1411, opened 2021-11-09
do not map non-numeric page numbers from web of science

peetucket:
## Why was this change made?

Sometimes WOS records include non-numeric page numbers in the source data (e.g. "E23").  This (pending verification if we really want to do it) will not map these non-numeric pages when creating citations.

Updated test

### DIFF
--- a/lib/web_of_science/map_citation.rb
+++ b/lib/web_of_science/map_citation.rb
@@ -55,6 +55,11 @@ module WebOfScience
 
       fst = page['begin'].to_s.strip
       lst = page['end'].to_s.strip
+
+      # ignore anything that doesn't look like a number
+      fst = '' unless fst.to_i.to_s == fst
+      lst = '' unless lst.to_i.to_s == lst
+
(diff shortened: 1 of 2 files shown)


justinlittman:
Roman numerals?

peetucket:
> Roman numerals?

good point, this would discard them, though the example i saw reported by a faculty member was "e2001741", which is neither.  perhaps if we want to go down this route, we'd be better served by a clever regex that could pick up both roman numerals and arabic numerals

justinlittman:
This seems like a good idea until we get complaints from another researcher that we're stripping page numbers in a format that we weren't expecting. And then this seems like a bad idea.

peetucket:
di

> This seems like a good idea until we get complaints from another researcher that we're stripping page numbers in a format that we weren't expecting. And then this seems like a bad idea.

Indeed.  There may be cases where we are stripping page numbers that look odd to us but make sense in the context of the data.

peetucket:
Unlcear improvement.

PR #1424, opened 2022-01-06
update to ruby 3.0.3

peetucket:
## Why was this change made?

Update to ruby 3.0.3

~~Note: I had to update the mysql2 gem locally to a version that the gemfile says is not currently supported on the VM, else I got errors like this:~~

~~`expected ActiveRecord::RecordNotUnique, got #<ActiveRecord::StatementInvalid: TypeError: no implicit conversion of Hash into String>`~~

~~We need to see if mysql2 0.5.3 will work on the VMs.~~ seems to install fine on both ruby 2 and ruby 3 on the vms

Locally.  Testing on CI now

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,8 +9,8 @@ gem 'sass-rails', '~> 5.0'
 # Use Uglifier as compressor for JavaScript assets
 gem 'uglifier', '~> 4.1'
 
-# pinning myql2 since 0.5.3 is currently not compatible with our version of CentOS and Ruby (as of Apr 2020)
-gem 'mysql2', '~> 0.5', '< 0.5.3'
+# mysql 0.5.3 is required for ruby 3 and is supported on latest OS in use: Oracle Linux (as of Jan 2022)
+gem 'mysql2', '>= 0.5.3'
 
 gem 'nokogiri', '>= 1.7.1'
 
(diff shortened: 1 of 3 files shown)


Gemfile
```diff
@@ -9,8 +9,9 @@ gem 'sass-rails', '~> 5.0'
 # Use Uglifier as compressor for JavaScript assets
 gem 'uglifier', '~> 4.1'
 
-# pinning myql2 since 0.5.3 is currently not compatible with our version of CentOS and Ruby (as of Apr 2020)
-gem 'mysql2', '~> 0.5', '< 0.5.3'
+# mysql2 0.5.3 is not compatible with older versions of CentOS and Ruby (as of Apr 2020)
```
jcoyne, line 12: We're not on centos anymore are we? Can we drop this comment?

peetucket: dropped comment

peetucket:
Appears to work fine on sul-pub-cap-dev-a and sul-pub-stage, which have now both been updated to ruby 3 techstack as well.

peetucket:
Note: tests pass on both ruby 2.7 and ruby 3 with these changes (confirmed locally)

jmartin-sul (approved):

PR #1426, opened 2022-01-11
remove spring; remove delayed_job; adjust autoloading

peetucket:
## Why was this change made?

Getting ready for rails 7 upgrade:

- remove spring gem (not needed anymore)
- remove delayed_job (not compatible with rails 7, only used in one spot in the job that is almost never used, and not part of the regular workflow)
- setup autoloading for the new rails 7 default

Existing  tests


config/initializers/okcomputer.rb
```diff
@@ -1,4 +1,10 @@
 require 'okcomputer'
+require 'cap/client'
+require 'pubmed'
+require 'mais'
+require 'orcid'
+require 'web_of_science'
+require 'clarivate/links_client'
```
peetucket, line 8: gets rid of deprecation warnings, which will come an error in rails 7:

```
DEPRECATION WARNING: Initialization autoloaded the constants Cap, Cap::Client, Pubmed, Mais, Orcid, WebOfScience, Clarivate, and Clarivate::LinksClient.

Being able to do this is deprecated. Autoloading during initialization is going
to be an error condition in future versions of Rails.
```

spec/lib/smci_report_spec.rb
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-describe SMCIReport do
+describe SmciReport do
   before { allow(Logger).to receive(:new).and_return(null_logger) }
```
peetucket, line 4: zeitwerk identified this class name as not matching the filename, fixed it

lib/smci_report.rb
```diff
@@ -53,7 +53,7 @@
 
 require 'csv'
 
-class SMCIReport
+class SmciReport
```
peetucket, line 56: zeitwerk identified this class name as not matching the filename, fixed it

jcoyne (approved):

PR #1427, opened 2022-01-11
update to rails 7

peetucket:
## Why was this change made?

Update to rails 7.

Note: we use a gem called `paper_trail` which is currently not determined to be compatible with rails 7.  There is an ongoing discussion here: https://github.com/paper-trail-gem/paper_trail/pull/1356

While it is not blocking the gem from being installed, it spits out a warning.

I recommend we wait a bit to see how this shakes out.  Currently paper_trail is only used on destroy for three models, and I am quite certain we don't delete rows from any of those tables as part of normal operations, so we could also consider removing the gem in the future if it won't be upgraded to work with Rails 7.

Existing tests


peetucket:
Already upgraded to rails 7

PR #1442, opened 2022-02-18
Parse dates from Pubmed records into pub_hash

### LINKED ISSUE #1441: Map dates for Pubmed harvested records in pub hash
For WoS and ORCID, we currently map the date into the pub_hash, but we don't do this Pubmed soured records.  

See https://github.com/sul-dlss/sul_pub/blob/main/app/models/pubmed_source_record.rb#L109 for where the Pubmed records are mapped to pub hash.

See https://github.com/sul-dlss/sul_pub/blob/main/lib/web_of_science/map_citation.rb#L28 for where the date is mapped in WoS records

See https://github.com/sul-dlss/sul_pub/blob/main/lib/orcid/work_mapper.rb#L31 for where the date is mapped in ORCID records

peetucket:
## Why was this change made?

Fixes #1441 (upcoming request from Profiles):  if available, we should map dates from PubmedSource Records into the same YYYY-MM-DD format as we do for WoS and ORCID records

Notes:
- refactored how the year (currently done in our code) is extracted so we can use the same type of code for the month/year
- could add some more Pubmed fixtures to test some more permutations of date locations/formats to be parsed
- once done, we will need to go back and rebuild the pubhash for Pubmed sourced records if we want to the date added

There is a rake task for rebuilding Pubmed pub hashes from a previous remediation, but it starts with cap_profile_ids: see https://github.com/sul-dlss/sul_pub/blob/main/lib/tasks/pubmed.rake#L7  In this case, we may want to do *all* Pubmed sourced publications (or all retrieved after a certain date), so we'd have a much bigger set to start from.  Basically all you have to do is call `publication#rebuild_pub_hash` on any publication that has a Pubmed provenance to remap.  See https://github.com/sul-dlss/sul_pub/issues/1441#issuecomment-1054842850 for how to do this (tested in cap-dev)

Note that rebuilding pubhashes will touch the "last_updated" field in the publication record, which will then cause the publication to be returned to Profiles as updated in their nightly update API call.  So after we do the remediation (if we do it), they will end up with a huge batch of updated publications (when they are used to a much smaller number).  Edit: there will be ~17910 (see https://github.com/sul-dlss/sul_pub/issues/1441#issuecomment-1054842850)

See https://github.com/sul-dlss/sul_pub/wiki/Useful-console-commands-(harvest,-WoS,-Pubmed,-others)#provenance-publication-stats for some ideas on how to query for the Pubmed provenance publications.

- Added some new tests
- Deployed to cap-dev, which has ~16k publications that are Pubmed provenance.  Added the date to all of these and then verified that the date's year matched the year using https://github.com/sul-dlss/sul_pub/issues/1441#issuecomment-1054842850

### DIFF
--- a/app/models/pubmed_source_record.rb
+++ b/app/models/pubmed_source_record.rb
@@ -99,6 +99,58 @@ class PubmedSourceRecord < ApplicationRecord
     mesh_headings_for_record
   end
 
+  def extract_year_from_pubmed_record(publication)
+    year = nil
+
+    # look for a year in all of the xpath locations in order
+    #  stop after the first produces something that looks like a year
+    pubmed_date_xpaths('Year').each do |path|
+      match = publication.xpath(path).text.match(/[12][0-9]{3}/)
+      next unless match
+
+      year = match.to_s
+      break
+    end
+
+    year
+  end
+
+  def extract_month_from_pubmed_record(publication)
+    month = nil
+
+    # look for a month in all of the xpath locations in order
+    #  stop after the first produces something that looks like a month (12 or Aug)
+    pubmed_date_xpaths('Month').each do |path|
+      month_value = publication.xpath(path).text
+      match = month_value.match(/\A[012]?[0-9]{1}\z/) # e.g. 01, 11, 06, 6, 12
+      name_match = month_value.match(/\A[a-zA-Z]{3}\z/) # e.g. Aug, aug, Oct
+      match = Date::ABBR_MONTHNAMES.index(name_match.to_s) if name_match # convert month name to number
+      next unless match
+
+      month = match.to_s
+      break
+    end
+
+    month
+  end
+
+  def extract_day_from_pubmed_record(publication)
+    day = nil
+
+    # look for a day in all of the xpath locations in order
+    #  stop after the first produces something that looks like a month (12 or Aug)
+    pubmed_date_xpaths('Day').each do |path|
+      day_value = publication.xpath(path).text
+      match = day_value.match(/\A[0123]?[0-9]{1}\z/) # e.g. 01, 11, 06, 6, 12, 23, 30
+      next unless match
+
+      day = match.to_s
+      break
+    end
+
+    day
+  end
+
   # Convert MEDLINE®PubMed® XML to pub_hash
   # @param [Nokogiri::XML::Node] publication
   # @return [Hash<Symbol => Object>] pub_hash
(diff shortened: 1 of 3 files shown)


peetucket:
See https://github.com/sul-dlss/sul_pub/issues/1441#issuecomment-1054842850 for an example of how to remediate the previous harvest pubmed source dates

app/models/pubmed_source_record.rb
```diff
@@ -99,6 +99,61 @@ def extract_mesh_headings_from_pubmed_record(pubmed_record)
...
+    pubmed_date_xpaths('Year').each do |path|
+      match = publication.xpath(path).text.match(/[12][0-9]{3}/)
+      next unless match
+
+      year = match.to_s
+      break
+    end
+
+    year
+  end
```
mjgiarlo, line 117: @peetucket I wonder if it's worth simplifying this method (and its buddies below) to something more like:

```suggestion
potential_years.find { |year| year.match?(/[12][0-9]{3}/) }.to_s
  # see https://dtd.nlm.nih.gov/ncbi/pubmed/doc/out/180101/el-Year.html
  def extract_year_from_pubmed_record(publication)
    pubmed_date_xpaths('Year').find { |path| publication.xpath(path).text.match?(/[12][0-9]{3}/) }.to_s.presence
  end
```

peetucket: Thanks for the suggestion.  I tweaked slightly to make it work (this wasn't quite working as expected), and applied a similar change to the other day and month methods.

app/models/pubmed_source_record.rb
```diff
@@ -139,22 +171,16 @@ def source_as_hash(publication = Nokogiri::XML(source_data).xpath('//PubmedArtic
...
+    year = extract_year_from_pubmed_record(publication)
+    record_as_hash[:year] = year if year
+
+    month = extract_month_from_pubmed_record(publication)
+    day = extract_day_from_pubmed_record(publication)
+    if year && month && day
+      record_as_hash[:date] =  "#{year}-#{month.rjust(2, '0')}-#{day.rjust(2, '0')}"
+    elsif year && month
+      record_as_hash[:date] =  "#{year}-#{month.rjust(2, '0')}"
     end
```
peetucket, line 184: note, we want to 0 pad months and days to be consistent with other date formats, so 2022-1-4 becomes 2022-01-04

mjgiarlo (approved):

peetucket:
@justinlittman : just pinging you in case you wanted to have a look and this before I merge ... since it was pretty straightforward, I think I'm going to move on to refactoring this a bit more to pull out the Pubmed pubhash generation into it's own class like we do for WoS and ORCID (instead of having it in the model "source record" class)

justinlittman (approved):

PR #1443, opened 2022-02-21
Update dependencies

sul-devops-team:


lib/clarivate/links_client.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'faraday/httpclient'
+
```
ndushay, line 4: specs passed without this, presumably due to some rails or app magic, but I included it to be clearer (?)

lib/mais/client.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'faraday/httpclient'
+require 'faraday/retry'
```
ndushay, line 4: specs passed without this, presumably due to some rails or app magic, but I included it to be clearer (?)

lib/orcid/client.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'faraday/httpclient'
+require 'faraday/retry'
```
ndushay, line 4: specs passed without this, presumably due to some rails or app magic, but I included it to be clearer (?)

lib/pubmed/client.rb
```diff
@@ -1,5 +1,8 @@
 # frozen_string_literal: true
 
+require 'faraday/httpclient'
+require 'faraday/retry'
```
ndushay, line 4: specs passed without this, presumably due to some rails or app magic, but I included it to be clearer (?)

ndushay (approved):
Approved by automated merge script

PR #1447, opened 2022-02-21
remove unnec requires from a spec

ndushay:
## Why was this change made?

cruft is confusing

test suite


spec/lib/web_of_science/wsdl.rb
```diff
@@ -1,8 +1,5 @@
 # frozen_string_literal: true
 
-require 'faraday'
-require 'webmock'
-
```
ndushay, line 5: the tests should load whatever the app loads, by default.

jcoyne (approved):

PR #1452, opened 2022-03-08
temporarily disable MaIS update job during maintenance

peetucket:
## Why was this change made?

MaIS is performing maintenance the weekend of March 18 - March 21.  We should disable the MaIS job during this window to avoid errors by deploying this branch on March 18 and then letting main get redeployed on the 21st.  We can then delete this branch.


peetucket:
No longer needed -- MaIS schedule can be re-enabled.

PR #1455, opened 2022-03-11
updates to harvest from orcid

peetucket:
## Why was this change made?

When we turn on the code to harvest works from ORCID, we may need to add an additional field to the pub hash to show the original source of the work that came from ORCID (e.g. it came to ORCID via CrossRef, we record "CrossRef").  This data is useful for Profiles to show to users. 

This is not a hard requirement yet, just did some research and got started on the work.
Would still need to add the `work_source` field to the mapped pub_hash (in a new field presumably).


lib/orcid/work_mapper.rb
```diff
@@ -38,7 +38,8 @@ def map
         conference: map_conference,
         series: map_series,
         pages: work.pages,
-        publisher: work.publisher
+        publisher: work.publisher,
+        orcid_work_source: work.work_source
       }.compact
```
peetucket, line 43: this is the new pub_hash field we are mapping to

lib/orcid/work_record.rb
```diff
@@ -14,6 +14,11 @@ def put_code
       @put_code ||= work_response['put-code']
     end
 
+    # the name of the source that provided the work to ORCID (e.g. "Crossref")
+    def work_source
+      @work_source ||= work_response.dig('source', 'source-name', 'value')
+    end
+
     def work_type
```
peetucket, line 22: this is the work response from ORCID we pull the new value from

PR #1457, opened 2022-03-11
[HOLD] Add provenance column to publication table

### LINKED ISSUE #652: Make provenance a publication level AR field and enforce its presence

peetucket:
SOMETHING GOT MESSED UP ... EXTRA COMMITS.  I will fix and force push.

## Why was this change made?

Fixes #652 - this will make it easier to run some analysis queries about where publications originally came from without resorting to identifiers or inspecting the pub_hash.  Note that this change will need a migration to the database, adding a column and an index to a big table, so may want to do off hours in case it takes a while.  

Hold for now so we can plan this for a good time.

we will then want to run the rake task to back-fill data

Localhost

### DIFF
--- a/app/models/author.rb
+++ b/app/models/author.rb
@@ -48,6 +48,25 @@ class Author < ApplicationRecord
     Settings.HARVESTER.INSTITUTION.name
   end
 
+  # rubocop:disable Metrics/AbcSize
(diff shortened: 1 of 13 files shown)


peetucket:
Closed in favor of #1458

PR #1458, opened 2022-03-11
add provenance convenience column to pub table table

peetucket:


app/models/publication.rb
```diff
@@ -270,7 +272,9 @@ def authoritative_doi_source?
 
   private
 
-  # @return [String] might be empty, won't be nil
+  # @return [String] might be empty, won't be nil, normalize since we have some older data in varying cases
+  # @note obscures ActiveRecord field/attribute getter for provenance, once we are sure we have backfilled all previous
+  #  records with the rake data:add_provenance rake task, we can get rid of this method
```
jmartin-sul, line 277: might be good to file an issue for that follow-on work and turn this `@note` to a `@todo` with a link to it?

jmartin-sul (approved):
one nitpick, but wouldn't hold up merging over it

app/models/publication.rb
peetucket: good idea, added the note ... will keep this on HOLD for now until ready to run the migration

jmartin-sul: thanks!

PR #1460, opened 2022-03-15
[WIP] Refactor pubmed pubhash

peetucket:
## Why was this change made?

WIP ... needs to have tests updated and rebased. 

Move pubmed generation for Pubmed sourced publications into it's own class (as it is for WoS and ORCID).  Mostly a straight move of all the methods from the source record class to a new class to separate the concerns.

Just create a single class for now to do the whole mapping.

Updated tests.


peetucket:
closed in favor of #1463

PR #1461, opened 2022-03-15
replace mysql2 gem with sqllite3 in dev/test

peetucket:
## Why was this change made?

We do not need mysql installed locally just to run in development/test mode.  Also, installing the mysql2 gem can be a challenge to install on a laptop due to dependencies.  This moves the mysql2 gem so it is only used in production, and allows sqlite3 to be used for dev/test.  

Since mysql defaults to case insensitive text searches, and sqllite3 does not, I made one small change to the code to ensure the tests continue to run successfully with sqlite3.

All environments deploy in production mode, so will continue to use mysql.

Test suite, and tried deploying to cap-dev to ensure it still deploys cleanly.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,9 +9,6 @@ gem 'sass-rails', '~> 5.0'
 # Use Uglifier as compressor for JavaScript assets
 gem 'uglifier', '~> 4.1'
 
-# mysql 0.5.3 is required for ruby 3 and is supported on latest OS in use: Oracle Linux (as of Jan 2022)
-gem 'mysql2', '>= 0.5.3'
-
 gem 'nokogiri', '>= 1.7.1'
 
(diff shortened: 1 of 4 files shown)


peetucket:
> Looks good. Should this PR also touch `config/database.yml`?

Yes, because in local/dev, we want to use sqlite3 and not mysql.  When deployed, the database.yml file is replaced with the version from puppet (which continues to use mysql).

mjgiarlo:
@peetucket 💬 
> > Looks good. Should this PR also touch `config/database.yml`?
> 
> Yes, because in local/dev, we want to use sqlite3 and not mysql. When deployed, the database.yml file is replaced with the version from puppet (which continues to use mysql).

Oy. I didn't scroll down far enough to see you'd already done it. 💫

PR #1463, opened 2022-03-15
Refactor pubmed pubhash generation to it's own class

peetucket:
## Why was this change made?

This moves the pubhash generation for Pubmed sourced publications into it's own class (as it is for Web of Science and ORCID). This is mostly a straight move of all the methods from the PubmedSoruceRecord database class to a new class to separate the concerns.  The existing method in the source record now just classes our new class to generate the pub hash.

~~Note: this will need to be adjusted to match #1470 after it's merged~~

- Existing and updates tests.
- Deployed so sul-pub-cap-dev and then confirmed with 300 Pubmed sourced publications on cap-dev that the current hash and the hash from the refactoring match.


app/models/pubmed_source_record.rb
```diff
@@ -79,258 +79,8 @@ def pubmed_update
     update! attrs
   end
 
-  def extract_abstract_from_pubmed_record(pubmed_record)
-    pubmed_record.xpath('MedlineCitation/Article/Abstract/AbstractText').text
+  def source_as_hash
+    Pubmed::MapPubHash.map(source_data)
   end
```
peetucket, line 85: remove all of the pub hash generation related methods to a new class, and call our new class when we ask for the pub hash

app/models/pubmed_source_record.rb
```diff
@@ -79,258 +79,8 @@ def pubmed_update
...
-
-  # Convert MEDLINE®PubMed® XML to pub_hash
-  # @param [Nokogiri::XML::Node] publication
-  # @return [Hash<Symbol => Object>] pub_hash
-  # @see https://www.nlm.nih.gov/bsd/licensee/elements_descriptions.html XML Element Descriptions and their Attributes
-  # rubocop:disable Metrics/AbcSize
-  # rubocop:disable Metrics/CyclomaticComplexity
-  # rubocop:disable Metrics/PerceivedComplexity
-  def source_as_hash(publication = Nokogiri::XML(source_data).xpath('//PubmedArticle'))
-    record_as_hash = {}
```
peetucket, line 142: We remove all of the pub hash generation related methods to a new class, and replace this `source_as_hash` method with a simple call to our new class when we ask for the pub hash (as shown above)

app/models/pubmed_source_record.rb
```diff
@@ -79,258 +79,7 @@ def pubmed_update
...
-      "MedlineCitation/Article/ArticleDate/#{date_part}",
-      "PubmedData/History/PubMedPubDate[@PubStatus='accepted']/#{date_part}",
-      "PubmedData/History/PubMedPubDate[@PubStatus='pubmed']/#{date_part}",
-      "PubmedData/History/PubMedPubDate[@PubStatus='medline']/#{date_part}",
-      "PubmedData/History/PubMedPubDate[@PubStatus='entrez']/#{date_part}"
-    ]
+  def source_as_hash
+    Pubmed::MapPubHash.map(source_data)
   end
 end
```
peetucket, line 85: the old `source_as_hash` method now just uses our new class to do the work

lib/pubmed/fetcher.rb
```diff
@@ -31,7 +31,7 @@ def self.fetch_remote_pubmed(pmid)
       # PubMed, oddly enough, the last resort
       pm_xml = Pubmed.client.fetch_records_for_pmid_list(pmid)
       Nokogiri::XML(pm_xml).xpath('//PubmedArticle').map do |doc|
-        add_citation(PubmedSourceRecord.new.source_as_hash(doc))
+        add_citation(Pubmed::MapPubHash.map(doc.to_xml))
       end
```
peetucket, line 35: no need to create a new source record in this case just to do a mapping ... since we already have the XML, just call the new class directly

lib/pubmed/map_pub_hash.rb
```diff
@@ -0,0 +1,270 @@
+# frozen_string_literal: true
+
+# rubocop:disable Metrics/ClassLength
+module Pubmed
+  # Map Pubmed source record data into the SUL PubHash data
+  class MapPubHash
+    attr_reader :pubmed_article
```
peetucket, line 7: this is the new class that gets all of the Pubmed pub hash related methods ported over (in the Pubmed namespace)

spec/lib/pubmed/map_pub_hash_spec.rb
```diff
@@ -0,0 +1,237 @@
+# frozen_string_literal: true
```
peetucket, line 1: this gets the pub hash related specs from the source record ported over, along with some refactoring

spec/models/pubmed_source_record_spec.rb
```diff
@@ -264,37 +50,33 @@ def check_author_hash(author)
...
-        it 'works when ArticleId is missing' do
-          expect(doi(26_858_277)).to include(id: '10.1136/bmj.i493')
-        end
-      end
+  describe '#source_as_hash' do
+    it 'returns a pub_hash' do
+      record = described_class.send(:get_pubmed_record_from_pubmed, pmid_created_1999)
+      expect(record.source_as_hash).to eq(
+        { provenance: 'pubmed', pmid: '10000166',
+          title: 'Fluid permeability in porous media: Comparison of electrical estimates with hydrodynamical calculations.',
```
peetucket, line 58: this verifies that we can still call `source_as_hash` on the source record and get the pub hash back, as this method is used throughout the code

spec/models/pubmed_source_record_spec.rb
```diff
@@ -3,121 +3,7 @@
 describe PubmedSourceRecord, :vcr do
   let(:pmid_created_1999) { 10_000_166 }
 
-  def author_doc(xml)
```
peetucket, line 6: the tests related to generating a pub_hash were moved (and refactored a bit) into the new spec file above

jcoyne (approved):

PR #1465, opened 2022-03-15
adjust db migrations to work with sqlite3

peetucket:
## Why was this change made?

If you run migrations on a brand new blank database, you will hit a couple failures.  One seems to be change in syntax, the other uses a datatype that is only applicable to MySQL.  These changes are needed to get them to run correctly and work with sqlite3.  They also work with mysql (I tested locally).  

Also, I added in some comments into the database.yml file in case you want to use mysql locally.

Running migrations from a blank database on sqlite3 and mysql (on laptop)

### DIFF
--- a/db/migrate/20160412224434_create_author_identities.rb
+++ b/db/migrate/20160412224434_create_author_identities.rb
@@ -1,10 +1,9 @@
 class CreateAuthorIdentities < ActiveRecord::Migration[4.2]
   def change
     # On MySQL, foreign keys require the InnoDB storage engine, and we need to ensure UTF8 charsets for strings
-    opts = {}
-    opts[:options] = 'ENGINE=InnoDB DEFAULT CHARSET=utf8' if ActiveRecord::Base.connection.adapter_name.downcase.starts_with? 'mysql'
+    options = ActiveRecord::Base.connection.adapter_name.downcase.starts_with?('mysql') ? 'ENGINE=InnoDB DEFAULT CHARSET=utf8' : ''
 
-    create_table :author_identities, opts do |t|
+    create_table :author_identities, options: options do |t|
       t.belongs_to :author,       index: true, foreign_key: true, null: false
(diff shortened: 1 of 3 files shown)


db/migrate/20160412224434_create_author_identities.rb
```diff
@@ -1,10 +1,9 @@
 class CreateAuthorIdentities < ActiveRecord::Migration[4.2]
   def change
     # On MySQL, foreign keys require the InnoDB storage engine, and we need to ensure UTF8 charsets for strings
-    opts = {}
-    opts[:options] = 'ENGINE=InnoDB DEFAULT CHARSET=utf8' if ActiveRecord::Base.connection.adapter_name.downcase.starts_with? 'mysql'
+    options = ActiveRecord::Base.connection.adapter_name.downcase.starts_with?('mysql') ? 'ENGINE=InnoDB DEFAULT CHARSET=utf8' : ''
 
-    create_table :author_identities, opts do |t|
+    create_table :author_identities, options: options do |t|
       t.belongs_to :author,       index: true, foreign_key: true, null: false
```
peetucket, line 7: this syntax changed at some point and no longer worked, this fixes it

db/migrate/20210520185112_create_orcid_source_records.rb
```diff
@@ -1,7 +1,8 @@
 class CreateOrcidSourceRecords < ActiveRecord::Migration[6.0]
+  # `:text` column with 16MB is the same as MySQL's MEDIUMTEXT but works on sqlite3
   def change
     create_table :orcid_source_records do |t|
-      t.mediumtext :source_data
+      t.text :source_data, limit: 16_777_215
       t.bigint :last_modified_date
```
peetucket, line 6: see https://github.com/sul-dlss/sul_pub/blob/main/db/migrate/20160405194617_switch_to_medium_text.rb for a similar choice in a different migration

jcoyne (approved):

Issue #1469, opened 2022-03-19
Fix pubmed date parsing


peetucket:
The work in #1442 didn't account for the fact that days and months need to come from the same part of the XML structure.  By getting them separately, it is possible you can pull the day and month from two different parts of the XML, resulting in the wrong date.  This is because we end up looking at multiple parts of the XML for year, month and day in a certain order (see https://github.com/sul-dlss/sul_pub/blob/main/app/models/pubmed_source_record.rb#L326-L335), but the month can be present in one part of the XML and not in the others.

For example: for PMID:11096574, we are pulling the "Feb" from 

```
                    <PubDate>
                        <Year>1999</Year>
                        <Month>Feb</Month>
                    </PubDate>
```

but since there is no day listed, we then end up getting the day from:

```
            <PubMedPubDate PubStatus="pubmed">
                <Year>2000</Year>
                <Month>11</Month>
                <Day>30</Day>
            </PubMedPubDate>
```

This results in a date of `02-30-1999`, which is invalid and wrong.

We should instead look for both the day and month at the same time and stop when we have at least one of them.  If we don't get a day, we leave it blank (verified with the Profiles team this is ok).

On the rails console
```
pmsr = PubmedSourceRecord.find_by(pmid:'11096574')
publication = Nokogiri::XML(pmsr.source_data).xpath('//PubmedArticle')
pm
pmsr.send(:extract_year_from_pubmed_record,publication)
=> "1999"
pmsr.send(:extract_month_from_pubmed_record,publication)
=> "2"
pmsr.send(:extract_day_from_pubmed_record,publication)
=> "30"
pmsr.source_as_hash[:date]
=> "1999-02-30" #WRONG
```

with new code, we don't have day or month anymore separately, so:
```
pmsr.send(:extract_month_day_from_pubmed_record,publication)
=> "02" # no day
pmsr.source_as_hash[:date]
=> "1999-02" # ok
```

PR #1470, opened 2022-03-19
[HOLD] pick month and day from same part of xml

peetucket:
## Why was this change made?

Fix for #1469 - when parsing dates from Pubmed, we need to be sure to pick the month and day from the same part of the XML structure, still going in the preferred order.  We still return the first match, even if it's only the month (and even if a month-day combo is found later in one of less preferred locations).  This is because the places to look in the XML (the xpaths) are set in order of "best date to use".  

The problem was discovered when the Profiles team updated all of their publications on -dev (where I had back-filled with the current date mapping code) and started hitting exceptions on publications with bad dates (e.g. PMID:11096574, SulPubId:8787 for example, where we had sent a date of 2019-02-30).

This makes the needed changes and re-enables the date parsing for Pubmed records that was disabled in #1468 

Hold:
~~1. to verify this is the correct approach with the Profiles team~~ Verified
~~2. to see if when there is only a month and no day if we should leave the day off (as in this PR) or just set the day to some value (like 01 or 15)~~  Ok to send just month and year
3. until testing of parsing of dates again on -dev

Testing suggested:
- deploy this branch to -dev
- re-run all of the date parsing for previously harvested Pubmed applications (see https://github.com/sul-dlss/sul_pub/issues/1441#issuecomment-1054842850 for how to do this)
- have Profiles ran the publication update on their end, and see if any new errors occur

if successful, we can merge, deploy everywhere and then back-port dates on other environments as well

Note: these changes were already ported into #1463 , the refactoring of pubmed pub hash generation

Existing tests, and added a new test that confirms we will not grab a day from a different part of the XML if we find a valid month only in an earlier part.

### DIFF
--- a/app/models/pubmed_source_record.rb
+++ b/app/models/pubmed_source_record.rb
@@ -102,33 +102,34 @@ class PubmedSourceRecord < ApplicationRecord
   # see https://dtd.nlm.nih.gov/ncbi/pubmed/doc/out/180101/el-Year.html
   def extract_year_from_pubmed_record(publication)
     # look for a year in all of the xpath locations in order, pick the first that produces something that looks like a four digit year
-    pubmed_date_xpaths('Year')
-      .map { |path| publication.xpath(path).text.match(/[12][0-9]{3}/).to_s }
+    pubmed_date_xpaths
+      .map { |path| publication.xpath("#{path}/Year").text.match(/[12][0-9]{3}/).to_s }
       .compact_blank.first
   end
 
   # see https://dtd.nlm.nih.gov/ncbi/pubmed/doc/out/180101/el-Month.html
-  def extract_month_from_pubmed_record(publication)
+  # see https://dtd.nlm.nih.gov/ncbi/pubmed/doc/out/180101/el-Day.html
+  def extract_month_day_from_pubmed_record(publication)
     # look for a month in all of the xpath locations in order
     #  pick the first that produces something that looks like a month (12 or Aug)
-    pubmed_date_xpaths('Month').map do |path|
-      month_value = publication.xpath(path).text
-      match = month_value.match(/\A[012]?[0-9]{1}\z/) # e.g. 01, 11, 06, 6, 12
+    # then look for a day in the same place as the month ... since they must go together
+    #  we will return just a month or a month and a day, going in the order set in the xpaths
+    result = pubmed_date_xpaths.map do |path|
+      month_value = publication.xpath("#{path}/Month").text
+      month_match = month_value.match(/\A[012]?[0-9]{1}\z/) # e.g. 01, 11, 06, 6, 12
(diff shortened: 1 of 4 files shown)


app/models/pubmed_source_record.rb
```diff
@@ -175,13 +176,8 @@ def source_as_hash(publication = Nokogiri::XML(source_data).xpath('//PubmedArtic
...
 
-    month = extract_month_from_pubmed_record(publication)
-    day = extract_day_from_pubmed_record(publication)
-    if year && month && day
-      record_as_hash[:date] =  "#{year}-#{month.rjust(2, '0')}-#{day.rjust(2, '0')}"
-    elsif year && month
-      record_as_hash[:date] =  "#{year}-#{month.rjust(2, '0')}"
-    end
+    month_day = extract_month_day_from_pubmed_record(publication)
+    record_as_hash[:date] = "#{year}-#{month_day}" if year && month_day
```
peetucket, line 181: combine the day and month methods into one method, so this part of the code gets simpler

app/models/pubmed_source_record.rb
```diff
@@ -108,27 +108,28 @@ def extract_year_from_pubmed_record(publication)
   end
 
   # see https://dtd.nlm.nih.gov/ncbi/pubmed/doc/out/180101/el-Month.html
-  def extract_month_from_pubmed_record(publication)
+  # see https://dtd.nlm.nih.gov/ncbi/pubmed/doc/out/180101/el-Day.html
+  def extract_month_day_from_pubmed_record(publication)
     # look for a month in all of the xpath locations in order
```
peetucket, line 113: we now have a single method to return both a month and day together, instead of two separate methods

app/models/pubmed_source_record.rb
```diff
@@ -323,14 +319,14 @@ def author_to_hash(author)
...
-      "PubmedData/History/PubMedPubDate[@PubStatus='pubmed']/#{date_part}",
-      "PubmedData/History/PubMedPubDate[@PubStatus='medline']/#{date_part}",
-      "PubmedData/History/PubMedPubDate[@PubStatus='entrez']/#{date_part}"
+      'MedlineCitation/Article/Journal/JournalIssue/PubDate',
+      'MedlineCitation/Article/ArticleDate',
+      "PubmedData/History/PubMedPubDate[@PubStatus='accepted']",
+      "PubmedData/History/PubMedPubDate[@PubStatus='pubmed']",
+      "PubmedData/History/PubMedPubDate[@PubStatus='medline']",
+      "PubmedData/History/PubMedPubDate[@PubStatus='entrez']"
     ]
```
peetucket, line 330: this method now makes more sense without passing in a parameter, since we now get the month and day together in one pass through this array and it makes it easier to do this way

justinlittman (approved):

Issue #1477, opened 2022-04-01
NoMethodError: undefined method `id' for nil:NilClass


lwrubel:
## Backtrace

    line 49 of [PROJECT_ROOT]/lib/orcid/add_works.rb: rescue in add_work
    line 39 of [PROJECT_ROOT]/lib/orcid/add_works.rb: add_work
    line 30 of [PROJECT_ROOT]/lib/orcid/add_works.rb: block in add_for_orcid_user

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/50046/faults/84764481)

lwrubel:
This shows up in `production.log` as:
```
I, [2022-04-01T04:45:47.102448 #24716]  INFO -- :   Rendered requests/clarivate_links.xml.erb (Duration: 0.2ms | Allocations: 87)
I, [2022-04-01T04:46:06.794364 #24716]  INFO -- :   Rendered requests/clarivate_links.xml.erb (Duration: 0.2ms | Allocations: 48)
W, [2022-04-01T04:46:11.021243 #24716]  WARN -- : nil publication assignment for author id 133895
W, [2022-04-01T04:46:11.145485 #24716]  WARN -- : nil publication assignment for author id 133895
```

peetucket:
There appear to be some contributions that are unlinked from Publications, which causes a problem when we try to push them.  This is not a bug in this part of the code, it is a data problem of unknown origin.

Looking at the HB alert, we can find the implicated author and then examine the contributions, where we find the contributions with no publication_id:

```
author=Author.find(65205)
author.contributions.where(publication_id: nil) 
```

There are six of these, all created on May 21, 2021.  Having a nil publication_id should not be possible.

peetucket:
It appears there are 22 total in the database, all from May 25-26, 2021.  So that's good that it appears to be limited, though I am not sure how it happened.  They are associated with 5 different authors.

```
Contribution.where(publication_id: nil).size
=> 22
Contribution.where(publication_id: nil).collect {|c| c.author.cap_profile_id}.uniq
=> [82582, 238607, 227623, 4313, 61123]
[23] pry(main)> Contribution.where(publication_id: nil).collect {|c| c.created_at.to_date}.uniq
=> [Tue, 25 May 2021, Wed, 26 May 2021]
```

peetucket:
[...]

PR #1482, opened 2022-04-04
prevent bad data from being created in contributions

peetucket:
## Why was this change made?

To deal with data issues highlighted in #1477 

Just add a database constraint to prevent nulls from being created (it will throw an exception if we try to do this).

~~HOLD Until the existing nulls are dealt with in the production database.~~ Data deleted in #1477

~~Not sure we really need to do this, since we really should never have a nil publication_id in a contribution (though we somehow did in this case) and it is bad data.   Perhaps it's better to know by getting an exception so we can cleanup the bad data?~~

Existing tests

### DIFF
--- a/db/migrate/20220405041748_contributions_not_null.rb
+++ b/db/migrate/20220405041748_contributions_not_null.rb
@@ -1,0 +1,6 @@
+class ContributionsNotNull < ActiveRecord::Migration[6.1]
+  def change
+    change_column_null(:contributions, :author_id, false)
+    change_column_null(:contributions, :publication_id, false)
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2022_01_11_185021) do
+ActiveRecord::Schema.define(version: 2022_04_05_041748) do
 
   create_table "author_identities", id: :integer, charset: "utf8", collation: "utf8_unicode_ci", force: :cascade do |t|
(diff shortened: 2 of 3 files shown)


jcoyne (commented):
Have you considered adding a database constraint instead?

peetucket:
> Have you considered adding a database constraint instead?

Yeah, probably better.  The ActiveRecord model should guard against this with a validation but it somehow slipped through.  Odd it was all in one day (and only in production, there are no bad records in QC).

db/schema.rb
```diff
@@ -80,9 +80,9 @@
...
 
   create_table "contributions", id: :integer, charset: "utf8", collation: "utf8_unicode_ci", force: :cascade do |t|
-    t.integer "author_id"
+    t.integer "author_id", null: false
     t.integer "cap_profile_id"
-    t.integer "publication_id"
+    t.integer "publication_id", null: false
     t.string "status"
     t.boolean "featured"
     t.string "visibility"
```
peetucket, line 88: ran migrations using local mysql and this is the file that got re-created

db/migrate/20220405041748_contributions_not_null.rb
```diff
@@ -0,0 +1,6 @@
+class ContributionsNotNull < ActiveRecord::Migration[6.1]
+  def change
+    change_column_null(:contributions, :author_id, false)
+    change_column_null(:contributions, :publication_id, false)
+  end
```
peetucket, line 5: we will allow cap_profile_id to continue to be null since some older records have it as null and the rest of the app seems to work fine with it being null as well

jcoyne (approved):

PR #1483, opened 2022-04-05
Run bin/rails db:migrate

jcoyne:
## Why was this change made?

I believe this different because we changed the testing db to sqlite3 (https://github.com/sul-dlss/sul_pub/pull/1465)

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,15 +10,15 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2022_01_11_185021) do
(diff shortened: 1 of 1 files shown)


peetucket (changes requested):
This has the migration from #1482 which we do not want in main yet.  Also, do we care that schema.rb is not SQlite instead of MySQL?  I don't think so, but want to verify.  If not, I can do this in that PR instead (in fact I manually handcrafted changes to the schema.rb to avoid the format change from Mysql to SQlite in that PR).

peetucket:
Not needed anymore now that #1482 is merged

PR #1484, opened 2022-04-06
Create a critical checks endpoint for okcomputer

mjgiarlo:
## Why was this change made?

This gives us an okcomputer endpoint to hit that includes all checks we might want to alert about, not the ones that trigger partner email. Helps with Nagios integration.

### /status/all screenshot

Screenshot from 2022-04-06 16-36-39

### /status/critical screenshot

Screenshot from 2022-04-06 16-36-57

Tested in cap-dev env.

### DIFF
--- a/config/initializers/okcomputer.rb
+++ b/config/initializers/okcomputer.rb
@@ -9,6 +9,13 @@ require 'clarivate/links_client'
 OkComputer.mount_at = 'status' # use /status or /status/all or /status/<name-of-check>
 OkComputer.check_in_parallel = true
 
+# Place critical checks in their own collection so they may be checked together without the external/optional checks below
+critical_checks = OkComputer::CheckCollection.new('Critical Checks')
+
+# Mark the automatically registered checks as critical ones
+critical_checks.register 'default', OkComputer::Registry.fetch('default')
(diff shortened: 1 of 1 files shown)


ndushay:
could you put in a screen shot of the okcomputer page with and without this?

ndushay:
it kind of feels like we're just reinventing "checks that should alert if they ever aren't passing" and "optional" ... I'm not sure what we gain from this.  Am I missing something?

mjgiarlo:
@ndushay 💬 
> could you put in a screen shot of the okcomputer page with and without this?

Yep! ☝🏻

mjgiarlo:
@ndushay 💬 
> it kind of feels like we're just reinventing "checks that should alert if they ever aren't passing" and "optional" ... I'm not sure what we gain from this. Am I missing something?

Puppet currently configures Nagios to hit `/status/default` instead of `/status/all` for most of our apps, but that gets us a basic up check and nothing else. And we can't just switch over to `/status/all` across the board in prod, because for some apps, `/status/all` contains checks that are not pager-worthy. This is a pattern to create a new `/status/critical` endpoint for these apps in prod so that we get more value out of Nagios checks without burdening Ops.

ndushay (approved):

PR #1495, opened 2022-04-29
Catch and retry WoS queries if session expires

peetucket:
## Why was this change made?

To better deal with this HB error: https://app.honeybadger.io/projects/50046/faults/79809428

Tested on qa.  Seems fine

### DIFF
--- a/lib/web_of_science/client.rb
+++ b/lib/web_of_science/client.rb
@@ -55,6 +55,15 @@ module WebOfScience
         log_level: @log_level,
         pretty_print_xml: true
       )
+    rescue Savon::SOAPFault => e
+      # this may occur if the session identifier timed out and we didn't catch it, this allows us to request
+      # a new session identifier and try the API call again
+      # see https://github.com/sul-dlss/sul_pub/wiki/Web-of-Sciences-Expanded-API-Throttle-Limits for known API limits
+      # the number of queries per session should already be accounted for, but not the time out limit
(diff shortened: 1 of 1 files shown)


peetucket:
@ndushay : fyi, this is now on qa (https://sul-pub-cap-uat.stanford.edu/status/all), looks ok for now, let's leave through next week and just watch to be sure we don't see any HB alerts

peetucket:
Have not seen any unusual errors in QA with this branch.   Merging.

lib/web_of_science/client.rb
```diff
@@ -55,6 +55,15 @@ def search
...
       )
+    rescue Savon::SOAPFault => e
+      # this may occur if the session identifier timed out and we didn't catch it, this allows us to request
+      # a new session identifier and try the API call again
+      # see https://github.com/sul-dlss/sul_pub/wiki/Web-of-Sciences-Expanded-API-Throttle-Limits for known API limits
+      # the number of queries per session should already be accounted for, but not the time out limit
+      raise e unless e.message.include? 'There is a problem with your session identifier (SID)'
+
+      session_close
+      retry
```
edsu, line 66: I didn't know that `retry` could be used outside of a loop like this! Would it be good to do some kind of exponential back off because of:

> 30 authentication requests in 5 minutes

Or does a subsequent request tend to work ok?

PR #1505, opened 2022-05-16
add ability to export specific date ranges in report

peetucket:
## Why was this change made?

Update a sul-pub report to allow for date ranges in query exports (instead of just a "date after").

See #1542 for the same changes to a different report.

UAT


peetucket:
Already done in #1562

Issue #1512, opened 2022-06-06
Null for middle names in publications


peetucket:
I have a ServiceNow ticket with a publication issue reported by Professor Kacper Rogala.  The issue was that the auto-imported version of his publication:
 
“Structural basis for the docking of mTORC1 on the lysosomal surface”
 
Is displaying the citation with ‘n.’ showing for each person’s middle name initial who did not provide a middle name, as seen here:
 
Rogala, K. B., Gu, X. n., Kedir, J. F., Abu-Remaileh, M. n., Bianchi, L. F., Bottino, A. M., Dueholm, R. n., Niehaus, A. n., Overwijn, D. n., Fils, A. P., Zhou, S. X., Leary, D. n., Laqtom, N. N., Brignole, E. J., Sabatini, D. M.
2019; 366 (6464): 468–75
 
In looking in the database at the Bib JSON stored, I could see that for those who did not have a middle name provided, that the word ‘null’ was in the field and thus what is being displayed.  Can you confirm if this issue is with the source or on the SUL side and how we can get the issue corrected?
 
In the interim, Professor Rogala had entered a manual version of the publication with the corrected author names, but could not enter the Abstract, as we have not ever exposed this field to the end users.  However, since we have this field in the database and I could see from the auto-imported publication version that this was not a restricted abstract, I added it to his manually entered publication.  I advised him to go ahead and hide the auto-imported version for now until it can be corrected.
 
Please let me know your feedback on this and potentially how many publications could be impacted by this.  The SUL Pub ID for the auto-imported publication shown above with the middle name issues is 776543.

peetucket:
So I had a quick look and here is what I found:

-	This publication was first harvested in early 2020.  While I did see the “null” for middle names in the specific author block of the JSON in our database, I didn’t see the the “n.” in the three types of formatted citations we are returning in the BibJson, so it’s not clear to me where/why that “n.” is getting added.
-	I believe we made some changes in our code since then, with does not add nulls, and verified this, because when I rebuilt the cached JSON for this publication, the “nulls” went away.  If you were to re-pull that publication, they should be gone now from the author block.  This means that any new publications should not be adding them.  This particular publication happened to be originally harvested in March 2020.


Either way, it would be interested to know how citations are being formatted/shown on the Profiles websites – I was under the impression it was just showing whatever string we returned in the JSON, but there might be something more complicated going on in Profiles.

peetucket:
[...]

PR #1513, opened 2022-06-06
better way to verify publication type

peetucket:
## Why was this change made?

While investigating #1512, noticed that the way we check publication types before rebuilding pub hashes could be done more appropriately by checking for the provenance of the publication (through an existing helper method) instead of by checking for the existence of an identifier.

Existing tests


peetucket:
Never mind, a test failed and it's worth untangling if this change actually breaks something real or not.

app/models/publication.rb
```diff
@@ -149,22 +149,24 @@ def update_from_pubmed
...
       sw_source_record = SciencewireSourceRecord.find_by_sciencewire_id(sciencewire_id)
       build_from_sciencewire_hash(sw_source_record.source_as_hash)
-    elsif pmid
+    elsif pubmed_pub?
       pubmed_source_record = PubmedSourceRecord.find_by_pmid(pmid)
       self.pub_hash = pubmed_source_record.source_as_hash
     end
-    sync_publication_hash_and_db
+    pubhash_needs_update!
     save
```
peetucket, line 170: calling `pubhash_needs_update!` is better, since this forces a call to `sync_publication_hash_and_db` and then denormalizes some data into the database too on save

app/models/publication.rb
```diff
@@ -149,22 +149,24 @@ def update_from_pubmed
...
   def rebuild_pub_hash
     raise 'rebuilding WOS records is unsupported, unimplemented' if wos_pub?
 
-    if sciencewire_id
+    if sciencewire_pub?
       sw_source_record = SciencewireSourceRecord.find_by_sciencewire_id(sciencewire_id)
       build_from_sciencewire_hash(sw_source_record.source_as_hash)
-    elsif pmid
+    elsif pubmed_pub?
       pubmed_source_record = PubmedSourceRecord.find_by_pmid(pmid)
```
peetucket, line 166: using these helper methods check the publication provenance instead of just looking for the presence of an identifier, which is safer

app/models/publication.rb
```diff
@@ -144,7 +144,7 @@ def sync_publication_hash_and_db
   end
 
   def update_from_pubmed
-    return false if pmid.blank?
+    return false if pmid.blank? || !pubmed_pub?
```
peetucket, line 148: we should really only update from pubmed if it's actually a pubmed publication

spec/models/publication_spec.rb
```diff
@@ -473,6 +481,11 @@
       pub = described_class.new(pub_hash: { provenance: 'wos' })
       expect { pub.rebuild_pub_hash }.to raise_error(RuntimeError)
     end
+
+    it 'raises for cap record' do
+      pub = described_class.new(pub_hash: { provenance: 'cap' })
+      expect { pub.rebuild_pub_hash }.to raise_error(RuntimeError)
+    end
   end
```
peetucket, line 489: some new behavior, we should also raise an error for other types that don't have rebuilding the pubhash supported

jcoyne:
@peetucket can you rebase this please?

aaron-collier (approved):

PR #1524, opened 2022-06-30
allow rebuilding of pub_hash for wos records

peetucket:
## Why was this change made?

Rebuilding the pub_hash from source data is sometimes helpful if we need to do some data correction after harvesting has been done.  This is not something that is done in the production code automatically, but we have rake tasks that allow this to happen.  Use cases are (1) we discover a bug in our code that maps source data to the pub_hash, and we need to remediate previously harvested data, (2) a single record is updated at the source and we'd like to update our mapped data.

Up until now, Web of Science records cannot be rebuilt in the code as easily as other source records, even though it is absolutely possible.  

1. This adds the ability to do this by calling the same method that we can call for Pubmed or Sciencewire source records.
2. It uses a better check of the record provenance instead of using identifiers to decide if the record can be rebuilt or not.
3. It raises an exception if you try to rebuild a manually entered publication

The methods changed are not called in running production code, they are called in rake tasks and sometimes manually on the rails console for data remediation.

Updated a couple tests.

### DIFF
--- a/app/models/publication.rb
+++ b/app/models/publication.rb
@@ -144,7 +144,7 @@ class Publication < ApplicationRecord
   end
 
   def update_from_pubmed
-    return false if pmid.blank?
+    return false if pmid.blank? || !pubmed_pub?
 
     pm_source_record = PubmedSourceRecord.find_by_pmid(pmid)
     return false unless pm_source_record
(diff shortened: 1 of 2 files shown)


app/models/publication.rb
```diff
@@ -144,7 +144,7 @@ def sync_publication_hash_and_db
   end
 
   def update_from_pubmed
-    return false if pmid.blank?
+    return false if pmid.blank? || !pubmed_pub?
```
peetucket, line 148: just being very cautions ... we really should not be updating from pubmed source unless this was originally sourced from pubmed (this verifies this a pubmed sourced record)

app/models/publication.rb
```diff
@@ -155,14 +155,18 @@ def update_from_pubmed
 
   # @return [Boolean] true if .save is successful
   def rebuild_pub_hash
-    raise 'rebuilding WOS records is unsupported, unimplemented' if wos_pub?
+    raise 'rebuilding non-harvested records is unsupported, unimplemented' unless harvested_pub?
```
peetucket, line 159: wos records are not supported, but manual records are definitely not supported (this adds a check for that)

app/models/publication.rb
```diff
@@ -155,14 +155,18 @@ def update_from_pubmed
...
 
-    if sciencewire_id
+    if sciencewire_pub?
       sw_source_record = SciencewireSourceRecord.find_by_sciencewire_id(sciencewire_id)
       build_from_sciencewire_hash(sw_source_record.source_as_hash)
-    elsif pmid
+    elsif pubmed_pub?
       pubmed_source_record = PubmedSourceRecord.find_by_pmid(pmid)
```
peetucket, line 164: use provenance instead of identifier to check the type of record we have

jcoyne (approved):

PR #1542, opened 2022-08-19
add ability to search for date ranges in smci report

peetucket:
## Why was this change made?

The SMCI report benefits from exporting specific date ranges (instead of just after a specific date)

See also #1505 for the same changes to a different report.

UAT

### DIFF
--- a/lib/smci_report.rb
+++ b/lib/smci_report.rb
@@ -37,6 +37,9 @@
 # The 'date_since' parameter is optional and specifies the date after which publications for profiles authors will be exported
 #  The format for 'date_since'  is DD/MM/YYYY.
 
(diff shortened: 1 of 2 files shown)


lib/smci_report.rb
```diff
@@ -81,7 +86,7 @@ def run
     start_time = Time.zone.now
     logger.info '*****************'
     logger.info 'Starting export.'
-    logger.info "Exporting all publications for #{total_authors} authors to #{@output_file}. Since date: '#{@date_since}'.  " \
+    logger.info "Exporting all publications for #{total_authors} authors to #{@output_file}. Date rnage: '#{@date_since}' to '#{@date_to}'.  " \
```
justinlittman, line 89: Typo.

justinlittman (approved):
Approve pending fixing typo.

Issue #1546, opened 2022-09-07
Ignore pubmed queries for names that produce bad queries


peetucket:
If the author's name contains certain boolean operators, Pubmed will incorrectly parse the query and return too many results.  (e.g. "Or" is a male Hebrew name)

Results of investigation, showing basically all of this user's publications from Pubmed:

```
author=Author.find_by(sunetid:'orzolty')
author.publications.size
=> 1151
author_query = Pubmed::QueryAuthor.new(author)
author_query.send(:term)
=> "((Zolty, Or[Author])) AND (Stanford University[Affiliation])"
pmids_from_query = query.pmids;
pmids_from_query.size
=> 1153
```

If you go the Pubmed web interface, you will see how Pubmed is parsing this query:

https://www.ncbi.nlm.nih.gov/pmc/?term=((Zolty%2C+Or%5BAuthor%5D))+AND+(Stanford+University%5BAffiliation%5D)

Bottom line is that Pubmed is turning it into this: `(Zolty[All Fields] OR Author[All Fields]) AND Stanford University[Affiliation]` which is not correct.

We looked at the Pubmed API documentation and tried different variations of our query and couldn't produce the intended results.

We also looked at the WoS query, which does seem to produce the expected results and does not appear to have an issue, so relying on WoS seems sufficient for these edge case users.

Actions:
- [x] In the pubmed query class, we should indicate the query is invalid if any of the author's names contains a configured list of disallowed terms ("OR", "AND", "NOT"): https://github.com/sul-dlss/sul_pub/blob/main/lib/pubmed/query_author.rb#L22-L24
- [x] Double-check that the author now returns invalid 

```
author=Author.find_by(sunetid:'orzolty')
query_author = Pubmed::QueryAuthor.new(author)
query_author.valid? # should return false instead of true
```
- [x] check to see if we should set all of the 'new' status publications for this author to 'denied', or if the Profiles team (in the #sul-cap-colab channel) wants to do this on their end

Something like:
```
author.contributions.where(status:'new').each {|contrib| contrib.update(status: 'denied')}
```

- [x] Communicate with Profiles team about the issue, which is that Pubmed is incorrectly parsing queries when an author's name is "Or", "And", "Not" and we can't seem to figure out how to get their query syntax to understand.  Our recommendation and course of action is to **not** query Pubmed for these users, and to rely on the Web of Science only, which does seem to work fine.

- [x] Send a support request to NCI/Pubmed to ask about how to query correctly.

[...]

PR #1548, opened 2022-09-07
look for invalid names for pubmed queries

peetucket:
## Why was this change made?

Code changes for #1546 

- if an author has a name that includes "Or", "And", or "Not" (case insensitive, either first or last), that name will be ignored when constructing the query ... this means a user could in theory have a second "alternate identity" name that is valid, and we would still use it
- if there are no valid names available, the entire query will be marked as invalid and thus not run, producing no results
- while I was making this change, I extended our existing check for invalid names (currently only first name must have at least one character) to ensure the last name also has to have at least one valid character (eg. a last name of blank or null will also be marked as an invalid name)
- reorganize spec to put all of the tests related to the `valid?` method in a describe block

Added new tests

### DIFF
--- a/spec/lib/pubmed/query_author_spec.rb
+++ b/spec/lib/pubmed/query_author_spec.rb
@@ -2,14 +2,7 @@
 
 describe Pubmed::QueryAuthor do
   let(:query_author) { described_class.new(author, options) }
-  let(:query_period_author) { described_class.new(period_author, options) }
-  let(:query_space_author) { described_class.new(space_author, options) }
-  let(:query_blank_author) { described_class.new(blank_author, options) }
-
   let(:author) { create :russ_altman }
-  let(:space_author) { create :author, :space_first_name }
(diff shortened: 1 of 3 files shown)


spec/lib/pubmed/query_author_spec.rb
```diff
@@ -195,4 +167,70 @@
...
+        expect(query_or_author.pmids).to be_empty
+      end
+
+      it 'indicates that a name with "Not" as a last name is not a valid query and returns no pmids' do
+        expect(query_not_author).not_to be_valid
+        expect(query_not_author.pmids).to be_empty
+      end
+    end
+  end
 end
```
peetucket, line 236: no tests were eliminated here, just moved to a describe block, along with adding new tests verifying that names that include "Or" and "Not" are ignored, as well as nil last names

ndushay (approved):

PR #1562, opened 2022-11-14
changes to publication export reports

peetucket:
## Why was this change made?

1. We sometimes need to export list of publications in specific time ranges (instead of just "since" a specific date until today).  This adds that capability to two of the rake tasks that export publications.
2. The rake task that exports publication for all authors and all statuses was changed to export some more metadata (such as identifiers and publication status) to make it more useful during analysis.

Ran on UAT

Updated this wiki page: https://github.com/sul-dlss/sul_pub/wiki/ORCID-and-Publication-Reports

### DIFF
--- a/lib/tasks/sul.rake
+++ b/lib/tasks/sul.rake
@@ -35,9 +35,9 @@ namespace :sul do
   # Produce statistics about the number of publications imported, number of unique authors, and numbers in each state
   #  in the specified time period
   # bundle exec rake sul:publication_import_stats['1/1/2022','1/31/2022']
-  task :publication_import_stats, %i[date1 date2] => :environment do |_t, args|
-    start_date = Date.strptime(args[:date1], '%m/%d/%Y')
-    end_date = Date.strptime(args[:date2], '%m/%d/%Y')
+  task :publication_import_stats, %i[start_date end_date] => :environment do |_t, args|
+    start_date = Date.strptime(args[:start_date], '%m/%d/%Y')
+    end_date = Date.strptime(args[:end_date], '%m/%d/%Y')
     new_author_count = Author.where('created_at > ? and created_at < ?', start_date, end_date).where(active_in_cap: true, cap_import_enabled: true).count
     total_publication_count = Contribution.where('created_at > ? and created_at < ?', start_date, end_date).count
     total_unique_publication_count = Contribution.select(:author_id).where('created_at > ? and created_at < ?', start_date, end_date).distinct.count
@@ -61,10 +61,10 @@ namespace :sul do
   desc 'Export publications as csv'
   # bundle exec rake sul:export_pubs_csv['/tmp/output_file.csv','01/01/2013','01/01/2014']
   # parameters are output csv file, start date and end date in format of mm/dd/yyyy
-  task :export_pubs_csv, %i[output_file date1 date2] => :environment do |_t, args|
+  task :export_pubs_csv, %i[output_file start_date end_date] => :environment do |_t, args|
     output_file = args[:output_file]
-    start_date = Date.strptime(args[:date1], '%m/%d/%Y')
-    end_date = Date.strptime(args[:date2], '%m/%d/%Y')
+    start_date = Date.strptime(args[:start_date], '%m/%d/%Y')
+    end_date = Date.strptime(args[:end_date], '%m/%d/%Y')
     raise 'missing required params' unless output_file && start_date && end_date
 
     total_pubs = Contribution.where('created_at > ? and created_at < ?', start_date, end_date).count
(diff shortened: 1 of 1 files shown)


lib/tasks/sul.rake
```diff
@@ -59,27 +59,39 @@ namespace :sul do
...
       csv << header_row
-      Contribution.where('created_at > ?', Date.strptime(date_since, '%m/%d/%Y')).find_each do |contribution|
+      Contribution.where('created_at > ? and created_at < ?', start_date, end_date).find_each do |contribution|
+        puts "#{n} of #{total_pubs} : pub_id #{contribution.publication_id}, author_id #{contribution.author_id}"
         pub = contribution.publication
         author = contribution.author
-        csv << [pub.title, author.last_name, author.first_name, author.sunetid, author.university_id,
-                contribution.created_at, pub.pub_hash[:apa_citation]]
+        doi = pub.pub_hash[:identifier].map { |ident| ident[:id] if ident[:type].downcase == 'doi' }.compact.join
+        pmid = pub.pub_hash[:identifier].map { |ident| ident[:id] if ident[:type].downcase == 'pmid' }.compact.join
```
mjgiarlo, line 85: You could also consider using `.filter_map` here:

```suggestion
        doi = pub.pub_hash[:identifier].filter_map { |ident| ident[:id] if ident[:type].downcase == 'doi' }.join
        pmid = pub.pub_hash[:identifier].filter_map { |ident| ident[:id] if ident[:type].downcase == 'pmid' }.join
```

lib/tasks/sul.rake
```diff
@@ -139,20 +151,21 @@ namespace :sul do
   desc 'Export publications for specific authors as csv'
   # exports all publications for the given sunets in a 'new' or 'approved' state after the date specified
   # input csv file should have a column with a header of 'sunetid' containing the sunetid of interest
-  # bundle exec rake sul:export_pubs_for_authors_csv['/tmp/input_file.csv','/tmp/output_file.csv','01/01/2013']
-  # parameters are input csv file with sunets, output csv file, and date to go back to in format of mm/dd/yyyy
-  task :export_pubs_for_authors_csv, %i[input_file output_file date_since] => :environment do |_t, args|
+  # bundle exec rake sul:export_pubs_for_authors_csv['/tmp/input_file.csv','/tmp/output_file.csv','01/01/2013','01/01/2014']
+  # parameters are input csv file with sunets, output csv file, start date and end datein format of mm/dd/yyyy
```
mjgiarlo, line 155: typo: `datein`

lib/tasks/sul.rake
```diff
@@ -176,8 +189,8 @@ namespace :sul do
                           else
                             ''
                           end
-            sunet_list = pub.contributions.where("status in ('new','approved') and created_at > ?",
-                                                 Time.zone.parse(date_since)).map do |c|
+            sunet_list = pub.contributions.where("status in ('new','approved') and created_at > ? and created_at < ?",
```
mjgiarlo, line 192: Because this query is repeated a bunch of times, you could consider making this a defined scope on the model and reusing it. But I don't think it's a big deal.

mjgiarlo (commented):
Some questions and comments

lib/tasks/sul.rake
```diff
@@ -59,27 +59,39 @@ namespace :sul do
   end
 
   desc 'Export publications as csv'
-  # bundle exec rake sul:export_pubs_csv['/tmp/output_file.csv','01/01/2013'] # parameters are output csv file, and date to go back to in format of mm/dd/yyyy
-  task :export_pubs_csv, %i[output_file date_since] => :environment do |_t, args|
+  # bundle exec rake sul:export_pubs_csv['/tmp/output_file.csv','01/01/2013','01/01/2014']
+  # parameters are output csv file, start date and end date in format of mm/dd/yyyy
+  task :export_pubs_csv, %i[output_file date1 date2] => :environment do |_t, args|
```
mjgiarlo, line 64: I wonder if you think it's worth considering changing the `output_file`, `date1`, and `date2` argument names so that their use is clearer throughout the task definition, which may also obviate the need for the code comments?

lib/tasks/sul.rake
```diff
@@ -139,20 +151,21 @@ namespace :sul do
   desc 'Export publications for specific authors as csv'
   # exports all publications for the given sunets in a 'new' or 'approved' state after the date specified
   # input csv file should have a column with a header of 'sunetid' containing the sunetid of interest
-  # bundle exec rake sul:export_pubs_for_authors_csv['/tmp/input_file.csv','/tmp/output_file.csv','01/01/2013']
-  # parameters are input csv file with sunets, output csv file, and date to go back to in format of mm/dd/yyyy
-  task :export_pubs_for_authors_csv, %i[input_file output_file date_since] => :environment do |_t, args|
+  # bundle exec rake sul:export_pubs_for_authors_csv['/tmp/input_file.csv','/tmp/output_file.csv','01/01/2013','01/01/2014']
+  # parameters are input csv file with sunets, output csv file, start date and end datein format of mm/dd/yyyy
+  task :export_pubs_for_authors_csv, %i[input_file output_file date1 date2] => :environment do |_t, args|
```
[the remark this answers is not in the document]
peetucket: Changing all of the `date_since` and `date1` to `start_date` and the `date_to` and `date2` to `end_date`

lib/tasks/sul.rake
peetucket: Sure, I'm changing the `date1` and `date2` but not sure about `output_file` ...

lib/tasks/sul.rake
peetucket: good point, but a bit ambivalent, since i am pretty sure this query is only used for rake task reports and not in the core part of the app itself

lib/tasks/sul.rake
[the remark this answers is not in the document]
mjgiarlo: Awesome!

lib/tasks/sul.rake
mjgiarlo: :shipit:

PR #1580, opened 2023-02-02
add HB checkins for cron jobs

### LINKED ISSUE #1564: Use Honeybadger checkins to ensure crons are running
See https://docs.honeybadger.io/guides/check-ins/

If feasible, this approach would be a fine way to ensure our cron jobs are running. Note similar ticket in prescat: https://github.com/sul-dlss/preservation_catalog/issues/2055, in etds, in dor-services app, happy-heron ...

peetucket:
## Why was this change made?

Fixes #1564 

Note: I don't think we need check ins for any environment other than production and QA (there are also cron jobs that run in dev, but don't warrant a checkin)

Also need shared_configs PRs:
- https://github.com/sul-dlss/shared_configs/pull/1898 (prod)
- https://github.com/sul-dlss/shared_configs/pull/1899 (qa)

### DIFF
--- a/config/schedule.rb
+++ b/config/schedule.rb
@@ -1,5 +1,9 @@
 # Learn more: http://github.com/javan/whenever
-job_type :envcommand, 'cd :path && RAILS_ENV=:environment :bundle_command :task :output'
+
+# These define jobs that checkin with Honeybadger.
+# If changing the schedule of one of these jobs, also update at https://app.honeybadger.io/projects/77112/check_ins
+job_type :rake_hb,
+         'cd :path && :environment_variable=:environment bundle exec rake --silent ":task" :output && curl --silent https://api.honeybadger.io/v1/check_in/:check_in'
 
 def stagger(hour)
   ('%02d' % hour) + (':%02d' % rand(60)) # some minute between 00 and 59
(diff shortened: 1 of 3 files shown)


config/schedule.rb
```diff
@@ -1,5 +1,9 @@
 # Learn more: http://github.com/javan/whenever
-job_type :envcommand, 'cd :path && RAILS_ENV=:environment :bundle_command :task :output'
+
```
peetucket, line 2: seems to be unneeded, removing

mjgiarlo: Thanks!

config/schedule.rb
```diff
@@ -1,5 +1,9 @@
...
+# These define jobs that checkin with Honeybadger.
+# If changing the schedule of one of these jobs, also update at https://app.honeybadger.io/projects/77112/check_ins
+job_type :rake_rb,
```
mjgiarlo, line 5: I think this copypasta may be broken (because it bit us in 1-2 other spots already). The `job_type` is named `rake_rb` here and invoked below as `rake_hb`. Changing this line to `rake_hb` should fix things up.

config/schedule.rb
```diff
@@ -9,30 +13,36 @@ def stagger(hour)
 
 # bi-weekly harvest at 5pm in qa, on the 8th and 23rd of the month
 every "0 17 7,23 * *", roles: [:harvester_qa] do
-  rake 'harvest:all_authors_update'
+  set :check_in, Settings.honeybadger_checkins.harvest_all_authors
+  rake_hb 'harvest:all_authors_update'
 end
```
mjgiarlo, line 18: This task looks like it only runs on QA. If so, do we want it to check in?

config/schedule.rb
```diff
@@ -9,30 +13,36 @@ def stagger(hour)
...
 
 # poll cap for new authorship information nightly at 4am-ish in prod, qa and dev
 every 1.day, at: stagger(4), roles: [:harvester_dev, :harvester_qa, :harvester_prod] do
-  rake 'cap:poll[1]'
+  set :check_in, Settings.honeybadger_checkins.cap_poll
+  rake_hb 'cap:poll[1]'
 end
```
mjgiarlo, line 30: I guess it's no big deal if this task winds up trying to check in on dev and QA? It'll be a no-op?

config/schedule.rb
```diff
@@ -9,30 +13,36 @@ def stagger(hour)
...
 
 # poll mais for new ORCID information nightly at 5am-ish in prod, qa and dev
 every 1.day, at: stagger(5), roles: [:harvester_dev, :harvester_qa, :harvester_prod] do
-  rake 'mais:update_authors'
+  set :check_in, Settings.honeybadger_checkins.mais_update_authors
+  rake_hb 'mais:update_authors'
 end
```
mjgiarlo, line 36: Same question as above.

config/schedule.rb
```diff
@@ -9,30 +13,36 @@ def stagger(hour)
...
 
 # send publications to ORCID profiles for all authorized users at 8am-ish every 7 days in qa
 every 7.days, at: stagger(8), roles: [:harvester_qa] do
-  rake 'orcid:add_all_works'
+  set :check_in, Settings.honeybadger_checkins.orcid_all_all_works
+  rake_hb 'orcid:add_all_works'
 end
```
mjgiarlo, line 42: This task looks like it only runs on QA. If so, do we want it to check in?

config/schedule.rb
peetucket: Yes, I think it makes sense to have the check-ins for QA as well since the Profiles team can use that environment for integration testing (and this helps keep the data up to date).  

Note that I created a check in key for all QA environments and even though there is a separate line in the `schedule.rb` file, it is actually calling the exact same rake task (just on a different frequency), so I think this is ok (and ditto for the other rake tasks on QA).

dev is being skipped on purpose (no checkin neeed, it has no hb key)

config/schedule.rb
peetucket: oops, thanks, pulled from a PR in another project, I guess it got fixed in a later PR

mjgiarlo: I did the same thing!

config/schedule.rb
mjgiarlo: okie dokie, thanks

mjgiarlo (approved):

PR #1590, opened 2023-03-20
allow exported pubs to be filtered; export all authors

peetucket:
## Why was this change made?

In order to run some new reports for OA analysis, we need to:

(1) Be able to filter publications exported to only specific publication years
(2) Be able to filter exported publications to only include approved publications
(3) Be able to export publications for all authors

This PR:

(1) Adds a new rake task to dump all authors in the database to CSV.  This can then be used an an import to the existing rake task.
(2) Modifies the publication export rake task to allow filtering by publication year and publication status.

Ran on QA to verify it works as intended.

### DIFF
--- a/lib/tasks/sul.rake
+++ b/lib/tasks/sul.rake
@@ -175,6 +175,8 @@ namespace :sul do
     rows = CSV.parse(File.read(input_file), headers: true)
     total_authors = rows.size
     total_pubs = 0
+    num_not_found = 0
+    users_not_found = []
     start_time = Time.zone.now
     puts "Exporting all pubs for #{total_authors} authors to #{output_file} from #{start_date} to #{end_date}.  Started at #{start_time}."
     header_row = %w[pub_title pub_id pmid doi publisher journal mesh pub_year provenance pub_associated_author_last_name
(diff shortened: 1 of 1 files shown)


lib/tasks/sul.rake
```diff
@@ -470,7 +485,7 @@ namespace :sul do
     total_users = users.size
     puts "Number of active users with harvesting enabled: #{total_users}"
 
-    header_row = %w[orcidid name last_name first_name sunet cap_profile_id total_publications total_approved_publications total_new_publications
+    header_row = %w[orcidid name last_name first_name sunetid cap_profile_id total_publications total_approved_publications total_new_publications
                     total_denied_publications]
```
peetucket, line 489: we use `sunetid` as an existing import column name in the above rake task, just making this column consistent with the other export rake task

ndushay (approved):
I'm obviously late to the party here ... but I think this is a giant rake file and maybe could have the code elsewhere and tested, and the rake take a thin wrapper calling the code.

Not holding up this PR, but please consider starting that approach for the NEXT change.   I would think a 500 line rake file would suggest that.  :-)

peetucket:
Yes, understood.  This rake task is a bit out of control.  I think the next change is at least splitting it out into separate rake files that are more organized by function.  I believe most (if not all) in this file are just data dumping tasks for reporting purposes, so I'm not sure there is as much value to spending cycles writing tests though (i.e. they are currently run on demand for data dumping for analysis).

ndushay:
Perhaps make a ticket for this and let someone else do it?   Spread the sul-pub knowledge around a little?

> Yes, understood. This rake task is a bit out of control. I think the next change is at least splitting it out into separate rake files that are more organized by function. I believe most (if not all) in this file are just data dumping tasks for reporting purposes, so I'm not sure there is as much value to spending cycles writing tests though (i.e. they are currently run on demand for data dumping for analysis).

PR #1608, opened 2023-06-14
change the cap_profile_id for health check, add to config

peetucket:
## Why was this change made?

The CAP health check is currently failing because the hard-coded `cap_profile_id` we currently use is returning a 404 from the CAP API (may be because that person's profile is archived).  This changes the cap_profile_id to one that is confirmed working and puts in the config in case we need to change it again.

Confirm separately with Profiles team that the 404 is expected from their API for this hardcoded profile id (4176 == https://profiles.stanford.edu/intranet/richard-lewis)

Deploy to UAT to confirm

Added code comments.


lwrubel (approved):
Approved pending green

PR #1616, opened 2023-06-23
Add ORCID to rake task output

edsu:
## Why was this change made?

It's helpful to also have the ORCID in the `sul:export_pubs_csv` and `sul:export_pubs_for_authors_csv` rake task output , since it can be used later to match up the data with other datasets.

Running it in prod! (shhhh)


peetucket:
Looks good.  While here, do you want to make the same change to the `export_pubs_for_authors_csv` task (around lines 182 and 224)?

edsu:
@peetucket I added that other task too!

peetucket (approved):

PR #1625, opened 2023-07-13
Replaces orcid client with gem.

justinlittman:
closes #3241

## Why was this change made?
Remove redundant code.

Unit, QA

```
pub@sul-pub-cap-dev-a:~/sul-pub/current$ RAILS_ENV=production bin/rake orcid:add_all_works
Added 52 works from 12 ORCID users.
```


config/initializers/orcid_client.rb
```diff
@@ -0,0 +1,7 @@
+SulOrcidClient.configure(
+  client_id: Settings.ORCID.CLIENT_ID,
+  client_secret: Settings.ORCID.CLIENT_SECRET,
+  base_url: Settings.ORCID.BASE_URL,
+  base_public_url: Settings.ORCID.BASE_PUBLIC_URL,
+  base_auth_url: Settings.ORCID.BASE_AUTH_URL
```
mjgiarlo, line 6: We typically use all lowercase to name Settings variables, but I'm assuming this PR is piggybacking off of existing shared_configs (that is still uppercased)? In which case I leave it to you whether it's timely or not to make this tweak.

lib/orcid.rb
```diff
@@ -25,7 +25,7 @@ def self.base_orcidid(orcidid)
 
   # Fetch works for a known ID for which we expect to get back publications
   def self.working?
-    response = client.fetch_works(Settings.ORCID.orcidid_for_check)
+    response = client.fetch_works(orcidid: Settings.ORCID.orcidid_for_check)
```
mjgiarlo, line 28: Trivial: consider `orcid_id` ❓

config/initializers/orcid_client.rb
justinlittman: Yeah, this is following the existing conventions in sul_pub and using existing configuration.

lib/orcid.rb
mjgiarlo: Totally fine not to tangle this change with a large refactoring/renaming 👍🏻

mjgiarlo (approved):
Thumbs-up from me, but would like @peetucket to take a quick gander as well.

peetucket (approved):

PR #1654, opened 2023-10-18
strip MEDLINE prefix when adding publications via PMID

### LINKED ISSUE #1653: Publications added by authorship controller can sometimes include MEDLINE: prefix
Reported in #sul-cap-collab : 

```
When trying to add a publication by searching, using the 'Add New Publication' search by title and year option, after getting the results of the search, and clicking on '+Add' for some publications, we are seeing the error message:
"Unable to update this publication right now.  Please try again later."
For example, trying searching for "Trends in Maxillofacial Trauma During COVID-19 at a Level 1 Trauma Center"  in "2022", the search results return:
Trends in Maxillofacial Trauma During COVID-19 at a Level 1 Trauma Center ENT-EAR NOSE & THROAT JOURNAL
Longino, E. S., Landeen, K. C., Wessinger, B. C., Kimura, K. S., Davis, S. J., Shastri, K. S., Stephan, S. J., Patel, P. N., Yang, S. F.
2022
Then I select to add the publication to the profile and it gives the above noted error.
```

peetucket:
## Why was this change made?

Fixes #1653 - see the analysis in the issue description.  Basically the new WoS API is returning PMIDs in the record now, and the Profiles team is preferentially picking the PMID in the result and sending that when adding publications via the manual title search (I verified both of these).  The WOS records adds a MEDINE: prefix to PMIDs, but when we get a request to add a publication for a specific PMID, we go straight to the Pubmed API to look for that PMID, and to do this, we do NOT want the MEDLINE: prefix.  So we should strip it in this case for any incoming PMID we are sent.

cap-dev-a and new specs

### DIFF
--- a/app/controllers/authorships_controller.rb
+++ b/app/controllers/authorships_controller.rb
@@ -220,7 +220,7 @@ class AuthorshipsController < ApplicationController
   # @param [String] pmid PubMed ID
   # @return [Publication]
   def get_publication_via_pubmed!(pmid)
-    pub = Publication.find_or_create_by_pmid(pmid)
+    pub = Publication.find_or_create_by_pmid(pmid.delete_prefix('MEDLINE:'))
     unless pub
       log_and_error!("The PMID:#{pmid} was not found either locally or at PubMed.")
       false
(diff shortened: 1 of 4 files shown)


peetucket:
closed in favor of #1655

Issue #1642, opened 2023-09-27
WoS REST API causes problems with some author searches


peetucket:
The new WoS REST API seems to have issues with some authors, throwing errors.

https://app.honeybadger.io/projects/50046/faults/99684154/01HB323CCE7F43DYFQ5SBH0QJT?page=0

For example, this author fails to get results back with an XML error as you can see from the HB error above.

You can reproduce the problem without attempting to harvest by just querying for the UIDs

```
author=Author.find(157210)
options = {load_time_span: '3W', relDate: '21'}
author_query = WebOfScience::QueryAuthor.new(author, options)
puts "WOS (by name): #{author_query.name_query.send(:name_query)}"
uids = author_query.uids
```

This is likely due to some problematic responses from Clarivate and needs to be investigated with them.

peetucket:
1. Perhaps add some additional logging around the XML parser in https://github.com/sul-dlss/sul_pub/blob/main/lib/web_of_science/xml_parser.rb#L25-L31 so we can identify the exact WoS ID record that is not parsing correctly, and provide this to info to Clarivate.
2. Ignore records that do not parse instead of blowing up the whole harvest for that author (this would allow other publications to be added instead of stopping the whole process for that author).

peetucket:
I did a test with the author query below by requesting only the WOS UIDs in the first query (by asking for 0 results), taking the queryID and then requesting the UIDs separately via a separate call (using the Swigger API view), as described at the bottom of the page here: https://github.com/sul-dlss/sul_pub/wiki/Clarivate-APIs#web-of-sciences-expanded-api-notes 

This is is fast, and then iterating over resulting WOS UID and requesting one record at a time, and this parsed a few times successfully.  So I wonder if our current approach of requesting all of the records all at once has issues with very large publications resulting in giant amounts of XML.

Query that I ran:
```
AU=("Miller,D" OR "Miller,D,Craig" OR "Miller,D,C") AND AD=("stanford")
```

peetucket:
The last WOS UID has 2900 authors and is a giant XML record.  It works fine processed singly but is plausible causes issue when wrapped up with other publications:

```
wos_uid='WOS:001035243000001'
results = WebOfScience.queries.retrieve_by_id([wos_uid]).next_batch.to_a;
pub_hash = results[0].pub_hash;nil
puts pub_hash[:authors].size
=> 2900
puts results[0].to_xml
```

peetucket:
[...]

PR #1612, opened 2023-06-21
Replaces Clarivate SOAP API with REST API.

justinlittman:
## Why was this change made?
SOAP is so 1990s.

Fixes #1609

Requires https://github.com/sul-dlss/puppet/pull/9661/files

## How was this change tested?
Fingers crossed.

See #1618 for testing strategies

PR #1660, opened 2023-10-19
only fetch UIDs and not full records from WoS

peetucket:
## Why was this change made?

To start dealing with issues in #1642

Basically, it is very inefficient to request the entire record from WoS when we only need the WOS_UIDs to determine if we have new publications or not.  This alters the logic so we use a different endpoint provided by WoS when harvesting (because for each harvest, we really only need to request UIDs to verify if we have new publications to add or not).

This won't fix the problem in #1642 completely, since if you have new publications to add, you still need to fetch the full records (which will still blow up if there are many many authors for many publications in the set of 100 you can request at once), but not only should it greatly increase the performance of each harvest but should also cut down on these exceptions, since those will only occur if and when we need to fetch actual new publications (instead of happening on each harvest as they do now).

It also eliminates what I believe to be an unused class (`WebOfScience::RestRetriever`) and it moves it's spec to the `WebOfScience::BaseRestRetriever` class (which is very similar, and is subclassed and thus used).  While the `WebOfScience::RestRetriever` class was added in the PR that switched APIs #1612, it doesn't appear to be used or subclassed anywhere.

Finally,  I removed (and thus allowed to be recreated) all of the VCR cassettes for the WoS API to be sure they the tests would still pass given these code changes.  Note that a few expectations had to be updated  because of the new cassettes (i.e. publication counts, which increased since the last time the VCR cassettes were created).

You may ask...why is this is a new problem given the switch the REST based API.  And the answer is that it is not entirely a new problem.  The old SOAP based API has similar issues when requesting full publication records for publications with lots and lots of authors.  BUT: the older SOAP based API had an easier mechanism to request only UIDs on harvesting, and that's what we were doing.  The new REST based API makes a bit more work to replicate this functionality.  So this PR basically makes the new REST based API code work more like the older SOAP based API code worked.  Namely, it only requests UIDs on each harvest instead of full publication records, and then goes back to fetch the full records only for any publications we actually need to add to our database.  Which should reduce the occurrence of the problem.

Existing specs (allowing VCR cassettes to be re-created)


lib/web_of_science/base_rest_retriever.rb
```diff
@@ -6,7 +6,7 @@ module WebOfScience
   # - the "next_batch?" is like "next?"
   # - the "next_batch" is like "next"
   class BaseRestRetriever
-    attr_reader :records_found, :records_retrieved
+    attr_reader :records_found, :records_retrieved, :query_id
```
peetucket, line 10: we need the query_id to be available to our new `UidsQueryRestsRetriever` class

lib/web_of_science/base_rest_retriever.rb
```diff
@@ -18,23 +18,15 @@ def initialize(path, params, batch_size: MAX_RECORDS)
...
     end
 
-    # Retrieve and collect all record UIDs
-    # @return [Array<String>] WosUIDs
-    def merged_uids
-      uids = batch_one.uids
-      uids += next_batch.uids while next_batch?
-      uids
-    end
-
```
peetucket, line 28: this inefficient method is going away, to be supplanted by our new more efficient method in `UidsQueryRestsRetriever`

the only consumer of it will now use our new method

lib/web_of_science/base_rest_retriever.rb
```diff
@@ -45,13 +37,13 @@ def next_batch
...
-    attr_reader :batch_size, :query, :query_id, :path, :params
+    attr_reader :batch_size, :query, :path, :params
 
     delegate :client, to: WebOfScience
 
     # @return [Boolean] all records retrieved?
     def records_retrieved?
-      @batch_one.nil? ? false : records_retrieved == records_found
+      @batch_one.blank? ? false : records_retrieved == records_found
     end
```
peetucket, line 47: Switching the `nil?` to `blank?` since our new method deals with arrays, and we want the same logic that applies here to work for an empty array too

lib/web_of_science/base_rest_retriever.rb
```diff
@@ -45,13 +37,13 @@ def next_batch
     # this is the maximum number that can be returned in a single query by WoS
     MAX_RECORDS = 100
 
-    attr_reader :batch_size, :query, :query_id, :path, :params
+    attr_reader :batch_size, :query, :path, :params
```
peetucket, line 41: made this attr_reader public

lib/web_of_science/queries.rb
```diff
@@ -13,7 +13,15 @@ def retrieve_by_id(uids)
...
     # Convenience method, does the params_for_search expansion
-    # @param query [String] Query string like 'TS=particle swarm AND PY=(2007 OR 2008)'
+    # @param query_string [String] Query string like 'TS=particle swarm AND PY=(2007 OR 2008)'
+    # @return [WebOfScience::UidsQueryRestRetriever]
+    def user_query_uids(query_string, query_params: nil, batch_size: WebOfScience::BaseRestRetriever::MAX_RECORDS)
+      query = WebOfScience::UidsQueryRestRetriever::Query.new(user_query: query_string)
+      WebOfScience::UidsQueryRestRetriever.new(query, query_params:, batch_size:)
+    end
+
+    # Convenience method, does the params_for_search expansion
```
peetucket, line 23: our new method to query just for uids (we will use it from `QueryName` and `QueryOrcid`)

spec/lib/web_of_science/base_rest_retriever_spec.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
-describe WebOfScience::RestRetriever, :vcr do
-  let(:retriever) { described_class.new(params) }
+describe WebOfScience::BaseRestRetriever, :vcr do
+  let(:retriever) { described_class.new('/', params) }
```
peetucket, line 5: as far as I can tell, the `RestRetriever` class is not used anywhere and instead we use the `BaseRestRetriever` class ... just updating the existing spec for the unused class for the one we use

lib/web_of_science/query_name.rb
```diff
@@ -16,7 +16,7 @@ def initialize(author, options = {})
     def uids
       return [] unless valid?
 
-      queries.user_query(name_query, query_params:).merged_uids
+      queries.user_query_uids(name_query, query_params:).merged_uids
     end
```
peetucket, line 20: hits our new efficient method instead of the old one (which is now removed)

lib/web_of_science/query_orcid.rb
```diff
@@ -16,7 +16,7 @@ def initialize(author, options = {})
     def uids
       return [] unless valid?
 
-      queries.user_query(orcid_query, query_params:).merged_uids
+      queries.user_query_uids(orcid_query, query_params:).merged_uids
     end
```
peetucket, line 20: hits our new efficient method instead of the old one (which is now removed)

lib/web_of_science/rest_retriever.rb
peetucket, line 1: Class deleted, didn't seem to be used anywhere.

justinlittman (approved):

PR #1681, opened 2024-02-26
fix issue with pub export task

peetucket:
## Why was this change made?

We need to be able to export publications for all publication years with this rake task ... this rake task had a bug which was preventing this, and this fixes it.

Using the modified rake task for a report right now.

### DIFF
--- a/lib/tasks/sul.rake
+++ b/lib/tasks/sul.rake
@@ -197,7 +197,7 @@ namespace :sul do
           puts "#{message} : #{contributions.size} publications"
           contributions.each do |contribution|
             pub = contribution.publication
-            next unless !pub_years.empty? && pub_years.include?(pub.year)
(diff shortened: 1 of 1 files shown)


lib/tasks/sul.rake
```diff
@@ -197,7 +197,7 @@ namespace :sul do
           puts "#{message} : #{contributions.size} publications"
           contributions.each do |contribution|
             pub = contribution.publication
-            next unless !pub_years.empty? && pub_years.include?(pub.year)
+            next unless pub_years.empty? || (pub_years.present? && pub_years.include?(pub.year))
```
peetucket, line 201: if the `pub_years` array was empty, the current logic just skips every publication, which is not correct

jmartin-sul (approved):

PR #1693, opened 2024-04-22
Update dependencies

sul-devops-team:


app/models/orcid_source_record.rb
```diff
@@ -3,5 +3,5 @@
 class OrcidSourceRecord < ApplicationRecord
   belongs_to :publication, inverse_of: :orcid_source_record, optional: true
 
-  serialize :source_data, JSON
+  serialize :source_data, coder: JSON
 end
```
peetucket, line 7: fixes deprecation warnings like this: `DEPRECATION WARNING: Passing the class as positional argument is deprecated and will be removed in Rails 7.2.`

app/models/publication.rb
```diff
@@ -57,7 +57,7 @@ class Publication < ApplicationRecord
            after_add: :pubhash_needs_update!,
            after_remove: :pubhash_needs_update!
 
-  serialize :pub_hash, Hash
+  serialize :pub_hash, type: Hash
```
peetucket, line 60: ditto re: deprecation warning

justinlittman (approved):

PR #1695, opened 2024-05-01
Remove unicode gem

justinlittman:
## Why was this change made?
It doesn't seem necessary and doesn't compile on new Macs.

Unit


peetucket (approved):
Agreed that this gem doesn't seem in the dependency list for the citeproc gem

PR #1769, opened 2025-02-13
speed up orcid integration stats rake task by avoiding many API calls

### LINKED ISSUE #1768: Fix sul-pub rake task for fetching orcids so we don't iterate over all orcids
Looks like the MaIS ORCID query to fetch all orcids returns the scope now, so no need to iterate over all one by one.

peetucket:
## Why was this change made?

Fixes #1768

Localhost, returned valid results

### DIFF
--- a/lib/tasks/sul.rake
+++ b/lib/tasks/sul.rake
@@ -12,22 +12,22 @@ namespace :sul do
     orcid_users = Mais.client.fetch_orcid_users
     # NOTE: that the `fetch_orcid_users` command will returned some duplicated sunets, because it returns a recent history
     # of all changes for that sunet, with the last entry being the current scope (i.e. if they change scope, they may be returned twice)
-    sunets = orcid_users.map(&:sunetid).uniq
-    num_sunets = sunets.size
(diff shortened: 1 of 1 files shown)


lib/tasks/sul.rake
```diff
@@ -12,22 +12,22 @@ namespace :sul do
...
+    num_sunets = scopes.size
+    read_count = scopes.count { |_, scope| scope == 'read' }
+    write_count = scopes.count { |_, scope| scope == 'write' }
+
     puts "Report run: #{Time.zone.now}"
     puts "Total users: #{num_sunets}"
-    puts "Total users with read only scope: #{scopes[:read]}"
-    puts "Total users with read/write scope: #{scopes[:write]}"
+    puts "Total users with read only scope: #{read_count}"
+    puts "Total users with read/write scope: #{write_count}"
```
peetucket, line 30: this change just iterates over the "all orcid users" structure and figures out the counts using a hash, where the latest scope "wins", instead of making one by one API calls back to the MaIS API for each user

lwrubel (approved):

PR #1780, opened 2025-03-28
Update dependencies

sul-devops-team:


edsu:
This was a test to make sure dep-updates is running ok now. It is!

Issue #1783, opened 2025-03-31
[STUB] Replace DB with Postgres


mjgiarlo:
Stub

mjgiarlo:
Done in all envs now.

PR #1787, opened 2025-04-16
Switch default database from MySQL to Postgres

mjgiarlo:
# Why was this change made?

Connects to #1783

Includes:
* Change DB config to use Postgres for all envs
* Remove MySQL-isms from DB migrations and allow `db:migrate` to function with Postgres underneath
* Remove unnecessary customization from CI config and use the ruby-rails-orb instead
* Add Postgres container to `docker compose`
* Reflect above changes in README

# How was this change tested?

`bin/rake db:drop db:create db:migrate`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,63 +1,13 @@
 version: 2.1
-
 orbs:
-  codecov: codecov/codecov@5.2.0
-
-parameters:
-  ruby-version:
-    type: string
-    default: "3.4.1"
-
-jobs:
-  test:
-    docker:
-    - image: 'cimg/ruby:<< pipeline.parameters.ruby-version >>-node'
-      environment:
-        BUNDLE_JOBS: 3
-        BUNDLE_RETRY: 3
-        BUNDLE_PATH: vendor/bundle
-        RAILS_ENV: test
-        NOKOGIRI_USE_SYSTEM_LIBRARIES: true
-    - image: circleci/mysql:8
-      # https://discuss.circleci.com/t/solved-mysql-8-0-without-mysql2-authentication-plugin-caching-sha2-password-cannot-be-loaded/25791
-      command: [--default-authentication-plugin=mysql_native_password]
-      environment:
-        MYSQL_DATABASE: sulbib_test
-    steps:
-    - checkout
-    - run:
-        name: Install/Upgrade Bundler
-        command: gem install bundler
-    - run:
-        name: Which bundler?
-        command: bundle -v
-    - restore_cache:
-        keys:
-        - sul_pub-bundle-v2-{{ checksum "Gemfile.lock" }}
-        - sul_pub-bundle-v2-
-    - run:
-        name: Bundle Install
-        command: bundle check || bundle install
-    - save_cache:
-        key: sul_pub-bundle-v2-{{ checksum "Gemfile.lock" }}
-        paths:
-        - vendor/bundle
-    - run:
-        name: Check styles using rubocop
-        command: bundle exec rubocop
-    - run:
-        name: Wait for DB
-        command: dockerize -wait tcp://localhost:3306 -timeout 1m
-    - run:
-        name: Database setup
-        command: bin/rails db:test:prepare
-    - run:
-        name: Run rspec
-        command: bundle exec rspec
-    - codecov/upload
-
+  ruby-rails: sul-dlss/ruby-rails@4.6.0
 workflows:
-  version: 2
-  test:
+  build:
     jobs:
-    - test
+      - ruby-rails/lint:
+          name: lint
+          context: dlss
+      - ruby-rails/test-rails:
+          name: test
+          context: dlss
+          api-only: true # do not try to muck with JavaScript
--- a/Gemfile
+++ b/Gemfile
@@ -1,9 +1,5 @@
 source 'https://rubygems.org'
 
-gem 'rails', '~> 7.2.0'
-
-gem 'nokogiri', '>= 1.7.1'
-
 gem 'activerecord-import'
 gem 'bibtex-ruby'
 gem 'bootsnap', '>= 1.1.0', require: false
@@ -16,6 +12,8 @@ gem 'dotiw'
 gem 'faraday'
 gem 'faraday-httpclient'
 gem 'faraday-retry'
+gem 'hashie' # this used to be part of grape, but we still need it since we believe pub_hashes may still be seralized in the database this way 9/10/2018
+gem 'honeybadger'
 gem 'htmlentities', '~> 4.3'
 gem 'httpclient', '~> 2.8'
 gem 'identifiers', '~> 0.12' # Altmetric utilities related to the extraction, validation and normalization of various scholarly identifiers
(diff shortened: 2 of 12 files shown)


Gemfile
```diff
@@ -1,9 +1,5 @@
 source 'https://rubygems.org'
 
-gem 'rails', '~> 7.2.0'
-
-gem 'nokogiri', '>= 1.7.1'
-
```
mjgiarlo, line 6: I put the gems in alphabetical order. Didn't think we needed to special-case Rails and nokogiri.

db/migrate/20130703222701_change_pub_hash_to_medium_text.rb
```diff
@@ -1,5 +1,5 @@
 class ChangePubHashToMediumText < ActiveRecord::Migration[4.2]
   def change
-    change_column :publications, :pub_hash, :text, limit: 16_777_215
+    change_column :publications, :pub_hash, :text, limit: 16_777_215 if ActiveRecord::Base.connection.adapter_name.match?(/mysql/i)
```
mjgiarlo, line 3: sul_pub's current migrations are coupled to MySQL in a few spots. To make the app able to migrate up and down in both PG and MySQL, use this `if` statement. Once we're on PG in all envs, we can remove these checks for good.

db/migrate/20160405194617_switch_to_medium_text.rb
```diff
@@ -3,23 +3,44 @@
...
+      change_column :user_submitted_source_records, :source_data, :text, limit: 16_777_215
 
-    # changing versions
-    change_column :versions, :object, :text, limit: 16_777_215
+      # changing versions
+      change_column :versions, :object, :text, limit: 16_777_215
+    else
+      # changing batch_uploaded_source_records
+      change_column :batch_uploaded_source_records, :bibtex_source_data, :text
+      change_column :batch_uploaded_source_records, :error_message, :text
```
mjgiarlo, line 28: With PG as the DB in all envs, we no longer care about this limit (which was added earlier to allow both MySQL and sqlite to be used).

db/migrate/20171213201429_wos_src_record_medium_text.rb
```diff
@@ -1,6 +1,10 @@
 class WosSrcRecordMediumText < ActiveRecord::Migration[4.2]
   def up
-    change_column :web_of_science_source_records, :source_data, :mediumtext
+    if ActiveRecord::Base.connection.adapter_name.match?(/mysql/i)
+      change_column :web_of_science_source_records, :source_data, :mediumtext
```
mjgiarlo, line 4: `:mediumtext` is not a universal data type.

db/migrate/20171215001732_wos_source_record_integer_pmid.rb
```diff
@@ -1,6 +1,10 @@
 class WosSourceRecordIntegerPmid < ActiveRecord::Migration[4.2]
   def up
-    change_column :web_of_science_source_records, :pmid, :integer
+    if ActiveRecord::Base.connection.adapter_name.match?(/mysql/i)
+      change_column :web_of_science_source_records, :pmid, :integer
+    else
+      change_column :web_of_science_source_records, :pmid, :integer, using: 'pmid::integer'
```
mjgiarlo, line 6: This is the PG-ified way to cast strings to ints.

db/schema.rb
```diff
@@ -10,8 +10,11 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.0].define(version: 2022_07_18_164661) do
-  create_table "author_identities", force: :cascade do |t|
+ActiveRecord::Schema[7.2].define(version: 2022_07_18_164661) do
+  # These are extensions that must be enabled in order to support this database
+  enable_extension "plpgsql"
+
+  create_table "author_identities", id: :serial, force: :cascade do |t|
```
mjgiarlo, line 17: You'll see `id: :serial` added to each table creation, which is what rails adds for PG DB migrations.

db/schema.rb
```diff
@@ -25,11 +28,11 @@
     t.index ["author_id"], name: "index_author_identities_on_author_id"
   end
 
-  create_table "authors", force: :cascade do |t|
-    t.integer "cap_profile_id", limit: 4
+  create_table "authors", id: :serial, force: :cascade do |t|
+    t.integer "cap_profile_id"
```
mjgiarlo, line 32: You'll notice a lot of missing `limit: 4` on integer fields, and that's because a limit of `4` is the default setting for PG.

db/schema.rb
```diff
@@ -56,32 +59,32 @@
...
 
-  create_table "batch_uploaded_source_records", force: :cascade do |t|
+  create_table "batch_uploaded_source_records", id: :serial, force: :cascade do |t|
     t.string "sunet_id", limit: 255
-    t.integer "author_id", limit: 4
-    t.integer "cap_profile_id", limit: 4
+    t.integer "author_id"
+    t.integer "cap_profile_id"
     t.boolean "successful_import"
-    t.text "bibtex_source_data", limit: 16777215
```
mjgiarlo, line 64: We no longer need to know or care about the 16,777,215 number.

db/schema.rb
```diff
@@ -96,8 +99,8 @@
   end
 
   create_table "orcid_source_records", force: :cascade do |t|
-    t.text "source_data", limit: 16777215
-    t.integer "last_modified_date"
+    t.text "source_data"
+    t.bigint "last_modified_date"
```
mjgiarlo, line 103: `bigint` maps to `bigint` in PG.

.circleci/config.yml
```diff
@@ -1,63 +1,13 @@
...
-  test:
+  build:
     jobs:
-    - test
+      - ruby-rails/lint:
+          name: lint
+          context: dlss
+      - ruby-rails/test-rails:
+          name: test
+          context: dlss
```
mjgiarlo, line 12: Now that we don't need MySQL, we can use the ruby-rails-orb for CI like most of our other apps.

Gemfile.lock
```diff
@@ -461,6 +462,8 @@ GEM
       rspec-mocks (~> 3.13)
       rspec-support (~> 3.13)
     rspec-support (3.13.2)
+    rspec_junit_formatter (0.6.0)
```
mjgiarlo, line 465: Needed for CircleCI, as annotated in the Gemfile

README.md
```diff
@@ -34,26 +34,10 @@ bundle install
...
-
-For docker:
-```
-docker run --rm -e MYSQL_ALLOW_EMPTY_PASSWORD=true -p 3306:3306 -d mysql:8
-```
-and then:
-```
-RAILS_ENV=test bundle exec rake db:create
-RAILS_ENV=test bundle exec rake db:migrate
+bin/rake db:prepare
```
mjgiarlo, line 40: Prefer binstubs to explicit `bundle exec`. Less typing, more idiomatic.

README.md
```diff
@@ -64,16 +48,16 @@ To run the test suite:
...
-bundle exec rake ci
+bin/rake ci
 ```
 
 To run specific tests, use `rspec` directly, e.g.
 
 ```sh
 # Run only the publications_api_spec:
-RAILS_ENV=test bundle exec rspec spec/api/publications_api_spec.rb
+bundle exec rspec spec/api/publications_api_spec.rb
```
mjgiarlo, line 58: RSpec automatically uses the `test` env.

compose.yaml
```diff
@@ -1,16 +1,19 @@
 services:
-  mysqldb:
-   image: mysql
-   command: --default-authentication-plugin=mysql_native_password
+  mariadb:
+   image: mariadb:10.3 # the version installed on prod
```
mjgiarlo, line 3: Rename to be clearer and specify the same version running on our VMs. 

We can remove this when we're on PG in all envs.

db/migrate/20130703222701_change_pub_hash_to_medium_text.rb
peetucket: i guess the limit for text fields in postgres is at least this big?

peetucket:
thanks for sorting through this, and it looks like we at least pass the first step of getting a brand new pg database up and running with migrations and tests complete

db/migrate/20130703222701_change_pub_hash_to_medium_text.rb
mjgiarlo: @peetucket even better, `:text` fields are unlimited in Postgres 🤯

peetucket: quick, download wikipedia and put it in a text field

mjgiarlo: Why are we bothering with this repository business when PG has unlimited text fields?

jmartin-sul (approved):

PR #1795, opened 2025-05-13
Remove temp VMs and restore harvesting in UAT and QA

mjgiarlo:
# Why was this change made?

Connects to sul-dlss/operations-tasks#3952

We no longer need these temp VMs now that the Profiles services are pointing back at the non-temp VMs, which are running the latest prod DB snapshot from March.

# How was this change tested?

CI

# Which documentation and/or configurations were updated?

Removed unnecesary shared_configs branches for temp VMs


bin/bundle
```diff
@@ -1,3 +1,109 @@
 #!/usr/bin/env ruby
-ENV['BUNDLE_GEMFILE'] ||= File.expand_path('../Gemfile', __dir__)
-load Gem.bin_path('bundler', 'bundle')
+# frozen_string_literal: true
```
mjgiarlo, line 2: Opportunistic: regenerate the binstub to reflect the latest in bundler.

config/deploy.rb
```diff
@@ -27,6 +27,9 @@
   'config/settings'
 )
 
+set :log_level, :info
+set :rails_env, 'production'
+set :bundle_without, %w(development test).join(' ')
```
mjgiarlo, line 32: Common properties can be moved outside of env-specific configs.

config/deploy/uat.rb
```diff
@@ -1,9 +1,2 @@
 # see https://github.com/sul-dlss/sul_pub/wiki/Servers-Deployment-environment
-server 'sul-pub-cap-uat.stanford.edu', user: 'pub', roles: %w(web db app external_monitor)
-
-
-set :rails_env, 'production'
-
-set :bundle_without, %w(test development).join(' ')
-
-set :log_level, :info
+server 'sul-pub-cap-uat.stanford.edu', user: 'pub', roles: %w(web db app harvester_uat external_monitor)
```
mjgiarlo, line 2: Re-add `harvester_uat` role to UAT VM

peetucket (approved):

PR #1796, opened 2025-05-19
Update dependencies

sul-devops-team:


jmartin-sul:
skipping dep updates for sul_pub this week, to keep from deploying and interfering with WIP mysql to postgres migrations (which have been done in stage and QA, but not UAT or prod, both of which are scheduled for this week).  i've also commented out sul_pub in my local sdr-deploy config, to keep from tagging it or pushing it to any envs.

